PR #9, opened 2016-01-12
Implement color arithmetics

### LINKED ISSUE #2: Color arithmetics
Being able to do some basic math with colors is often useful, so it should be supported. This is just element wise arithmetics, so it's more naive than blending and can usually be applied to any color space.

Ogeon:
This implements `Add`, `Sub`, `Mul`  and `Div` as simple component wise operations, and closes #2. This is not necessarily the same thing as blending, depending on which color space we are talking about.

Hue based color spaces does only implement `Add` and `Sub` for now, since multiplication of angles isn't well defined. This may change in the future, as the behavior of the hue types solidifies.


Ogeon:
@homu r+

PR #17, opened 2016-01-21
Make all colors generic over f32 and f64

### LINKED ISSUE #13: Allow more than just f32 colors
I can be useful for some applications to have colors that are based on other numerical types than just `f32`. Most probably for the sake of higher precision or lower memory usage. It would therefore be good to support more than just `f32`, and preferably in a generic way. We can still keep `f32` as the default.

sidred:
Closes #13

Notes:
- ```T::from( constant ).unwrap()``` etc was used in a number of places to convert constants to num::Float trait ( via the num::Numcast trait).
- make_color and make_hues macros has issues due to enum names clashing with the imported colors which need a parameter. So a separate file with the macro was created ( hues_macros_gen.rs and  lib_macros_gen.rs). rustc --pretty=expanded was executed manually on these files and the result copied to lib.rs and hues.rs files

- For the hues macro gen, ```impl<T: Float> Into<T> for $name<T>```  was manually implemented for f64 and f32

- For impl for RgbPixel<T> , was manually implemented for [u8;4], [f32;4] and [f64;4]. The same was done for the tuple variants.


Ogeon:
Oh, this is a big one. A few first impression comments:

* It seems like you forgot to document `to_float`.
* I would really, really appreciate it if you _didn't_ realign everything to use visual indentation.
* I think it's safe to just lock the `num` crate to 0.1.
* Using a default type, like in `pub struct Rgb<T: Float = f32>`, may make it possible to use them without specifying what `T` is.

sidred:
I have rustfmt set to run on every save, none of the indentation was manual.

Ogeon:
Yeah, I suspected that rustfmt had something to do with it, and it's a great tool, but automatically restyling every project you contribute to isn't as great. Note that this isn't about liking or disliking visual indentation. It's pretty and all, and I would probably use it with rustfmt if it wasn't for the downsides:

* It has the tendency to result in _a lot_ of rightwards drift. I have seen code becoming ridiculously squeezed along the character limit because of long function names.
* It's noise in git diffs. Reindenting stuff because of a name change will have an unnecessarily large impact and may cause merge conflicts. Just because it's pretty.
* It's tedious and unpractical unless it's automated.

I did use visual indentation once upon a time, but I have since moved on to simply indent one tab (four spaces) for each "level".

Ogeon:
Also, regarding rustfmt in general; I prefer not to depend on it, since it's not really ready yet. There are still too many cases where it may mangle the code too much.

sidred:
The major fomatting issues are in the examples. Let me see if I can revert those and squash all the commits. Might be easier to read.

Ogeon:
Take your time. As I said, it's not personal or anything else. Just a style and conventions thing. I should probably add a section to CONTRIBUTING.md about it...

src/hsl.rs
```diff
@@ -13,70 +16,70 @@ use {Color, Rgb, Luma, Xyz, Lab, Lch, Hsv, ColorSpace, Mix, Shade, GetHue, Hue,
 ///
 ///See [HSV](struct.Hsv.html) for a very similar color space, with brightness instead of lightness.
 #[derive(Clone, Copy, Debug, PartialEq)]
-pub struct Hsl {
+pub struct Hsl<T: Float = f32> {
     ///The hue of the color, in degrees. Decides if it's red, blue, purple,
     ///etc.
-	pub hue: RgbHue,
+    pub hue: RgbHue<T>,
```
Ogeon: Keep these tab -> space changes. The tabs must have been from my stupid laptop and spaces is correct.

src/hsl.rs
```diff
@@ -13,70 +16,70 @@ use {Color, Rgb, Luma, Xyz, Lab, Lch, Hsv, ColorSpace, Mix, Shade, GetHue, Hue,
...
-    ///The colorfulness of the color. 0.0 gives gray scale colors and 1.0 will
+    ///The colorfulness of the color. T::zero() gives gray scale colors and T::one() will
     ///give absolutely clear colors.
-	pub saturation: f32,
+    pub saturation: T,
 
-    ///Decides how light the color will look. 0.0 will be black, 0.5 will give
-    ///a clear color, and 1.0 will give white.
-	pub lightness: f32,
+    ///Decides how light the color will look. T::zero() will be black, T::from(0.5).unwrap() will give
```
Ogeon: I wonder if it would be better to keep these as they were. I don't think people will be more confused by `0.0` and `1.0` than `T::zero()` and `T::one()` ...and `T::from(0.5).unwrap()` looks very awkward to have in the documentation, to be honest.

src/hues.rs
```diff
@@ -12,81 +16,102 @@ macro_rules! make_hues {
...
 
-        impl Into<f32> for $name {
+        // TO-DO:
+        // Error: with  conflicting implementations of trait `core::convert::Into<_>` for type  `hues::RgbHue<_>`
+        // Error: conflicting implementation in crate `core`
+        // impl<T:Float> Into<T> for $name<T> {
+        //     fn into(self) -> T {
+        //         normalize_angle(self.0)
+        //     }
+        // }
```
Ogeon: I don't think that will be possible at all before specialization arrives. In any case, it would be great if we can keep the to-dos in issues instead of comments :smile:

src/hues.rs
```diff
@@ -12,81 +16,102 @@ macro_rules! make_hues {
...
+            fn into(self) -> f64 {
+                normalize_angle(self.0).to_f64().unwrap()
+            }
+        }
+        impl<T: Float> Into<f32> for $name<T> {
             fn into(self) -> f32 {
-                normalize_angle(self.0)
+                normalize_angle(self.0).to_f32().unwrap()
             }
         }
```
Ogeon: I'm not sure about `Into` implementations that may panic... Maybe we should just make those completely explicit. What do you think?

src/hues.rs
```diff
@@ -12,81 +16,102 @@ macro_rules! make_hues {
...
-            }
-        }
+        // TO-DO
+        // Error: type parameter `T` must be used as the type parameter for some local type (e.g. `MyStruct<T>`);
+        // Error: only traits defined in the current crate can be implemented for a type parameter
+        // impl<T:Float> PartialEq<$name<T>> for T {
+        //     fn eq(&self, other: &$name<T>) -> bool {
+        //         other.eq(self)
+        //     }
+        // }
```
Ogeon: Looks like we may have to scrap that one...

src/lab.rs
```diff
@@ -224,69 +225,75 @@ impl Div<f32> for Lab {
...
     }
 }
 
-fn f(t: f32) -> f32 {
-    //(6/29)^3
-    const C_6_O_29_P_3: f32 = 0.00885645167;
-    //(29/6)^2
-    const C_29_O_6_P_2: f32 = 23.3611111111;
+fn f<T: Float>(t: T) -> T {
+    // (T::from(6/29).unwrap())^3
```
Ogeon: No need to change these comments. They are just there to show where the numbers comes from.

src/lch.rs
```diff
@@ -1,6 +1,9 @@
+use num::traits::Float;
+
 use std::ops::{Add, Sub};
 
-use {Color, ColorSpace, Mix, Shade, GetHue, Hue, Rgb, Luma, Xyz, Lab, Hsv, Hsl, Saturate, LabHue, clamp};
+use {Color, ColorSpace, Mix, Shade, GetHue, Hue, Rgb, Luma, Xyz, Lab, Hsv, Hsl, Saturate, LabHue,
+     clamp};
```
Ogeon: Hah! I would rather have it completely vertical or grouped. No big deal, though :smile:

src/lch.rs
```diff
@@ -10,72 +13,72 @@ use {Color, ColorSpace, Mix, Shade, GetHue, Hue, Rgb, Luma, Xyz, Lab, Hsv, Hsl,
...
 }
 
-impl ColorSpace for Lch {
+impl<T: Float> ColorSpace for Lch<T> {
     fn is_valid(&self) -> bool {
-        self.l >= 0.0 && self.l <= 1.0 &&
-        self.chroma >= 0.0 && self.chroma <= 1.41421356 && //should include all of L*a*b*, but will also overshoot...
-        self.alpha >= 0.0 && self.alpha <= 1.0
+        self.l >= T::zero() && self.l <= T::one() && self.chroma >= T::zero() &&
+        self.chroma <= T::from(1.41421356).unwrap() && self.alpha >= T::zero() &&
```
Ogeon: Where did the comment go?

src/lch.rs
```diff
@@ -128,27 +131,27 @@ impl Hue for Lch {
...
 }
 
-impl Saturate for Lch {
-    fn saturate(&self, factor: f32) -> Lch {
+impl<T: Float> Saturate<T> for Lch<T> {
+    fn saturate(&self, factor: T) -> Lch<T> {
         Lch {
             l: self.l,
-            chroma: self.chroma * (1.0 + factor),
+            chroma: self.chroma * (T::from(1.0).unwrap() + factor),
```
Ogeon: Why not `T::one()`?

src/lch.rs
```diff
@@ -128,27 +131,27 @@ impl Hue for Lch {
...
         }
     }
 }
 
-impl Default for Lch {
-    fn default() -> Lch {
-        Lch::lch(0.0, 0.0, 0.0.into())
+impl<T: Float> Default for Lch<T> {
+    fn default() -> Lch<T> {
+        Lch::<T>::lch(T::zero(), T::zero(), LabHue::from(T::zero()))
```
Ogeon: Is the `Lch::<T>::` part really necessary?

src/lib.rs
```diff
@@ -101,60 +105,60 @@ macro_rules! make_color {
         ///but it can easily be converted to a fixed color space in those
         ///cases.
         #[derive(Clone, Copy, Debug)]
-        pub enum Color {
-            $(#[$variant_comment] $variant($variant)),+
+        pub enum Color<T:Float> {
```
Ogeon: I think this one can get a default type, as well.

src/lib.rs
```diff
@@ -307,13 +316,13 @@ pub trait ColorSpace {
 ///assert_eq!(a.mix(&b, 0.5), Rgb::linear_rgb(0.5, 0.5, 0.5));
 ///assert_eq!(a.mix(&b, 1.0), b);
 ///```
-pub trait Mix {
+pub trait Mix<T:Float> {
```
Ogeon: It may be nice to turn `T` into an associated type.

src/rgb.rs
```diff
@@ -281,30 +276,32 @@ impl Shade for Rgb {
...
 
-impl GetHue for Rgb {
-    type Hue = RgbHue;
+impl<T: Float> GetHue for Rgb<T> {
+    type Hue = RgbHue<T>;
 
-    fn get_hue(&self) -> Option<RgbHue> {
-        const SQRT_3: f32 = 1.73205081;
+    fn get_hue(&self) -> Option<RgbHue<T>> {
+        let sqrt_3: T = T::sqrt(T::from(3).unwrap());
```
Ogeon: Square roots are expensive. I would prefer to keep the pre-computed one.

src/rgb.rs
```diff
@@ -531,96 +534,379 @@ fn to_gamma(x: f32, gamma: f32) -> f32 {
...
+// impl<T: Float> RgbPixel<T> for (T, T, T) {
+//     fn from_rgba(red: T, green: T, blue: T, _alpha: T) -> (T, T, T) {
+//         (red, green, blue)
+//     }
+//
+//     fn to_rgba(&self) -> (T, T, T, T) {
+//         let (r, g, b) = *self;
+//         (r, g, b, T::one())
+//     }
+// }
```
Ogeon: I would, once again, prefer it if we didn't keep commented out code around.

src/lab.rs
```diff
@@ -89,10 +90,10 @@ impl Mix for Lab {
     }
 }
 
-impl Shade for Lab {
-    fn lighten(&self, amount: f32) -> Lab {
+impl<T: Float> Shade<T> for Lab<T> {
+    fn lighten(&self, amount: T) -> Lab<T> {
         Lab {
-            l: self.l + amount * 1.0,
+            l: self.l + amount * T::one(),
```
Ogeon: That `* 1.0` isn't supposed to be there and can safely be removed :smile:

Ogeon:
I have now made a first read-through and commented on some things (but not on every occurrence). It's otherwise very good and will be great with some polish.

src/hsl.rs
sidred: These were all mistakes on my part. Correcting them.

src/hues.rs
sidred: This should not panic. We are doing f64-> f32 conversion. If f64 is too large, it will become nan in f32

Ogeon: We are doing any `T` to `f32` and every `unwrap` on an unknown type is a potential panic.

sidred: Impl specifically restricts it to float. So its either a f64->f32 or f32 -> f32 conversion

src/lib.rs
sidred: Doesn't seem to be ergonomic.

```
min_color.mix(max_color, factor)
```
errors with
 ``` 
 expected `<C as Mix>::T`,
    found `T`

```

**Comment by sidred** [comment]
Closing in favour of #18

File: src/hues.rs (hunk 4 above)
Comment (Ogeon) [inline src/hues.rs]: `Float` can be implemented by anything. It's just a trait.

File: src/lib.rs (hunk 12 above)
Comment (Ogeon) [inline src/lib.rs]: Huh? What did you change it to?

Comment (sidred) [inline src/lib.rs]: I changed it to 
```
pub trait Mix {
   type T: Float;
    fn mix(&self, other: &Self, factor: T) -> Self;
}

```

Ogeon: That looks just right. I wonder why it would complain... Alright, we can leave it as it is and deal with that later.

PR #22, opened 2016-01-23
Extract the alpha component as a wrapper type

### LINKED ISSUE #11: Find a good way to decouple the alpha channel
Having the alpha channel being affected by operations as arithmetics isn't always what you want, so it would be nice if it can be separated from the color types in a good, convenient and intuitive enough way. This is also useful for when the alpha channel is insignificant (always 1.0), to save some space. I have been thinking about an `Alpha<C>` wrapper type along these lines:

```rust
struct Alpha<C> {
    pub color: C,
    pub alpha: f32,
}

impl<C: Mix> Mix for Alpha<C> {
    //...
}

//etc.
```

and maybe even make it dereference as the color type `C`. I don't know if this will end up being too inconvenient or not, and better solutions are always welcome. I'll see if I can figure something else out before making a test implementation.

Ogeon:
This closes #11 by adding a wrapper type that bolts the alpha component to any color. This works well because of how passive the alpha is in most of the calculations and conversions, and results in a relatively low amount of code duplication.

Each color space (and `Color`) has a sibling type alias with an `a` added to the name, such as `Rgba`. Those are alias for `Alpha<C<T>, T>`, where `C` is the color space type, and can be used more or less in the same way as their non-alpha variants. Conversion between alpha types and non-alpha types is as easy as always.

This is a breaking change, since the alpha component is removed from every color and constructors are moved and/or renamed. The new constructors follows the same convention as the constructors in `Srgb` and `GammaRgb`, which is `new` for float values and `new_u8` for `u8` values. The constructors on `Color` are roughly the same as before, with the exception of changing things like `rgb8` to `rgb_u8`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,13 +110,13 @@ extern crate palette;
 use palette::{Rgb, Hsv, Gradient};
 
 let grad1 = Gradient::new(vec![
-    Rgb::rgb(1.0, 0.1, 0.1),
-    Rgb::rgb(0.1, 1.0, 1.0)
+    Rgb::new(1.0, 0.1, 0.1),
+    Rgb::new(0.1, 1.0, 1.0)
 ]);
 
 let grad2 = Gradient::new(vec![
-    Hsv::from(Rgb::rgb(1.0, 0.1, 0.1)),
-    Hsv::from(Rgb::rgb(0.1, 1.0, 1.0))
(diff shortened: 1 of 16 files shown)


Ogeon:
@homu r+

PR #23, opened 2016-01-23
Change Mix, Shade and Saturate to use an associated type

Ogeon:
The `Mix`, `Shade` and `Saturate` traits are rewritten to have an associated `Scalar` type instead of a type parameter, as mentioned in [#17](https://github.com/Ogeon/palette/pull/17#discussion_r50449575). This removes the need for extra type parameters in some cases, like in `Gradient`.

This is a breaking change, since it removes a number of type parameters.

### DIFF
--- a/examples/readme_examples.rs
+++ b/examples/readme_examples.rs
@@ -69,7 +69,7 @@ fn display_colors(filename: &str, colors: &[[u8; 3]]) {
     }
 }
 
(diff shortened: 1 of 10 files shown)


Ogeon:
The type name `Scalar` is up for some bike shedding, but I think it's technically correct.

Ogeon:
@homu r+

PR #18, opened 2016-01-22
Convert all colors to be generic over floats, f32 and f64

sidred:
**Convert all colors to be generic over floats**

Change all colors (Color, Rgb, Hsl, Hsv, Lab, Xyz) and traits (Mix and Saturation etc) to be generic over floats and closes #13.

**_This is a breaking change. Type f32 or f64 must be explicitly annotated._**

```
// OLD
let new_color: Rgb = lch_color.shift_hue(180.0.into()).into();

// NEW
let new_color: Rgb<f32> = lch_color.shift_hue(180.0.into()).into();
```

Since Rust defaults to f64, most of the color types will default to f64. For example Rgb::linear_rgb(1.0, 0.0, 0.0) will be inferred as Rgb\<f64\>. To use f32, constants must be explicitly set like rgb::linear_rgb(1.0_f32, 0.0, 0.0) or the type explicitly annotated.

T::from( constant ).unwrap() etc is used in a number of places to convert constants to num::Float trait ( via the num::Numcast trait).

For Hues, the Into<T> trait is not generic over floats due to unwrap() and possibility of runtime panic and is separately implemented for f32 and f64 for Hue\<32\> and Hue\<64\>. Also the Partial Eq trait could not be implemented for Hues as float and was removed.

The RgbPixel could not be made generic for [T;3] and (T,T,T) and is separately implemented for f32, f64 and u8.

Using associated typed for Mix and Saturation is not ergonomic and the trait was made generic over float.

PR #24, opened 2016-01-24
Separate sRGB and gamma encoded RGB from the Rgb type

### LINKED ISSUE #7: Separation of the RGB types
This makes it possible to make use of the type system to differ between them and linear RGB, and lowers the number of constructors on `Rgb`.

[Suggestion from Reddit user CryZe92](https://www.reddit.com/r/rust/comments/40mc8d/presenting_palette_a_linear_color_calculation_and/cyv9dyp)

Ogeon:
This closes #7 by adding separate types for sRGB and gamma encoded RGB. `Rgb` becomes a purely linear variant of sRGB and the types for encoded sRGB and gamma RGB can be found in a newly added `pixel` module. The `RgbPixel` trait has also been moved to the `pixel` module.

The `linear_*` prefix has also been dropped from the `Rgb` constructors and some bugs from #18 were fixed as a bonus.

This is a breaking change, since it moves, removes and renames a lot of functions, and the `RgbPixel` trait is no longer available in the crate root.


Ogeon:
There were some talk somewhere about supporting AdobeRGB, but that can be added later. I'll call this one done.

@homu r+

Issue #10, opened 2016-01-13
Ranges for CIE color spaces


sidred:
For XYZ, both http://www.easyrgb.com/ and http://www.brucelindbloom.com/index.html?ColorCalculator.html show z as 1.088830 for srgb(1,1,1).

The is_valid() method for xyz assumes a range of [0-1]. 

For Lab, both http://www.easyrgb.com/ and http://www.brucelindbloom.com/index.html?ColorCalculator.html have values above 1 and also negative. For example, srgb(1,0.5,1) -> lab(72.087  ,65.181,  -42.228)

The is_valid() method for lab assumes a range of [0-1]. 

Not sure what the correct range is. Opening this issue for tracking this.

Ogeon:
Thank you for pointing this out. Lab has normally a range of L: 0 - 100, a: -127 - 128, b: -127 - 128, as I understand it, which is adapted to an 8 bit storage format. I have chosen to normalize these to 0 - 1 for L and -1 - 1 for a and b. The reason for this is the same as in `Rgb`, to make calculations (especially multiplication) easier. Other common ranges for these are 0 - 100 and -100 - 100, respectively.

The `Xyz` situation, on the other hand, is probably a mistake from my side. These should be corrected in the best possible way.

Ogeon:
[This answer on mathematica.stackexchange.com](http://mathematica.stackexchange.com/a/22772) suggests that the ranges of XYZ depends on the white point. Maybe we should just remove the upper limits for now.

sidred:
Since the struct values are all public, we should try to use existing conventions where possible. This will ensure less surprises for the end users of the library. I'll try to look at ImageMagik, OpenCV and GIMP sources to see what they use. 

Can you explain how normalizing makes multiplication easier? I am not sure I understand what you are trying to say.

Ogeon:
I have to admit that multiplying two L*a*b* values is not so intuitive, compared to RGB, so it may not matter as much. The _general_ thought can be illustrated with light and reflection in RGB space. If you have white light (rgb(1, 1, 1)) that should be reflected from a red surface (rgb(1, 0, 0)) you want the resulting light to also be red. Having an upper limit >1 makes this operation require additional computations to scale the resulting color down by the white point.

[...]

PR #25, opened 2016-01-27
Fix or relax some color ranges and clamping

### LINKED ISSUE #19: Add and Substract - clamp after operations?
For colors with known ranges like rgb, hsl etc., should we clamp the values? A possible issue is with the f32 -> u8 conversion panicking if not clamped.

Ogeon:
The `ColorSpace` trait has been repurposed as a trait for checking limits and clamping, and is now called `Limited`. The limits of `Xyz` has been changed to use the white point as its maximum (closing #10), and the maximum of `Lch.chroma` has been removed, since it's not very well defined. The ranges of `Lch` stays as they are, for the sake of consistency.

Extra clamping is also added when converting RGB colors to `u8` based pixel representation, to prevent float shenanigans when multiplying values with 255.0. It's impossible to know what a type `T` will do when it's asked to convert `255.12312312` or whatever to `u8`. This closes #19.

This is a breaking change, since it changes the ranges of some color spaces and renames the `ColorSpace` trait to `Limited`.

### DIFF
--- a/src/alpha.rs
+++ b/src/alpha.rs
@@ -2,7 +2,7 @@ use std::ops::{Deref, DerefMut, Add, Sub, Mul, Div};
 
 use num::Float;
 
-use {Mix, Shade, GetHue, Hue, Saturate};
+use {Mix, Shade, GetHue, Hue, Saturate, Limited, clamp};
 
 ///An alpha component wrapper for colors.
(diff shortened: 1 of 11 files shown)


Ogeon:
@homu r+

PR #26, opened 2016-01-29
Offer both 0 centered and positive hue -> float conversion

### LINKED ISSUE #15: Consider converting hue types to numbers within [0, 360) instead of (-180, 180]
Numbers within [0, 360) degrees seems to be more common when working with hues, but the trigonometric functions in Rust gives negative angles instead of angles above 180 degrees. The big question is which one to adopt for our hues. Note that this is only relevant when printing a hue type or when converting it to a "regular" number (e.g. in `impl Into<f32>`).

Ogeon:
This is a compromise for the 0 centered vs all positive hue representation problem, where both alternatives are offered, but the current 0 centered behavior is kept as the default. The argument for this decision is that both variants are useful in different situations, and there is no obvious reason why one would be better than the other. The default is kept because it's consistent with how the trigonometric functions in `libstd` works. It's not the strongest reason, but it's as good as any.

Changes in this PR:

 * Rename `to_float` to `to_degrees`.
 * Add `to_positive_degrees` and `to_positive_radians`.
 * Mention the ranges in the descriptions of the conversion functions.

This is a breaking change, since it renames `to_float` to `to_degrees`. It closes #15.

### DIFF
--- a/src/hsl.rs
+++ b/src/hsl.rs
@@ -77,7 +77,7 @@ impl<T: Float> Mix for Hsl<T> {
 
     fn mix(&self, other: &Hsl<T>, factor: T) -> Hsl<T> {
         let factor = clamp(factor, T::zero(), T::one());
-        let hue_diff: T = (other.hue - self.hue).to_float();
+        let hue_diff: T = (other.hue - self.hue).to_degrees();
 
         Hsl {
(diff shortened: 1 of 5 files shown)


Ogeon:
@homu r+

PR #29, opened 2016-01-31
Derive Clone for Take and MaybeSlice

paholg:
This allows `cycle()` to be called on `Take`.


Ogeon:
Nice, thank you!

@homu r+

PR #30, opened 2016-02-01
Add Cie Yxy (xyY) Color Space

### LINKED ISSUE #28: xyY color space
It's common and useful. We will probably need it sooner or later, anyway.

sidred:
# Add Cie Yxy color space

### Breaking Change
Breaking change in the Luma Color to XYZ color conversion.  The previous conversion 
did not scale the X and Y values by the  luma amount as per the formula

Previous formula:
x = X_N
y = luma
z = Z_N

Corrected formula:
x = X_N * luma
y = luma
z = Z_N * luma

### Cie Yxy color space
Add support for the Cie Yxy color space and conversions to the
other color spaces.

Color Space is named Yxy and not Xyy. I have seen Yxy used in a few
places and it fits the rust style naming better. Easier to distinguish
Yxy and Xyz than Xyy and Xyz.

The brightness/ luminance component Y is named luma and the x and y
chromacity coordinates are named x and y.

The clamping for this color space defined to be between 0-1. In reality,
Y is bounded to 0-1 and x and y have no bounds. I have seen some rgb color
spaces have their blue points negative, but for most cases 0-1 is a good
rule of thumb to use.

The default value is (0.312727, 0.329023, 0.0) and not (0,0,0). The default for x and y 
are the white point D65 x and y values which is the default white point. Since Y (luma) is 0.0, this 
makes the default color black just like for other colors. The reason for not using 0 for 
x and y is that this outside the usual color gamut and might cause scaling issues.

The Luma to Yxy conversion also uses the x and y values from the D65 white point, 
for the same reasons mentioned above.

closes  #28

### DIFF
--- a/src/yxy.rs
+++ b/src/yxy.rs
@@ -1,0 +1,320 @@
+use num::traits::Float;
+
+use std::ops::{Add, Sub, Mul, Div};
+
+use {Color, Alpha, Rgb, Luma, Lab, Lch, Hsv, Hsl, Limited, Mix, Shade, clamp, Xyz};
+
+const D65_X: f64 = 0.312727;
+const D65_Y: f64 = 0.329023;
+
+///CIE 1931 Yxy (xyY) with an alpha component. See the [`Yxya` implementation in `Alpha`](struct.Alpha.html#Yxya).
+pub type Yxya<T = f32> = Alpha<Yxy<T>, T>;
+
+///The CIE 1931 Yxy (xyY)  color space.
+///
+///Yxy is a luminance-chromaticity color space derived from the CIE XYZ
+///color space. It is widely used to define colors. The chromacity diagrams
+///for the color spaces are a plot of this color space's x and y coordiantes.
+///
+///Conversions and operations on this color space assumes the CIE Standard
+///Illuminant D65 as the white point, and the 2° standard colorimetric
+///observer.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Yxy<T: Float = f32> {
+
+    ///x chromacity co-ordinate derived from XYZ color space as X/(X+Y+Z).
+    ///Typical range is between 0 and 1
+    pub x: T,
+
+    ///y chromacity co-ordinate derived from XYZ color space as Y/(X+Y+Z).
+    ///Typical range is between 0 and 1
+    pub y: T,
+
+    ///luma (Y) was a measure of the brightness or luminance of a color.
+    ///It is the same as the Y from the XYZ color space. Its range is from
+    ///0 to 1, where 0 is black and 1 is white.
+    pub luma: T,
+}
+
+impl<T: Float> Yxy<T> {
+    ///CIE Yxy.
+    pub fn new(x: T, y: T, luma: T,) -> Yxy<T> {
+        Yxy {
+            x: x,
+            y: y,
+            luma: luma,
+        }
+    }
+}
+
+///<span id="Yxya"></span>[`Yxya`](type.Yxya.html) implementations.
+impl<T: Float> Alpha<Yxy<T>, T> {
+    ///CIE Yxy and transparency.
+    pub fn new(x: T, y: T, luma: T, alpha: T) -> Yxya<T> {
+        Alpha {
+            color: Yxy::new(x, y, luma),
+            alpha: alpha,
+        }
+    }
+}
+
+impl<T: Float> Limited for Yxy<T> {
+    fn is_valid(&self) -> bool {
+        self.x >= T::zero() && self.x <= T::one() &&
+        self.y >= T::zero() && self.y <= T::one() &&
+        self.luma >= T::zero() && self.luma <= T::one()
+    }
+
+    fn clamp(&self) -> Yxy<T> {
+        let mut c = *self;
+        c.clamp_self();
+        c
+    }
+
+    fn clamp_self(&mut self) {
+        self.x= clamp(self.x, T::zero(), T::one());
+        self.y = clamp(self.y, T::zero(), T::one());
+        self.luma = clamp(self.luma, T::zero(), T::one());
+    }
+}
+
+impl<T: Float> Mix for Yxy<T> {
+    type Scalar = T;
+
+    fn mix(&self, other: &Yxy<T>, factor: T) -> Yxy<T> {
+        let factor = clamp(factor, T::zero(), T::one());
+
+        Yxy {
+            x: self.x + factor * (other.x - self.x),
+            y: self.y + factor * (other.y - self.y),
+            luma: self.luma + factor * (other.luma - self.luma),
+        }
+    }
+}
+
+impl<T: Float> Shade for Yxy<T> {
+    type Scalar = T;
+
+    fn lighten(&self, amount: T) -> Yxy<T> {
+        Yxy {
+            x: self.x,
+            y: self.y,
+            luma: self.luma + amount,
+        }
+    }
+}
+
+impl<T: Float> Default for Yxy<T> {
+    fn default() -> Yxy<T> {
+        // The default for x and y are the white point x and y ( from the default D65).
+        // Since Y (luma) is 0.0, this makes the default color black just like for other colors.
+        // The reason for not using 0 for x and y is that this outside the usual color gamut and might
+        // cause scaling issues.
+        Yxy::new(T::from(D65_X).unwrap(), T::from(D65_Y).unwrap(), T::zero())
+    }
+}
+
+impl<T: Float> Add<Yxy<T>> for Yxy<T> {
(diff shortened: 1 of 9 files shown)


src/yxy.rs
```diff
@@ -0,0 +1,301 @@
...
+///Illuminant D65 as the white point, and the 2° standard colorimetric
+///observer.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Yxy<T: Float = f32> {
+
+    ///x chromacity co-ordinate derived from XYZ color space as X/(X+Y+Z)
+    pub x: T,
+
+    ///y chromacity co-ordinate derived from XYZ color space as Y/(X+Y+Z)
+    pub y: T,
```
Ogeon: It may be worth mentioning the recommended ranges for `x` and `y`.

src/yxy.rs
```diff
@@ -0,0 +1,301 @@
...
+            y_luma: xyz.y,
+        }
+    }
+}
+
+impl<T: Float> From<Luma<T>> for Yxy<T> {
+    fn from(luma: Luma<T>) -> Yxy<T> {
+        Yxy {
+            x: T::zero(),
+            y: T::zero(),
```
Ogeon: Shouldn't these be the coordinates of the white point? (0, 0) is some point outside the visible color gamut.

src/yxy.rs
```diff
@@ -0,0 +1,301 @@
...
+        let b = Yxy::new(0.3, 0.6, 0.715152);
+        assert_approx_eq!(a, b, [x, y, y_luma]);
+    }
+
+    #[test]
+    fn blue() {
+        let a = Yxy::from(Rgb::new(0.0, 0.0, 1.0));
+        println!("A: {:?}",a );
+        let b = Yxy::new(0.15, 0.06, 0.072175);
+        println!("B : {:?}",b );
```
Ogeon: Was these prints intentionally added?

Ogeon:
Thank you for this! I've added a few comments in the source, but it's otherwise great.

> Easier to distinguish Yxy and Xyz than Xyy and Xyz.

Good point. I'm sure people will get it anyway :smile:

> The brightness/ luminance component Y is named y_luma and the x and y
chromacity coordinates are named x and y.

What about just `luma`? Or is that too obscure?

> but for most cases 0-1 is a good rule of thumb to use

That's how it's defined, as far as I know. Clamping is optional, in any case, so it won't be crippling.

src/yxy.rs
sidred: Yeah Sorry

src/yxy.rs
sidred: I was just following the XYZ space which uses 0 for x and z in the luma calculation, which translates to 0 for x and y

sidred:
Changed the XYZ - Yxy conversions to be safer when denominator in zero.

src/yxy.rs
Ogeon: The luma -> XYZ conversion is probably wrong. I didn't think about how X, Y and Z changes together when I wrote it.

sidred:
> The brightness/ luminance component Y is named y_luma and the x and y
> chromacity coordinates are named x and y.

> What about just luma? Or is that too obscure?

I have made this change, luma seems cleaner.

> 
> but for most cases 0-1 is a good rule of thumb to use
> 
> That's how it's defined, as far as I know. Clamping is optional, in any case, so it won't be crippling.

I remember seeing some allcolor rgb or wide rgb with the blue point negative.

src/yxy.rs
sidred: Done

Ogeon:
> I remember seeing some allcolor rgb or wide rgb with the blue point negative.
Yes, those are strange. I'm trying to find where I read about the x and y ranges... Limiting them may be unnecessary if their limits aren't well defined.

src/yxy.rs
```diff
@@ -0,0 +1,304 @@
...
+///observer.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Yxy<T: Float = f32> {
+
+    ///x chromacity co-ordinate derived from XYZ color space as X/(X+Y+Z)
+    ///Typical range is between 0 and 1
+    pub x: T,
+
+    ///y chromacity co-ordinate derived from XYZ color space as Y/(X+Y+Z)
+    ///Typical range is between 0 and 1
```
Ogeon: Forgot the `.` between sentences :wink:

sidred: Damn...

src/yxy.rs
sidred: Should we just use 0.3 for Yxy and 1.0 for XYZ?

Ogeon: We are still assuming D65, so it should be the D65 white point. We should add tests for this...

sidred: What sort of test did you have in mind?

Ogeon: Something to check that gray scale conversion is correct and prevent us from making this mistake again. Maybe check that `Luma` with 1.0 and some gray value, like 0.5, converts to the correct xyY and XYZ, similar to the other color tests.

sidred: Added the d65 values for luma conversiion for xyz and yxy

src/yxy.rs
```diff
@@ -0,0 +1,305 @@
...
+            x: self.x,
+            y: self.y,
+            luma: self.luma + amount,
+        }
+    }
+}
+
+impl<T: Float> Default for Yxy<T> {
+    fn default() -> Yxy<T> {
+        Yxy::new(T::zero(), T::zero(), T::zero())
```
Ogeon: I suppose this should also be changed to the x and y of the white point.

sidred: I was thinking about this. Rgb's are using black as the default (0,0,0) . Lab is using 0,0,0 as well which translates to rgb 0's. both yxy and xyz as 0's translate to rgb 0's. So 0's might be a better value for these

Ogeon: Both ways of reasoning makes sense, and x and y doesn't matter as long as Y is 0. Things becomes stranger when it's mixed with other colors and that's where I think the white point default makes more sense. That makes the default gray scale black, and not just 0.

src/yxy.rs
sidred: May be easier to convert luma to rgb and then to xyz or yxy 

for luma = 0.5

rgb = 0.5, 0.5, 0.5
xyz = 0.475235, 0.5, 0.544415
yxy = 0.312727, 0.329023, 0.5

This fails for xyz. Seems that x and z need to be scaled by the luma value

for luma = 1.0

rgb = 1.0, 1.0, 1.0
xyz = 0.950470, 0.5, 1.088830
yxy = 0.312727, 0.329023, 1.0

Ogeon: >xyz = 0.950470, 0.5, 1.088830

Shouldn't this be xyz = 0.950470, 0.1, 1.088830? Seems like the general formula is

```
x = X_N * luma
y = luma
z = Z_N * luma
```

sidred:
Corrected the luma -> xyz conversion formula and added test cases for both luma -> xyz, yxy conversions.

src/yxy.rs
sidred: yeah all the elements and scaled by the luma. I made this change

src/yxy.rs
sidred: I've used 0.312727, 0.329023, 0.0 which is black, as the default value

Ogeon:
Nice! I think that's all, unless you have something else you want to add. It would be great if you could mention the change of how luma is converted to XYZ in the PR description. It's technically a breaking change, but it's also a correction of incorrect behavior. I'll let it pass as a bug fix. It's still good to let people know that it happened.

Ogeon:
Oh, and it's probably best to squash the commits.

sidred:
done

Ogeon:
You don't have to copy the PR description in the commit message. It will still go into the log, through Homu. Just so you know.

sidred:
Its nicer to have it show up in git history. That's why I copy it.

Ogeon:
As I said, it does anyway. Homu will put your PR description in the merge commit.

Ogeon:
See [here](https://github.com/Ogeon/palette/commit/1dca1b5d61c4f7e197a2d127b900fe85f3b03ad6), for example.

sidred:
Thanks. I was looking in the wrong place.

Ogeon:
Do you want to change it before I merge?

sidred:
ok

sidred:
Hold on there is a mistake in lib.rs

Ogeon:
Oh, the z instead of luma? Looks like I glossed over that one... Good catch :smile:

Ogeon:
Alright, here we go. Thank you, once again!

@homu r+

PR #34, opened 2016-02-01
Add color constants

### LINKED ISSUE #5: Named colors
A collection of named color constants (i.e. the named CSS and HTML colors) may be good to have. It should be possible to choose what kind of gamma correction they are assumed have when converting them to a linear RGB value.

Ogeon:
This adds the SVG1.1/CSS3 keyword colors as constants, and a function which maps name strings to them. They are accessible through the `named` module and the documentation shows colored preview rectangles as descriptions. There are even more named colors out there (https://en.wikipedia.org/wiki/List_of_colors:_A%E2%80%93F, for example), but this is a good start.

The whole `named` module can be toggled with the `"named"` feature, and the `named::from_str` function can be toggled with `"named_from_str"`.

This closes #5.


Ogeon:
I'm considering adding a `&str` -> color function. It could certainly be useful, but it would be big.

Ogeon:
That'll be all, this time.

@homu r+

Issue #41, opened 2016-02-04
Converting from a wider gamut to a smaller one should return an option


enolan:
This is real confusing:

```
extern crate palette;
use palette::*;

fn main() {
    let foo = Lab {l: 1., a: -1., b: 0.};
    assert!(foo.is_valid());
    assert!(Rgb::from(foo).is_valid());
}
```

```
enolan at Behemoth in ~/mystuff/code/palettebug  
$ cargo run
     Running `target/debug/palettebug`
thread '<main>' panicked at 'assertion failed: Rgb::from(foo).is_valid()', main.rs:7
Process didn't exit successfully: `target/debug/palettebug` (exit code: 101)
```

I assumed it was a bug in the conversion routine, but it's just that L*a*b* is wider than sRGB. Returning an option type, `assert!`ing when the input is outside the target color space, or clamping the output value would all be preferable to the current situation.

sidred:
Yeah the output of the conversion needs to be clamped. The example prints:
```
Rgb { red: -0.8116134892469634, green: 1.5417271999753697, blue: 0.9686961172534542 }

```

Ogeon:
The reason for having unchecked conversion is to keep it lossless in situations where imaginary or otherwise invalid colors are acceptable for pure mathematical reasons, but I do see why this can be problematic. A simple alternative to forcing valid colors could be to provide additional conversion methods. Something like:

```rust
fn try_convert<T: From<Self> + Limited>(self) -> Option<T>;
fn convert_clamped<T: From<Self> + Limited>(self) -> T;
```

If we, on the other hand, would turn the table and force `From` conversion to be checked, what would then be the best way? Panicking isn't nice, since it may come as an unpleasant surprise, because of float inaccuracies. Returning an option will introduce an extra step where the color has to be unwrapped in some way, and suffers from the same problem as panicking. I'm also not sure if we are even allowed to implement `From` for `Option`, since we are playing with type parameters. That leaves us with clamping, which is limiting, but the colors will be within range.

A possible middle ground could be to only clamp values that are _definitely_ invalid in almost every case, like negative RGB. That would still allow brighter than white and such, which is important in applications like 3D rendering.

Ogeon:
Still a bit fuzzy. Adding it to the wish list for now.

justinpombrio:
[...]

PR #43, opened 2016-02-07
Add conversion trait

sidred:
This commit adds a FromColor and IntoColor conversion trait. These traits require from_xyz and into_xyz respectively to be implemented and auto derive the conversions for other colors. 

To implement FromColor and IntoColor for a type, a few things must by kept in mind.
- Derived colors like Hsl must override the default from_rgb  and use this for from_xyz for efficiency.

- The FromColor for the same color must override the default. For example, from_rgb for Rgb will convert via Xyz which needs to be overridden with self to avoid the unnecessary conversion.

- If a direct transform exists between the colors, then it should be overridden in both the colors. For example, Luma has direct conversion to Rgb. So from_rgb for Luma and from_luma for Rgb override the defaults.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -1,0 +1,231 @@
+use num::Float;
+
+use {Alpha, Rgb, Luma, Xyz, Yxy, Lab, Lch, Hsv, Hsl, Color};
+
+///FromColor provides conversion between the colors.
+///
+///It requires from_xyz and derives conversion to other colors as a default from this.
+///These defaults must be overridden when direct conversion exists between colors.
+///For example, Luma has direct conversion to Rgb. So from_rgb conversion for Luma and
+///from_luma for Rgb is implemented directly. The from for the same color must override
+///the default. For example, from_rgb for Rgb will convert via Xyz which needs to be overridden
+///with self to avoid the unnecessary converison.
+pub trait FromColor<T>: Sized
+    where T: Float,
+{
+    ///Convert from XYZ color space
+    fn from_xyz(Xyz<T>) -> Self;
+
+    ///Convert from Yxy color space
+    fn from_yxy(inp: Yxy<T>) -> Self {
+        Self::from_xyz(inp.into_xyz())
+    }
+
+    ///Convert from L*a*b* color space
+    fn from_lab(inp: Lab<T>) -> Self {
+        Self::from_xyz(inp.into_xyz())
+    }
+
+    ///Convert from L*C*h° color space
+    fn from_lch(inp: Lch<T>) -> Self {
+        Self::from_lab(inp.into_lab())
+    }
+
+    ///Convert from RGB color space
+    fn from_rgb(inp: Rgb<T>) -> Self {
+        Self::from_xyz(inp.into_xyz())
+    }
+
+    ///Convert from HSL color space
+    fn from_hsl(inp: Hsl<T>) -> Self {
+        Self::from_rgb(inp.into_rgb())
+    }
+
+    ///Convert from HSV color space
+    fn from_hsv(inp: Hsv<T>) -> Self {
+        Self::from_rgb(inp.into_rgb())
+    }
+
+    ///Convert from Luma
+    fn from_luma(inp: Luma<T>) -> Self {
+        Self::from_xyz(inp.into_xyz())
+    }
+
+}
+
+
+///IntoColor provides conversion between the colors.
+///
+///It requires into into_xyz and derives conversion to other colors as a default from this.
+///These defaults must be overridden when direct conversion exists between colors.
+pub trait IntoColor<T>: Sized
+    where T: Float,
+{
+
+    ///Convert into XYZ space
+    fn into_xyz(self) -> Xyz<T>;
+
+    ///Convert into Yxy color space
+    fn into_yxy(self) -> Yxy<T> {
+        Yxy::from_xyz(self.into_xyz())
+    }
+
+    ///Convert into L*a*b* color space
+    fn into_lab(self) -> Lab<T> {
+        Lab::from_xyz(self.into_xyz())
+    }
+
+    ///Convert into L*C*h° color space
+    fn into_lch(self) -> Lch<T> {
+        Lch::from_lab(self.into_lab())
+    }
+
+    ///Convert into RGB color space.
+    fn into_rgb(self) -> Rgb<T> {
+        Rgb::from_xyz(self.into_xyz())
+    }
+
+    ///Convert into HSL color space
+    fn into_hsl(self) -> Hsl<T> {
+        Hsl::from_rgb(self.into_rgb())
+    }
+
+    ///Convert into HSV color space
+    fn into_hsv(self) -> Hsv<T> {
+        Hsv::from_rgb(self.into_rgb())
+    }
+
+    ///Convert into Luma
+    fn into_luma(self) -> Luma<T> {
+        Luma::from_xyz(self.into_xyz())
+    }
+
+}
+
+macro_rules! impl_into_color {
+    ($self_ty:ident, $from_fn: ident) => {
+        impl< T: Float > IntoColor<T> for $self_ty<T> {
+
+            fn into_xyz(self) -> Xyz<T> {
+                Xyz::$from_fn(self)
+            }
+
+            fn into_yxy(self) -> Yxy<T> {
+                Yxy::$from_fn(self)
+            }
+
+            fn into_lab(self) -> Lab<T> {
+                Lab::$from_fn(self)
+            }
+
+            fn into_lch(self) -> Lch<T> {
+                Lch::$from_fn(self)
+            }
+
+            fn into_rgb(self) -> Rgb<T> {
+                Rgb::$from_fn(self)
+            }
+
+            fn into_hsl(self) -> Hsl<T> {
+                Hsl::$from_fn(self)
+            }
+
+            fn into_hsv(self) -> Hsv<T> {
+                Hsv::$from_fn(self)
+            }
+
+            fn into_luma(self) -> Luma<T> {
+                Luma::$from_fn(self)
+            }
+
+        }
+
+    }
+}
+
+
+macro_rules! impl_from_trait {
+    (($self_ty:ident, $into_fn: ident) => {$($other:ident),+}) => (
+        impl<T: Float> From<Alpha<$self_ty<T>, T>> for $self_ty<T> {
+            fn from(color: Alpha<$self_ty<T>, T>) -> $self_ty<T> {
+                color.color
+            }
+        }
+
+        impl<T: Float> From<Color<T>> for $self_ty<T> {
+            fn from(color: Color<T>) -> $self_ty<T> {
+                match color {
+                    Color::Luma(c) => c.$into_fn(),
+                    Color::Rgb(c) => c.$into_fn(),
+                    Color::Xyz(c) => c.$into_fn(),
+                    Color::Yxy(c) => c.$into_fn(),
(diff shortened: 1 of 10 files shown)


Ogeon:
This depends on an unstable feature, so it won't work in stable and beta. It's an interesting idea and I would like to entertain it a bit more, but I'm not willing to drop stable and beta support for it.

That being said, any improvement that can minimize the amount of `impl From` spamming is most welcome. The good thing about the way `From` and `Into` works is that they are generic, so you don't have to have `from_*` and `into_*` for every single type. It's also trivial to integrate custom types and make them work like the other. The more specific conversion functions can still be added for extra convenience.

sidred:
I am not sure I understand what you are trying say regarding from_*,  into_* and custom types.

I have removed the negative impl trait (unstable feature) and used a macro instead. Do you think this is any better?

sidred:
One of the reasons I was looking at this was to avoid rewriting macros, for some of the upcoming changes regarding WhitePoint and RgbVariants. I guess we can't really avoid it.

I think there is an advantage in having a conversion trait and I am happy to make these changes if you see a benefit.

Ogeon:
Oh, wait, now I get it! I must have been too distracted... This is actually a very nice approach, and making ColorConvert public would probably be even nicer. I thought it was a replacement for `From` and `Into`, but I see that it's not.

sidred:
I have implemented the color trait and changed the default behaviour of xyz -> rgb to clamp. All color enum / alpha From traits implemented via macro

Ogeon:
I only skimmed through it, so far, but this looks very nice. One thing, though; wouldn't it be a non-breaking change if we didn't clamp? Also, why only clamp RGB? An other alternative for now, provided that it's not breaking, is to add generic `clamped_from<C>` and `clamped_to<C>` functions. That would keep status quo for now, and we can merge it under 0.2.1 instead of 0.3.0. The clamping behavior can be changed retroactively when a concrete decision has been made in #41. What do you think about that?

sidred:
Sounds good. I revert the clamping for Rgb.

Why only clamp rgb?
The way I see it is that Xyz, Yxy, Luv and Lab conversions do not need clamping. 
The Rgb , Hsl and Hsv do not need clamping for conversion between them and to the Cie colors.
Only cie colors to Rgb/Hsl/Hsv need clamping and all of these go through rgb.
Luma -> Rgb also will not need clamping.

sidred:
I think there was an error in the code as well. into_rgb clamped, but from_xyz for Rgb did not clamp. Need to think a bit more about it.

Ogeon:
You can still end up with strange values if you are working with strange values from the start, but that's for a later discussion. I'll look through it in a moment.

src/convert.rs
```diff
@@ -0,0 +1,192 @@
+use num::Float;
+
+use {Alpha, Rgb, Luma, Xyz, Yxy, Lab, Lch, Hsv, Hsl, Color};
+
+///Provides conversion between the colors.
+///Color Conver requires from and into xyz and derives others from this.
```
Ogeon: > `///Color Conver`

This looks like a mistake. I would also suggest swapping this line with the line below, to keep that as one paragraph, while removing this sentence from the trait list. I don't know if you have checked how things looks after running `cargo doc`...

src/convert.rs
```diff
@@ -0,0 +1,192 @@
...
+    //
+    //Third, if a direct transform exists between the colors, then it should be
+    //overridden in both the colors. For example, Luma has direct conversion to
+    //Rgb. So into_rgb and from_rgb conversion is implemented in Luma and into_luma
+    //and from_luma is overridden in Rgb to use the conversion from Luma.
+
+    ///Convert from XYZ color space
+    fn from_xyz(&Xyz<T>) -> Self;
+
+    ///Convert into xyz space
```
Ogeon: The capitalization of XYZ this comment and the one above is inconsistent with the rest. The conventions I try to follow is to write `Xyz`, `Lab`, etc. if I'm talking about the Rust types, and write XYZ, L*a*b*, etc. if I'm talking about the color spaces in general.

src/convert.rs
```diff
@@ -0,0 +1,192 @@
...
+        impl<T: Float> From<Alpha<$self_ty<T>, T>> for $self_ty<T> {
+            fn from(color: Alpha<$self_ty<T>, T>) -> $self_ty<T> {
+                color.color
+            }
+        }
+
+        impl<T: Float> From<Color<T>> for $self_ty<T> {
+            fn from(color: Color<T>) -> $self_ty<T> {
+                match color {
+                    Color::Luma(c) => c.$into_fn(),
```
Ogeon: Can't `$other` be used here?

sidred: It's not clear which line you are talking about.

Ogeon:
Great, that was all I could find :smile: The only thing, apart from those, that's bothering me a bit is that I can't decide which one of `&self` and `self` is the most appropriate. `&self` is obviously nice because it doesn't take ownership of the input, but it's limited because you can't deconstruct more complex types. `self` gives more freedom to the conversion implementation, but it takes ownership. All the standard operations and the `From` and `Into` traits takes ownership, but it's always possible to implement them for `&T`. Maybe we should change this to take `self`, after all. I'm leaning towards doing the same with the other trait, as well. What do you think?

src/convert.rs
Ogeon: Sorry, I meant rewriting the `Color::Luma(c) => c.$into_fn(),` lines to `$(Color::$other(c) => c.$into_fn()),+`.

sidred:
Regarding the ColorConvert trait, the code currently seems duplicated. For example into_yxy and from_yxy for Luma is implemented directly and I have to make sure that from_luma and into_luma should also be changed to use these conversions. Same for hsl <-> hsv conversion. 

I was also thinking about FromColor and IntoColor as separate traits, I think I got tripped up while trying to implement the IntoColor automatically for the FromColor. Separating the traits might be better though.

Another reason for From trait not to take &self is because this is a no-op
```
impl<T> From<T> for T {
    fn from(t: T) -> T { t }
}
```
Doing this might be expensive
```
impl<T> From<T> for T {
    fn from(t: &T) -> T { t.to_owned() }
}
```
But we have no such restriction. The color type is copy. So this should be pretty fast
```
    fn from_yxy(yxy: &Yxy<T>) -> Self {
        *yxy
    }
```
I think I now prefer &self. No need to  take ownership when not required.



I don't understand this part  "but it's limited because you can't deconstruct more complex types"

src/convert.rs
sidred: It will miss self type. Xyz as self_ty does not have Xyz as other type. Same for all other colors.

sidred:
```
impl<T: Float> From<Color<T>> for Alpha<$self_ty<T>,T>
```
can re use the

```
impl<T: Float> From<Color<T>> for $self_ty<T> {
```

No need to repeat the enum variants.

Ogeon:
> I don't understand this part "but it's limited because you can't deconstruct more complex types"

I was thinking about types where `Copy` isn't implemented. They could contain a `Box` or whatever, and `&self` would require them to be cloned behind the scenes. It doesn't matter for the built-in types, but it may become more complicated for third party types.

> Separating the traits might be better though.

That would be a good idea, I think. Some types may be possible to convert one way, but not the other. It could be destructive, semantically wrong, or whatever else.

sidred:
Let me take another crack at separating the from and into color traits.

src/convert.rs
Ogeon: Alright, let's keep it like this for now and see if we grow tired of it. I don't expect this list to grow an awful lot more, anyway.

Ogeon:
Go for it :+1: Just ask if you need an extra pair of eyes.

sidred:
Given FromColor and IntoColor
```
pub trait FromColor<T>: Sized
    where T: Float,
{...}


pub trait IntoColor<T>: Sized
    where T: Float,
{...}
```

How would you auto impl IntoColor for From 

```
impl<T:Float, U, V> IntoColor<U> for V where U: FromColor<V> {
```

Ogeon:
It would rather be something like

```rust 
impl<T:Float, C: FromColor<T>> IntoColor<T> for C {
```

but that's not right either, because something is missing. `From<T>` and `Into<T>` works because `T` is the type we are converting to and from. `FromColor` and `IntoColor` isn't generic over the "main type" but only its components, so I don't think it's possible to derive one from the other.

sidred:
I've used a macro to implement IntoColor trait. Not sure how sure how optimal this is. For example self conversion for into does this
```
  impl< T: Float > IntoColor<T> for Xyz<T> {

            fn into_xyz(&self) -> Xyz<T> {
                Xyz::from_xyz(self)
   }
```
instead of
```
  impl< T: Float > IntoColor<T> for Xyz<T> {

            fn into_xyz(&self) -> Xyz<T> {
               *self
   }
```

May be the compiler will optimize this away.


Is this any better than the ColorConvert trait?

Ogeon:
It's much better because it's divided into two parts and I think the compiler will inline those calls. It would have been possible to implement them in the same way as before, though, but just divided into two `impl`s.

I think we can just as well take everything by value, by the way. The implementor can still choose to limit the implementation to `&SomeColor`, so it won't ruin it for anyone. It will also make the trait names consistent with the Rust conventions, where "into" implies that it takes ownership of something.

Ogeon:
...It will also leave any copying to the user, instead of always forcing it if it's necessary.

src/convert.rs
```diff
@@ -0,0 +1,238 @@
+use num::Float;
+
+use {Alpha, Rgb, Luma, Xyz, Yxy, Lab, Lch, Hsv, Hsl, Color};
+
+///FromColor provides conversion between the colors. It requires from_xyz and derives conversion
```
Ogeon: Would you mind separating the first sentence into its own paragraph? It will look nicer in the docs.

src/convert.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    ///Convert from Luma
+    fn from_luma(inp: Luma<T>) -> Self {
+        Self::from_xyz(inp.into_xyz())
+    }
+
+}
+
+
+///IntoColor provides conversion between the colors. It requires into into_xyz
```
Ogeon: Same with this sentence.

src/convert.rs
```diff
@@ -0,0 +1,238 @@
...
+    //
+    //First, derived colors like Hsl must override from_rgb and convert from_xyz via Rgb.
+    //
+    //Second, the from for the same color must override the default. For example,
+    //from_rgb for Rgb will convert via Xyz which needs to be overridden with self to
+    //avoid the unnecessary converison.
+    //
+    //Third, if a direct transform exists between the colors, then it should be
+    //overridden. For example, Luma has direct conversion to Rgb. So from_rgb conversion
+    //for Luma and from_luma for Rgb is implemented directly.
```
Ogeon: These comments are a bit redundant now, when the same info can be found in the documentation comments.

sidred:
Not sure if anything else needs to be done here. I've made the recommended changes.

Ogeon:
No, not unless you want to add anything. Possibly those tests, but they can be added separately, as long as it's done before the release. The [WIP] tag is still there, so I thought you had something else in store, and then I got distracted by other things I had to do... The description is outdated, though.

sidred:
Let's close this and I'll add the tests separately.

Ogeon:
Alright :) Just edit the title and description, first.

sidred:
I didn't realize that you could edit the title :)

Ogeon:
It mentioned a "FromConvert" trait, but I changed that part just now.

@homu r+

PR #47, opened 2016-02-10
Change normalize to use floor and ceil formula 

### LINKED ISSUE #46: Change normalize function
```
fn normalize_angle_division(mut deg: f32) -> f32 {
    let c360 = 360.0_f32;
    deg - ((deg / c360).floor() * c360)

}
```
Faster than using mod and as fast as while. No slowdown for large numbers. Putting it here so I don't forget.

sidred:
Change normalize function to use floor and ceil based formula without loops. This is faster than using modulo operator and as fast as the existing while loop based solution. No slowdown for large numbers unlike while loop based solution. 

Closes #46

### DIFF
--- a/src/hues.rs
+++ b/src/hues.rs
@@ -132,36 +132,51 @@ make_hues! {
     struct RgbHue;
 }
 
-fn normalize_angle<T: Float>(mut deg: T) -> T {
-    let c180 = flt(180.0);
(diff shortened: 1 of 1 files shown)


Ogeon:
Nice, and thank you for adding some thorough tests!

@homu r+

PR #52, opened 2016-02-14
Add missing ApproxEq implementations

Ogeon:
#51 didn't implement `ApproxEq` for `Alpha` and `Color`, so this PR adds those implementations.

### DIFF
--- a/src/alpha.rs
+++ b/src/alpha.rs
@@ -2,6 +2,8 @@ use std::ops::{Deref, DerefMut, Add, Sub, Mul, Div};
 
 use num::Float;
 
+use approx::ApproxEq;
+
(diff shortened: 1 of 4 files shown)


Ogeon:
It's also missing for `Srgb` and `GammaRgb`, so I'll most likely implement it for those too. I don't see why not.

sidred:
Also equality.rs line 44, missing hue
```
impl_eq!( Lch, [l, chroma] );
```
Should be
```
impl_eq!( Lch, [l, chroma, hue] );
```

Ogeon:
Thanks, I'll add that one too.

Ogeon:
There we go. `ApproxEq` should now be implemented for every struct and enum where it's applicable.

Ogeon:
@homu r+

PR #53, opened 2016-02-15
Add hwb color

### LINKED ISSUE #32: Add support for HWB color model
I was reading through the CSS level 4 color module to see what's coming up and came across a new color model: [HWB](https://drafts.csswg.org/css-color-4/#the-hwb-notation). It was a total love at first sight. It works allowing you to select a hue and then mix in whiteness and blackness. 

Support for it in this library would be great.

sidred:
Added support for HWB color model. 

Added tests, implemented Alpha, ApproxEq, FromColor and IntoColor traits and added to color enum.

closes #32


sidred:
Did not implement Saturate for Hwb. Is it needed?

Ogeon:
Ah, nice! I'll look at this later, when my brain has received some nourishment.

> Did not implement Saturate for Hwb. Is it needed?

It shouldn't if it can't be done in a trivial way. I'm not exactly sure in this case, but maybe it's enough to scale both the B and W components.

src/hwb.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn from_hwb(hwb: Hwb<T>) -> Self {
+        hwb
+    }
+
+}
+
+impl<T: Float> Limited for Hwb<T> {
+    fn is_valid(&self) -> bool {
+        self.blackness >= T::zero() && self.blackness <= T::one() &&
+        self.whiteness >= T::zero() && self.whiteness <= T::one()
```
Ogeon: It's probably a good idea to have the sum check here, as well.

src/hwb.rs
```diff
@@ -0,0 +1,287 @@
...
+            blackness: self.blackness - amount,
+        }
+    }
+}
+
+impl<T: Float> GetHue for Hwb<T> {
+    type Hue = RgbHue<T>;
+
+    fn get_hue(&self) -> Option<RgbHue<T>> {
+        if self.whiteness == T::zero() || self.blackness == T::zero() {
```
Ogeon: Isn't this the other way around?

sidred: Lighten seems correct. Look at the colors in https://en.wikipedia.org/wiki/HWB_color_model . as whiteness -> 1, color becomes white.

Ogeon: My comment refers to line 132. The preview is misleading, but it's always the bottom line.

sidred: Yup

Ogeon: Wait, what? I'm talking about `get_hue`, not `lighten`. Returning `None` if `self.whiteness == T::zero() || self.blackness == T::zero()` is not right at all. That would make it return `None` along the left and upper edges of those color swatches. It should rather check for `self.whiteness + self.blackness >= 1.0`.

sidred: I've made it
```
 if self.whiteness == T::one() || self.blackness == T::one() {
```
Should they also be clamped before the check? It'll fail if whiteness = 1 and blackness = 4, which will clamp to 0.2 and 0.8?

Ogeon: `self.whiteness + self.blackness >= T::one()` feels more robust IMO. That's what we are checking for everywhere else.

sidred: > self.whiteness + self.blackness >= T::one() feels more robust IMO. That's what we are checking for everywhere else.

Thats not right. It'll fail for whiteness = 0.25 and blackness = 0.75, which is a valid hue. What we want to check is if w = 0 and b = 1 or the inverse when scaled.

Ogeon: How is it a valid hue if it's gray when whiteness + blackness = 1?

sidred: You are right. I misunderstood how whiteness and blackness work.

Ogeon: It's ok. I was starting to question my own view of it, but those swatches told me otherwise.

Ogeon:
I think that was all. We can leave out `Saturate` for now, since it's not completely trivial.

sidred:
I just derived the saturate formula from hsv as
```
hue_new = hue
w_new = w - (factor * ( 1 -w -b))
b_new = b
```
This will exactly match the saturation in Hsv from space. 

However it does not match the saturation from Hsl space

```
       // matches hwb_input.saturate(amt)
        hwb_input.into_hsv().saturate(amt).into_hwb();
       
       // does not match hwb_input.saturate(amt)
       hwb_input.into_hsl().saturate(amt).into_hwb();
       
        // These saturation in Hsl and Hsv space do not match either.
        let hsv1 = hwb_input.into_hsv().saturate(amt);
        let hsv2 = hwb_input.into_hsl().saturate(amt).into_hsv();
```

Not sure if saturation should be equivalent in the HSL and HSV spaces. 

Should I implement this saturation?

Ogeon:
HSL and HSV are affected by the difference between lightness and value. `Saturate` is currently only meant as a common interface for types where saturation/chroma is clearly and trivially defined and manipulated. HWB is a bit hazy, since both whiteness and blackness affects the saturation, depending on how you see it. That's why I'm not sure if it should even implement the trait at all, and it's probably better to convert the color to a space where the saturation has a better definition.

sidred:
Just wanted to confirm that the clamp_self check for 0 to infinity makes sense right? For all other colors we have checked 0-1, but here we are normalizing.

```

+    fn clamp_self(&mut self) {
+        self.whiteness = clamp(self.whiteness, T::zero(), T::infinity());
+        self.blackness = clamp(self.blackness, T::zero(), T::infinity());
+        let sum = self.blackness + self.whiteness;
+        if sum > T::one() {
+            self.whiteness = self.whiteness / sum;
+            self.blackness = self.blackness / sum;
+        }
+    }
```

Ogeon:
Yes, the normalization makes it ok. You can't fool it by making one parameter negative, since it would be clamped to 0. I just realized that `clamp` is unnecessary, since this is equal to `x.max(0)`.

sidred:
I updated Limited to use max fn instead of clamp.

Should be good to go!

Ogeon:
Great! I'll merge as soon as the tests are green. Thank you!

Ogeon:
@homu r+

PR #55, opened 2016-02-23
Add common color trait

Ogeon:
Add a common trait for all of the color types, which defines the most basic properties. Its current purpose is to provide the component scalar type, but it could also include other information in the future (such as the white point). The reason for doing this at all is that it removes the need for associated scalar types in operation traits, as well as the need for the additional `T` type parameter on `Alpha` and `PreAlpha`.

This is a very breaking change. It removes the `Scalar` type from a number of traits, and removes the `T` type parameter from `Alpha` and `PreAlpha`. An unfortunate side effect is that I had to define `ApproxEq::Epsilon` for `Alpha` and `PreAlpha` to be the same as the scalar type, instead of using the `Epsilon` of the scalar type. It will probably not matter, but it's still less generic.

### DIFF
--- a/src/alpha.rs
+++ b/src/alpha.rs
@@ -1,24 +1,28 @@
 use std::ops::{Deref, DerefMut, Add, Sub, Mul, Div};
 
-use num::Float;
+use num::{Float, One, Zero};
 
 use approx::ApproxEq;
 
-use {Mix, Shade, GetHue, Hue, Saturate, Limited, Blend, ComponentWise, clamp};
+use {Mix, Shade, GetHue, Hue, Saturate, Limited, Blend, ComponentWise, ColorType, clamp};
 use blend::PreAlpha;
 
 ///An alpha component wrapper for colors.
 #[derive(Clone, Copy, Debug, PartialEq)]
-pub struct Alpha<C, T: Float> {
+pub struct Alpha<C: ColorType> {
     ///The color.
     pub color: C,
 
     ///The transparency component. 0.0 is fully transparent and 1.0 is fully
     ///opaque.
-    pub alpha: T,
+    pub alpha: C::Scalar,
 }
 
-impl<C, T: Float> Deref for Alpha<C, T> {
+impl<C: ColorType> ColorType for Alpha<C> {
+    type Scalar = C::Scalar;
+}
+
+impl<C: ColorType> Deref for Alpha<C> {
     type Target = C;
 
     fn deref(&self) -> &C {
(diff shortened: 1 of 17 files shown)


Ogeon:
This is sort of a proposal and a proof of concept, so it's very open for changes. I got the idea and decided to try it out. It looks like things became mostly better, while some implementations for `Alpha` and `PreAlpha` became a bit more complex. A couple unanswered questions:

 * Should every operation trait require the basic color trait?
 * Is the name `ColorType` tolerable, or is there a better name?

sidred:
WhitePoint and GetRgbPrimaries will need to be separate traits as the color structs need to be bound by those traits.

Would it be better to merge the  mix, hue, saturate and limited traits to the ColorType?

Ogeon:
> WhitePoint and GetRgbPrimaries will need to be separate traits as the color structs need to be bound by those traits.

I'm not sure I follow. The RGB primaries are only relevant for RGB centric colors, so it should not be included. The white point, however, is relevant to every color type, so it could be included as an associated type. No matter if it's constant or not.

> Would it be better to merge the mix, hue, saturate and limited traits to the ColorType?

`Mix` and `Limited` could possibly be merged with it, but that would require every color to be mixable and limited. `Hue` and `Saturate` doesn't make sense for _every_ color, so they are better kept separate.

sidred:
> I'm not sure I follow. The RGB primaries are only relevant for RGB centric colors, so it should not be included. The white point, however, is relevant to every color type, so it could be included as an associated type. No matter if it's constant or not.

Yeah the primaries will only be relevant for the Rgb variants.

However, the  white point for a color is derived from the WhitePoint trait bound to the color. I am not sure if duplicating it on the ColorType trait adds any value.

```
impl ColorType for Lab<WP: WhitePoint, T> {
  fn get_white_point(...) -> ... {
     WP::get_white_point()
   }
 }
```


> Mix and Limited could possibly be merged with it, but that would require every color to be mixable and limited. Hue and Saturate doesn't make sense for every color, so they are better kept separate.

With the FromColor and IntoColor traits, we can implement these for all colors ( using into_rgb() for example). I guess the question is not whether we can implement it, but whether there is a need to implement for all colors. But thats a tangent to this issue :)

With just the scalar type parameter, I am not sure if the ColorType makes sense. What does the ColorTrait get you? For someone new reading the code, it adds another layer of abstraction without any obvious benefit.

Ogeon:
> I am not sure if duplicating it on the ColorType trait adds any value.

It will act as a common access point, just as it does with `Scalar`. It could look like this:

```rust
pub trait ColorType {
    type Scalar: Float;
    type WhitePoint: WhitePoint;
}
```

and be implemented like this:

```rust
impl<WP: WhitePoint, T: Float> ColorType for Lab<WP, T> {
    type Scalar = T;
    type WhitePoint = WP;
}
```

It could then be used instead of type parameters or associated types in other traits, just like with `Scalar`.

> With the FromColor and IntoColor traits, we can implement these for all colors ( using into_rgb() for example).

It will still not make sense to get the hue from a gray color, and I would generally prefer to not make assumptions like that. Anyway, `Limited` is universal enough to be merged into it. It was actually a similar trait before (called `ColorSpace`), but it was never used for anything more, so I renamed it to `Limited`. `Mix` is, well, I guess it should be ok to include that one too. I suppose the risk for strange non-mixable colors is minimal. I can make an attempt and see how it looks.

> What does the ColorTrait get you? For someone new reading the code, it adds another layer of abstraction without any obvious benefit.

I admit that the benefit is subtle in this state, but it's helpful to remove the extra type parameters on the alpha wrappers, and the associated types. There is always the risk of accidentally ending up with the useless `Alpha<Rgb<f32>, f64>`, and defining a trait that depends on both `Mix` and `Saturate` would previously require their `Scalar` types to be synced (I had a similar problem in `Blend`, before I simplified it).

I'm not in a hurry with this, so we can also put it on ice and see how it interacts with future changes.

Ogeon:
I just realized that `Color` doesn't implement `Limited`. It would depend on what space the inner color is in, and I'm not sure if that may be too surprising.

Ogeon:
This is so far behind... A new one can be created if it is ever needed.

PR #56, opened 2016-02-25
Make color spaces white point aware

### LINKED ISSUE #14: Make it possible to specify a custom white point
Some color conversions depends on a white point, and Palette is currently making some assumptions in this area. Avoiding these assumptions would of course allow a lot more freedom, so it would be nice to find a good way to customize which white point to use when converting. Even better if it was encoded into the color types, themselves, when relevant.

sidred:
This is a breaking change.
 
- All colors now have an additional trait bound on WhitePoint.
- All "new" methods on colors now assume a default white point of D65. Use "with_wp" method to get color with another white point.
- Delete tristimulus module.
- Add Xyz values for standard observers and illuminants (D65, D50 etc).
- Implement methods for chromatic adaptation using Bradford, VonKries and Xyz scaling.
- Add Srgb primaries for Rgb <-> Xyz converison.
- Make Rgb to luma conversion to convert via xyz. 

closes #14

### DIFF
--- a/src/blend/blend.rs
+++ b/src/blend/blend.rs
@@ -28,8 +28,9 @@ pub trait Blend: Sized {
     ///```
     ///use palette::{Rgb, Rgba, Blend};
     ///use palette::blend::PreAlpha;
+    ///use palette::white_point::D65;
     ///
-    ///type PreRgba = PreAlpha<Rgb<f32>, f32>;
+    ///type PreRgba = PreAlpha<Rgb<D65, f32>, f32>;
     ///
     ///fn blend_mode(a: PreRgba, b: PreRgba) -> PreRgba {
     ///    PreAlpha {
@@ -326,7 +327,7 @@ pub trait Blend: Sized {
                 } else if b * flt(4.0) <= dst.alpha {
                     let m2 = m * m;
                     let m3 = m2 * m;
-                    
+
                     dst.alpha * (two * a - src.alpha) * (m3 * flt(16.0) - m2 * flt(12.0) - m * flt(3.0)) + a - a * dst.alpha + b
                 } else {
                     dst.alpha * (two * a - src.alpha) * (m.sqrt() - m) + a - a * dst.alpha + b
--- a/src/chromatic_adaptation.rs
+++ b/src/chromatic_adaptation.rs
@@ -1,0 +1,261 @@
+//!Convert colors from one reference white point to another
+//!
+//!Chromatic adaptation is the human visual system’s ability to adjust to changes
+//!in illumination in order to preserve the appearance of object colors. It is responsible
+//!for the stable appearance of object colours despite the wide variation of light which
+//!might be reflected from an object and observed by our eyes.
+//!
+//!This library provides three methods for chromatic adaptation Bradford (which is the default),
+//!VonKries and XyzScaling
+//!
+//!```
+//!use palette::Xyz;
+//!use palette::white_point::{A, C};
+//!use palette::chromatic_adaptation::AdaptInto;
+//!
+//!
+//!let a = Xyz::<A, f32>::with_wp(0.315756, 0.162732, 0.015905);
+//!
+//!//Will convert Xyz<A, f32> to Xyz<C, f32> using Bradford chromatic adaptation
+//!let c: Xyz<C, f32> = a.adapt_into();
+//!
+//!//Should print {x: 0.257963, y: 0.139776,z: 0.058825}
+//!println!("{:?}", c)
+//!```
+use num::Float;
+
+use {Xyz, FromColor, IntoColor, flt};
+use white_point::WhitePoint;
+use matrix::{Mat3, multiply_xyz, multiply_3x3};
+
+
+///Chromatic adaptation methods implemented in the library
+pub enum  Method {
+    ///Bradford chromatic adaptation method
+    Bradford,
+    ///VonKries chromatic adaptation method
+    VonKries,
+    ///XyzScaling chromatic adaptation method
+    XyzScaling,
+}
+
+///Holds the matrix coeffecients for the chromatic adaptation methods
+pub struct ConeResponseMatrices<T: Float> {
+    ///3x3 matrix for the cone response domains
+    pub ma: Mat3<T>,
+    ///3x3 matrix for the inverse of the cone response domains
+    pub inv_ma: Mat3<T>,
+}
+
+///Generates a conversion matrix to convert the Xyz trisitmilus values from
+///one illuminant to another (Swp -> Dwp)
+pub trait TransformMatrix<Swp, Dwp, T>
+    where T: Float,
+        Swp: WhitePoint<T>,
+        Dwp: WhitePoint<T>,
+{
+    ///Get the cone response functions for the chromatic adaptation method
+    fn get_cone_response(&self) -> ConeResponseMatrices<T>;
+
+    ///Generates a 3x3 transformation matrix to convert color from one reference white point
+    ///to another with the given cone_response
+    fn generate_transform_matrix(&self) -> Mat3<T> {
+        let s_wp = Swp::get_xyz();
+        let t_wp = Dwp::get_xyz();
+        let adapt = self.get_cone_response();
+
+        let resp_src: Xyz<Swp,_> = multiply_xyz(&adapt.ma, &s_wp);
+        let resp_dst: Xyz<Dwp,_> = multiply_xyz(&adapt.ma, &t_wp);
+        let z = T::zero();
+        let resp = [resp_dst.x/resp_src.x, z , z , z, resp_dst.y/resp_src.y, z, z, z, resp_dst.z/ resp_src.z];
+
+        let tmp = multiply_3x3(&resp, &adapt.ma);
+        multiply_3x3(&adapt.inv_ma, &tmp)
+    }
+}
+
+
+impl<Swp, Dwp, T> TransformMatrix<Swp, Dwp, T> for Method
+    where T: Float,
+        Swp: WhitePoint<T>,
+        Dwp: WhitePoint<T>,
+{
+    fn get_cone_response(&self) -> ConeResponseMatrices<T> {
+        match *self {
+             Method::Bradford => {
+                ConeResponseMatrices::<T> {
+                    ma: [flt(0.8951000), flt(0.2664000), flt(-0.1614000),
+                         flt(-0.7502000), flt(1.7135000), flt(0.0367000),
+                         flt(0.0389000), flt(-0.0685000), flt(1.0296000)
+                         ],
+                    inv_ma: [flt(0.9869929), flt(-0.1470543), flt(0.1599627),
+                             flt(0.4323053), flt(0.5183603), flt(0.0492912),
+                             flt(-0.0085287), flt(0.0400428), flt(0.9684867)
+                            ],
+                }
+            }
+             Method::VonKries => {
+                ConeResponseMatrices::<T> {
+                    ma: [flt(0.4002400), flt(0.7076000), flt(-0.0808100),
+                         flt(-0.2263000), flt(1.1653200), flt(0.0457000),
+                         flt(0.0000000), flt(0.0000000), flt(0.9182200)
+                         ],
+                    inv_ma: [flt(1.8599364), flt(-1.1293816), flt(0.2198974),
+                             flt(0.3611914), flt(0.6388125), flt(-0.0000064),
+                             flt(0.0000000), flt(0.0000000), flt(1.0890636)
+                             ],
+                }
+            }
+             Method::XyzScaling => {
+                ConeResponseMatrices::<T> {
+                    ma: [flt(1.0000000), flt(0.0000000), flt(0.0000000),
+                         flt(0.0000000), flt(1.0000000), flt(0.0000000),
+                         flt(0.0000000), flt(0.0000000), flt(1.0000000)
+                         ],
+                    inv_ma: [flt(1.0000000), flt(0.0000000), flt(0.0000000),
+                             flt(0.0000000), flt(1.0000000), flt(0.0000000),
+                             flt(0.0000000), flt(0.0000000), flt(1.0000000)
+                             ],
+                }
+            }
+        }
+    }
+}
+
+
+///Trait to convert color from one reference white point to another
+///
+///Converts a color from the source white point (Swp) to the destination white point (Dwp).
+///Uses the bradford method for conversion by default.
+pub trait AdaptFrom<S, Swp, Dwp, T>: Sized
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+{
+    ///Convert the source color to the destination color using the bradford method by default
+    fn adapt_from(color: S) -> Self {
+        Self::adapt_from_using(color, Method::Bradford)
+    }
+    ///Convert the source color to the destination color using the specified method
+    fn adapt_from_using<M: TransformMatrix<Swp, Dwp, T>>(color: S, method: M) -> Self;
+}
+
+impl<S, D, Swp, Dwp, T> AdaptFrom<S, Swp, Dwp, T> for D
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+          S: IntoColor<Swp, T>,
+          D: FromColor<Dwp, T>,
+
+{
+    fn adapt_from_using<M: TransformMatrix<Swp, Dwp, T>>(color:S, method: M) -> D {
+        let src_xyz: Xyz<Swp, T> = color.into_xyz();
+        let transform_matrix = method.generate_transform_matrix();
+        let dst_xyz: Xyz<Dwp, T> = multiply_xyz(&transform_matrix, &src_xyz);
+        D::from_xyz(dst_xyz)
+    }
+}
+
+///Trait to convert color with one reference white point into another
+///
+///Converts a color with the source white point (Swp) into the destination white point (Dwp).
+///Uses the bradford method for conversion by default.
+pub trait AdaptInto<D, Swp, Dwp, T>: Sized
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+{
+    ///Convert the source color to the destination color using the bradford method by default
+    fn adapt_into(self) -> D {
+        self.adapt_into_using(Method::Bradford)
+    }
+    ///Convert the source color to the destination color using the specified method
+    fn adapt_into_using<M: TransformMatrix<Swp, Dwp, T>>(self, method: M) -> D;
+}
+
+impl<S, D, Swp, Dwp, T> AdaptInto<D, Swp, Dwp, T> for S
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+          D: AdaptFrom<S, Swp, Dwp, T>
+
+{
+    fn adapt_into_using<M: TransformMatrix<Swp, Dwp, T>>(self, method: M) -> D {
+        D::adapt_from_using(self, method)
+    }
+}
+
+
+
+
+#[cfg(test)]
+mod test {
+
+    use {Xyz};
+    use white_point::{D65, D50, A, C};
+    use super::{Method, TransformMatrix, AdaptInto, AdaptFrom};
+
+    #[test]
+    fn d65_to_d50_matrix_xyz_scaling() {
+        let expected = [1.0144665, 0.0000000, 0.0000000, 0.0000000, 1.0000000, 0.0000000, 0.0000000, 0.0000000, 0.7578869];
+        let xyz_scaling =  Method::XyzScaling;
+        let computed = <TransformMatrix<D65, D50, _>>::generate_transform_matrix(&xyz_scaling);
+        for (e, c) in expected.iter().zip(computed.iter()) {
+            assert_relative_eq!(e, c, epsilon = 0.0001)
+        }
+    }
+    #[test]
+    fn d65_to_d50_matrix_von_kries() {
+        let expected = [1.0160803, 0.0552297, -0.0521326, 0.0060666, 0.9955661, -0.0012235, 0.0000000, 0.0000000, 0.7578869];
+        let von_kries =  Method::VonKries;
+        let computed = <TransformMatrix<D65, D50, _>>::generate_transform_matrix(&von_kries);
+        for (e, c) in expected.iter().zip(computed.iter()) {
+            assert_relative_eq!(e, c, epsilon = 0.0001)
+        }
+    }
+    #[test]
+    fn d65_to_d50_matrix_bradford() {
+        let expected = [1.0478112, 0.0228866, -0.0501270, 0.0295424, 0.9904844, -0.0170491, -0.0092345, 0.0150436, 0.7521316];
+        let bradford =  Method::Bradford;
+        let computed = <TransformMatrix<D65, D50, _>>::generate_transform_matrix(&bradford);
+        for (e, c) in expected.iter().zip(computed.iter()) {
+            assert_relative_eq!(e, c, epsilon = 0.0001)
+        }
+    }
+
+    #[test]
+    fn chromatic_adaptation_from_a_to_c() {
+        let input_a = Xyz::<A, f32>::with_wp(0.315756, 0.162732, 0.015905);
+
+        let expected_bradford = Xyz::<C, f32>::with_wp(0.257963, 0.139776, 0.058825);
+        let expected_vonkries = Xyz::<C, f32>::with_wp(0.268446, 0.159139, 0.052843);
+        let expected_xyz_scaling = Xyz::<C, f32>::with_wp(0.281868, 0.162732, 0.052844);
+
+        let computed_bradford: Xyz<C, f32> = Xyz::adapt_from(input_a);
+        assert_relative_eq!(expected_bradford, computed_bradford, epsilon = 0.0001);
+
+        let computed_vonkries: Xyz<C, f32> = Xyz::adapt_from_using(input_a, Method::VonKries);
+        assert_relative_eq!(expected_vonkries, computed_vonkries, epsilon = 0.0001);
+
+        let computed_xyz_scaling: Xyz<C,_> = Xyz::adapt_from_using(input_a, Method::XyzScaling);
+        assert_relative_eq!(expected_xyz_scaling, computed_xyz_scaling, epsilon = 0.0001);
+    }
+
+    #[test]
+    fn chromatic_adaptation_into_a_to_c() {
+        let input_a = Xyz::<A, f32>::with_wp(0.315756, 0.162732, 0.015905);
+
+        let expected_bradford = Xyz::<C, f32>::with_wp(0.257963, 0.139776, 0.058825);
+        let expected_vonkries = Xyz::<C, f32>::with_wp(0.268446, 0.159139, 0.052843);
+        let expected_xyz_scaling = Xyz::<C, f32>::with_wp(0.281868, 0.162732, 0.052844);
+
+        let computed_bradford: Xyz<C, f32> = input_a.adapt_into();
+        assert_relative_eq!(expected_bradford, computed_bradford, epsilon = 0.0001);
+
+        let computed_vonkries: Xyz<C, f32> = input_a.adapt_into_using(Method::VonKries);
+        assert_relative_eq!(expected_vonkries, computed_vonkries, epsilon = 0.0001);
+
+        let computed_xyz_scaling: Xyz<C,_> = input_a.adapt_into_using(Method::XyzScaling);
+        assert_relative_eq!(expected_xyz_scaling, computed_xyz_scaling, epsilon = 0.0001);
+    }
+}
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -1,6 +1,7 @@
 use num::Float;
 
 use {Alpha, Rgb, Luma, Xyz, Yxy, Lab, Lch, Hsv, Hwb, Hsl, Color};
+use white_point::{WhitePoint, D65};
 
 ///FromColor provides conversion between the colors.
 ///
@@ -10,49 +11,50 @@ use {Alpha, Rgb, Luma, Xyz, Yxy, Lab, Lch, Hsv, Hwb, Hsl, Color};
 ///from_luma for Rgb is implemented directly. The from for the same color must override
 ///the default. For example, from_rgb for Rgb will convert via Xyz which needs to be overridden
 ///with self to avoid the unnecessary converison.
-pub trait FromColor<T>: Sized
+pub trait FromColor<Wp = D65, T = f32>: Sized
     where T: Float,
+        Wp: WhitePoint<T>
 {
     ///Convert from XYZ color space
-    fn from_xyz(Xyz<T>) -> Self;
+    fn from_xyz(Xyz<Wp, T>) -> Self;
 
     ///Convert from Yxy color space
-    fn from_yxy(inp: Yxy<T>) -> Self {
+    fn from_yxy(inp: Yxy<Wp, T>) -> Self {
         Self::from_xyz(inp.into_xyz())
     }
 
     ///Convert from L*a*b* color space
-    fn from_lab(inp: Lab<T>) -> Self {
+    fn from_lab(inp: Lab<Wp, T>) -> Self {
         Self::from_xyz(inp.into_xyz())
     }
 
     ///Convert from L*C*h° color space
-    fn from_lch(inp: Lch<T>) -> Self {
+    fn from_lch(inp: Lch<Wp, T>) -> Self {
         Self::from_lab(inp.into_lab())
     }
 
     ///Convert from RGB color space
-    fn from_rgb(inp: Rgb<T>) -> Self {
+    fn from_rgb(inp: Rgb<Wp, T>) -> Self {
         Self::from_xyz(inp.into_xyz())
     }
 
     ///Convert from HSL color space
-    fn from_hsl(inp: Hsl<T>) -> Self {
+    fn from_hsl(inp: Hsl<Wp, T>) -> Self {
         Self::from_rgb(inp.into_rgb())
     }
 
     ///Convert from HSV color space
-    fn from_hsv(inp: Hsv<T>) -> Self {
+    fn from_hsv(inp: Hsv<Wp, T>) -> Self {
         Self::from_rgb(inp.into_rgb())
     }
 
     ///Convert from HWB color space
-    fn from_hwb(inp: Hwb<T>) -> Self {
+    fn from_hwb(inp: Hwb<Wp, T>) -> Self {
         Self::from_hsv(inp.into_hsv())
     }
 
     ///Convert from Luma
-    fn from_luma(inp: Luma<T>) -> Self {
+    fn from_luma(inp: Luma<Wp, T>) -> Self {
         Self::from_xyz(inp.into_xyz())
     }
 
@@ -63,50 +65,51 @@ pub trait FromColor<T>: Sized
 ///
 ///It requires into into_xyz and derives conversion to other colors as a default from this.
 ///These defaults must be overridden when direct conversion exists between colors.
-pub trait IntoColor<T>: Sized
+pub trait IntoColor<Wp = D65, T = f32>: Sized
     where T: Float,
+     Wp: WhitePoint<T>
 {
 
     ///Convert into XYZ space
-    fn into_xyz(self) -> Xyz<T>;
+    fn into_xyz(self) -> Xyz<Wp, T>;
 
     ///Convert into Yxy color space
-    fn into_yxy(self) -> Yxy<T> {
+    fn into_yxy(self) -> Yxy<Wp, T> {
         Yxy::from_xyz(self.into_xyz())
     }
 
     ///Convert into L*a*b* color space
-    fn into_lab(self) -> Lab<T> {
+    fn into_lab(self) -> Lab<Wp, T> {
         Lab::from_xyz(self.into_xyz())
     }
 
     ///Convert into L*C*h° color space
-    fn into_lch(self) -> Lch<T> {
+    fn into_lch(self) -> Lch<Wp, T> {
         Lch::from_lab(self.into_lab())
     }
 
     ///Convert into RGB color space.
-    fn into_rgb(self) -> Rgb<T> {
+    fn into_rgb(self) -> Rgb<Wp, T> {
         Rgb::from_xyz(self.into_xyz())
     }
 
     ///Convert into HSL color space
-    fn into_hsl(self) -> Hsl<T> {
+    fn into_hsl(self) -> Hsl<Wp, T> {
         Hsl::from_rgb(self.into_rgb())
     }
 
     ///Convert into HSV color space
-    fn into_hsv(self) -> Hsv<T> {
+    fn into_hsv(self) -> Hsv<Wp, T> {
         Hsv::from_rgb(self.into_rgb())
     }
 
     ///Convert into HWB color space
-    fn into_hwb(self) -> Hwb<T> {
+    fn into_hwb(self) -> Hwb<Wp, T> {
         Hwb::from_hsv(self.into_hsv())
     }
 
     ///Convert into Luma
-    fn into_luma(self) -> Luma<T> {
+    fn into_luma(self) -> Luma<Wp, T> {
         Luma::from_xyz(self.into_xyz())
     }
 
@@ -114,37 +117,40 @@ pub trait IntoColor<T>: Sized
 
 macro_rules! impl_into_color {
     ($self_ty:ident, $from_fn: ident) => {
-        impl< T: Float > IntoColor<T> for $self_ty<T> {
+        impl<Wp, T> IntoColor<Wp, T> for $self_ty<Wp, T>
+            where T: Float,
+             Wp: WhitePoint<T>
+        {
 
-            fn into_xyz(self) -> Xyz<T> {
+            fn into_xyz(self) -> Xyz<Wp, T> {
                 Xyz::$from_fn(self)
             }
 
-            fn into_yxy(self) -> Yxy<T> {
+            fn into_yxy(self) -> Yxy<Wp, T> {
                 Yxy::$from_fn(self)
             }
 
-            fn into_lab(self) -> Lab<T> {
+            fn into_lab(self) -> Lab<Wp, T> {
                 Lab::$from_fn(self)
             }
 
-            fn into_lch(self) -> Lch<T> {
+            fn into_lch(self) -> Lch<Wp, T> {
                 Lch::$from_fn(self)
             }
 
-            fn into_rgb(self) -> Rgb<T> {
+            fn into_rgb(self) -> Rgb<Wp, T> {
                 Rgb::$from_fn(self)
             }
 
-            fn into_hsl(self) -> Hsl<T> {
+            fn into_hsl(self) -> Hsl<Wp, T> {
                 Hsl::$from_fn(self)
             }
 
-            fn into_hsv(self) -> Hsv<T> {
+            fn into_hsv(self) -> Hsv<Wp, T> {
                 Hsv::$from_fn(self)
             }
 
-            fn into_luma(self) -> Luma<T> {
+            fn into_luma(self) -> Luma<Wp, T> {
                 Luma::$from_fn(self)
             }
 
@@ -155,15 +161,21 @@ macro_rules! impl_into_color {
 
 
 macro_rules! impl_from_trait {
-    (($self_ty:ident, $into_fn: ident) => {$($other:ident),+}) => (
-        impl<T: Float> From<Alpha<$self_ty<T>, T>> for $self_ty<T> {
-            fn from(color: Alpha<$self_ty<T>, T>) -> $self_ty<T> {
+    (($self_ty: ident, $into_fn: ident) => {$($other: ident),+}) => (
+        impl<Wp, T> From<Alpha<$self_ty<Wp, T>, T>> for $self_ty<Wp, T>
+            where T: Float,
+                Wp: WhitePoint<T>
+        {
+            fn from(color: Alpha<$self_ty<Wp, T>, T>) -> $self_ty<Wp, T> {
                 color.color
             }
         }
 
-        impl<T: Float> From<Color<T>> for $self_ty<T> {
-            fn from(color: Color<T>) -> $self_ty<T> {
+        impl<Wp, T> From<Color<Wp, T>> for $self_ty<Wp, T>
+            where T: Float,
+                Wp: WhitePoint<T>
+        {
+            fn from(color: Color<Wp, T>) -> $self_ty<Wp, T> {
                 match color {
                     Color::Luma(c) => c.$into_fn(),
                     Color::Rgb(c) => c.$into_fn(),
@@ -178,8 +190,11 @@ macro_rules! impl_from_trait {
             }
         }
 
-        impl<T: Float> From<Color<T>> for Alpha<$self_ty<T>,T> {
-            fn from(color: Color<T>) -> Alpha<$self_ty<T>,T> {
+        impl<Wp, T> From<Color<Wp, T>> for Alpha<$self_ty<Wp, T>,T>
+            where T: Float,
+                Wp: WhitePoint<T>
+        {
+            fn from(color: Color<Wp, T>) -> Alpha<$self_ty<Wp, T>,T> {
                 Alpha {
                     color: color.into(),
                     alpha: T::one(),
@@ -187,8 +202,11 @@ macro_rules! impl_from_trait {
             }
         }
 
-        impl<T: Float> From<Alpha<Color<T>, T>> for Alpha<$self_ty<T>,T> {
-            fn from(color: Alpha<Color<T>, T>) -> Alpha<$self_ty<T>,T> {
+        impl<Wp, T> From<Alpha<Color<Wp, T>, T>> for Alpha<$self_ty<Wp, T>,T>
+            where T: Float,
+                Wp: WhitePoint<T>
+        {
+            fn from(color: Alpha<Color<Wp, T>, T>) -> Alpha<$self_ty<Wp, T>,T> {
                 Alpha {
                     color: color.color.into(),
                     alpha: color.alpha,
@@ -197,14 +215,20 @@ macro_rules! impl_from_trait {
         }
 
         $(
-            impl<T: Float> From<$other<T>> for $self_ty<T> {
-                fn from(other: $other<T>) -> $self_ty<T> {
+            impl<Wp, T> From<$other<Wp, T>> for $self_ty<Wp, T>
+                where T: Float,
+                    Wp: WhitePoint<T>
+            {
+                fn from(other: $other<Wp, T>) -> $self_ty<Wp, T> {
                     other.$into_fn()
                 }
             }
 
-            impl<T: Float> From<Alpha<$other<T>, T>> for Alpha<$self_ty<T>, T> {
-                fn from(other: Alpha<$other<T>, T>) -> Alpha<$self_ty<T>, T> {
+            impl<Wp, T> From<Alpha<$other<Wp, T>, T>> for Alpha<$self_ty<Wp, T>, T>
+                where T: Float,
+                    Wp: WhitePoint<T>
+            {
+                fn from(other: Alpha<$other<Wp, T>, T>) -> Alpha<$self_ty<Wp, T>, T> {
                     Alpha {
                         color: other.color.$into_fn(),
                         alpha: other.alpha,
@@ -212,8 +236,11 @@ macro_rules! impl_from_trait {
                 }
             }
 
-            impl<T: Float> From<$other<T>> for Alpha<$self_ty<T>, T> {
-                fn from(color: $other<T>) -> Alpha<$self_ty<T>, T> {
+            impl<Wp, T> From<$other<Wp, T>> for Alpha<$self_ty<Wp, T>, T>
+                where T: Float,
+                    Wp: WhitePoint<T>
+            {
+                fn from(color: $other<Wp, T>) -> Alpha<$self_ty<Wp, T>, T> {
                     Alpha {
                         color: color.$into_fn(),
                         alpha: T::one(),
@@ -221,8 +248,11 @@ macro_rules! impl_from_trait {
                 }
             }
 
-            impl<T: Float> From<Alpha<$other<T>, T>> for $self_ty<T> {
-                fn from(other: Alpha<$other<T>, T>) -> $self_ty<T> {
+            impl<Wp, T> From<Alpha<$other<Wp, T>, T>> for $self_ty<Wp, T>
+                where T: Float,
+                    Wp: WhitePoint<T>
+            {
+                fn from(other: Alpha<$other<Wp, T>, T>) -> $self_ty<Wp, T> {
                     other.color.$into_fn()
                 }
             }
@@ -231,23 +261,23 @@ macro_rules! impl_from_trait {
     )
 }
 
-impl_into_color!(Xyz,from_xyz);
-impl_into_color!(Yxy,from_yxy);
-impl_into_color!(Lab,from_lab);
-impl_into_color!(Lch,from_lch);
-impl_into_color!(Rgb,from_rgb);
-impl_into_color!(Hsl,from_hsl);
-impl_into_color!(Hsv,from_hsv);
-impl_into_color!(Hwb,from_hwb);
-impl_into_color!(Luma,from_luma);
+impl_into_color!(Xyz, from_xyz);
+impl_into_color!(Yxy, from_yxy);
+impl_into_color!(Lab, from_lab);
+impl_into_color!(Lch, from_lch);
+impl_into_color!(Rgb, from_rgb);
+impl_into_color!(Hsl, from_hsl);
+impl_into_color!(Hsv, from_hsv);
+impl_into_color!(Hwb, from_hwb);
+impl_into_color!(Luma ,from_luma);
 
 
-impl_from_trait!((Xyz,into_xyz) => {Yxy, Lab, Lch, Rgb, Hsl, Hsv, Hwb, Luma});
-impl_from_trait!((Yxy,into_yxy) => {Xyz, Lab, Lch, Rgb, Hsl, Hsv, Hwb, Luma});
-impl_from_trait!((Lab,into_lab) => {Xyz, Yxy, Lch, Rgb, Hsl, Hsv, Hwb, Luma});
-impl_from_trait!((Lch,into_lch) => {Xyz, Yxy, Lab, Rgb, Hsl, Hsv, Hwb, Luma});
-impl_from_trait!((Rgb,into_rgb) => {Xyz, Yxy, Lab, Lch, Hsl, Hsv, Hwb, Luma});
-impl_from_trait!((Hsl,into_hsl) => {Xyz, Yxy, Lab, Lch, Rgb, Hsv, Hwb, Luma});
-impl_from_trait!((Hsv,into_hsv) => {Xyz, Yxy, Lab, Lch, Rgb, Hsl, Hwb, Luma});
-impl_from_trait!((Hwb,into_hwb) => {Xyz, Yxy, Lab, Lch, Rgb, Hsl, Hsv, Luma});
-impl_from_trait!((Luma,into_luma) => {Xyz, Yxy, Lab, Lch, Rgb, Hsl, Hsv, Hwb});
+impl_from_trait!((Xyz, into_xyz) => {Yxy, Lab, Lch, Rgb, Hsl, Hsv, Hwb, Luma});
+impl_from_trait!((Yxy, into_yxy) => {Xyz, Lab, Lch, Rgb, Hsl, Hsv, Hwb, Luma});
+impl_from_trait!((Lab, into_lab) => {Xyz, Yxy, Lch, Rgb, Hsl, Hsv, Hwb, Luma});
+impl_from_trait!((Lch, into_lch) => {Xyz, Yxy, Lab, Rgb, Hsl, Hsv, Hwb, Luma});
+impl_from_trait!((Rgb, into_rgb) => {Xyz, Yxy, Lab, Lch, Hsl, Hsv, Hwb, Luma});
+impl_from_trait!((Hsl, into_hsl) => {Xyz, Yxy, Lab, Lch, Rgb, Hsv, Hwb, Luma});
+impl_from_trait!((Hsv, into_hsv) => {Xyz, Yxy, Lab, Lch, Rgb, Hsl, Hwb, Luma});
+impl_from_trait!((Hwb, into_hwb) => {Xyz, Yxy, Lab, Lch, Rgb, Hsl, Hsv, Luma});
+impl_from_trait!((Luma, into_luma) => {Xyz, Yxy, Lab, Lch, Rgb, Hsl, Hsv, Hwb});
--- a/src/gradient.rs
+++ b/src/gradient.rs
@@ -375,7 +375,8 @@ mod test {
 
         let v1: Vec<_> = g1.take(10).take(5).collect();
         let v2: Vec<_> = g2.take(5).collect();
-
-        assert_eq!(v1, v2);
+        for (t1, t2) in v1.iter().zip(v2.iter()){
+            assert_relative_eq!(t1, t2);
+        }
     }
 }
--- a/src/lab.rs
+++ b/src/lab.rs
@@ -1,15 +1,15 @@
 use num::Float;
 
 use std::ops::{Add, Sub, Mul, Div};
+use std::marker::PhantomData;
 
 use {Alpha, Xyz, Lch, LabHue};
 use {Limited, Mix, Shade, GetHue, FromColor, ComponentWise};
 use {clamp, flt};
-
-use tristimulus::{X_N, Y_N, Z_N};
+use white_point::{WhitePoint, D65};
 
 ///CIE L*a*b* (CIELAB) with an alpha component. See the [`Laba` implementation in `Alpha`](struct.Alpha.html#Laba).
-pub type Laba<T = f32> = Alpha<Lab<T>, T>;
+pub type Laba<Wp, T = f32> = Alpha<Lab<Wp, T>, T>;
 
 ///The CIE L*a*b* (CIELAB) color space.
 ///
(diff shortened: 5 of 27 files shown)


sidred:
#### To-Do:
- [x] Add white point
- [ ] Add chromatic adaptation conversions
- [ ] Add white point to color spaces

Ogeon:
Looks like a good start :+1: Just a few things I thought about:

 * The white point macro seems simple enough to not really be necessary. Will it become more complex?
 * The Rust conventions says that type names that are abbreviations should be camel cased as one word (`COW` -> `Cow`). Maybe we should do the same with the various variants of `WP`.
 * It's probably a good idea to make the `white_point` module public to not clutter the crate root too much.

sidred:
> The white point macro seems simple enough to not really be necessary. Will it become more complex?

Not really. Its just that I have to add a few more and right now its not clear what the correct x,y values are. I''l remove it in the end when I add the remaining white points.
> The Rust conventions says that type names that are abbreviations should be camel cased as one word (COW -> Cow). Maybe we should do the same with the various variants of WP.

D65Fov10 is already camel case. Are you talking about something else? Oh and if you can think of a better name, let me know :)

> It's probably a good idea to make the white_point module public to not clutter the crate root too much.

Ok

Ogeon:
> Not really. Its just that I have to add a few more and right now its not clear what the correct x,y values are. I''l remove it in the end when I add the remaining white points.

Ok, I was just checking. I have been thinking about phasing out some macros, in general, but that's a separate topic.

> D65Fov10 is already camel case. Are you talking about something else?

Those are correctly cased. I'm talking about the type parameters, like `WP` -> `Wp`, `SWP` -> `Swp`, and so on.

> Oh and if you can think of a better name, let me know :)

I'll see what I can come up with, but you have most likely witnessed how bad I am at naming things. The name "Palette" is a masterpiece among my creations ;) Speaking of which, you are always welcome to suggest alternate names for things I contribute with.

sidred:
> Those are correctly cased. I'm talking about the type parameters, like WP -> Wp, SWP -> Swp, and so on.

Yeah, those need to be changed. I thought trait bounds were all caps, but they need to be camel case too.

Ogeon:
They are temporary types, so it's the same rule for them as well, but they are usually only one letter long, so it's easy to forget :)

sidred:
I have implemeted the white point for lab and changed the into iterator from 

```
pub trait IntoColor<T>: Sized{
```

to 
```
pub trait IntoColor<WP = D65, T = f32>: Sized
    where T: Float,
     WP: WhitePoint<T>
{
```

I get an error  for  converting xyz to rgb
```
        let rgb: Rgb<T> = xyz.into_rgb();
        Self::from_rgb(rgb)
```
error
```
src/hsl.rs:63:31: 63:39 error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [E0282]
src/hsl.rs:63         let rgb: Rgb<T> = xyz.into_rgb();
                                            ^~~~~~~~

```
This works 
```
        let rgb = <Rgb<T>  as FromColor<WP, T>>::from_xyz(xyz);
        <Self as FromColor<WP, T>>::from_rgb(rgb)
```
Is there any other altenative?

Ogeon:
Is the white point of `xyz` known?

sidred:
For now I am not implementing the White Point for Xyz and Yxy. I'll implement it for everything else and only then add it in for Yxy and Xyz if it makes sense. 

I've been using the convert as notation for now. Let's see how it shapes up.

Ogeon:
Well, it can't possibly know what white is if it's not specified, so that's most likely the reason :)

sidred:
Tests still fail due to color enum. I will update it next.

Ogeon:
Why are you renaming almost every color to `*Space`?

sidred:
> Why are you renaming almost every color to *Space?


I was trying to keep the color spaces compatible with earlier versions and I have defined the original colors to use the default white point

```
pub type Lab<T> = LabSpace<D65, T>;
pub type Laba<T> = LabSpaceAlpha<D65, T>;
```

I am expecting this to work as usual before and after the white point change
```
   let lab1 = Lab::new(0.5, 0.5, 0.5)
   
   let lab2 = Lab {l: 0.5, a: 0.5, b:0.5}
```
Is it unnecessary since we are setting the defaults in the definition?
```
   pub struct LabSpace<Wp = D65, T = f32>
```

Ogeon:
Oh, ok. Well, most users should not be affected by the change, unless most users uses an other float type than f32. Adding the extra white point shouldn't be a huge step from changing the precision, and they can always make their own custom aliases when they need to.

I see the point and I understand that it may become tedious in some situations, but the fact that users can make their own aliases makes me think that it's probably not such a big deal, after all.

sidred:
I'll revert to using the short names. Having the defaults pretty much makes it unnecessary.

Ogeon:
We can do something about it later if it turns out to be a problem. There may even be a better solution then. Who knows?

You have probably noticed, but type aliases and the documentation don't really play nice together. You can't see any implementations for aliases, which is why I made the links for the alpha aliases. You can write `impl MyAlias { ... }` and it works fine, but it won't show up _anywhere_ in the documentation. It's a bit annoying and only somewhat related to this, so I'll stop there :)

sidred:
Tests still fail for Srgb, Gamma rgb and color enum conversion.

Ogeon:
I have mostly skimmed through the changes, but I think it looks good so far. The only things I can really comment on at the moment is that I'm leaning towards making the white point fields public, to allow `Rgb { red: ..., green: ... blue: ..., _wp: PhantomData }` style construction, and that I'm quite sure `Xyz` and `Yxy` should have a white point parameter.

You can't know what XYZ for white is if you haven't specified a white point, and the requirement of `<xyz as  as FromColor<WP, T>>::...` is an effect of that problem. We can't just throw away the reference to how the values are supposed to be interpreted. It would actually make more sense to do this with RGB, since we know that `rgb(1, 1, 1)` will always represent white, but the perception of an XYZ value differs depending on the reference white.

sidred:
The project builds, but tests fails due to private _wp field.

> The only things I can really comment on at the moment is that I'm leaning towards making the white point fields public

I was just thinking the same thing. I can't think of a downside to this. Is wp a good enough name or should I make it white_point (instead os _wp).

> I'm quite sure Xyz and Yxy should have a white point parameter.

I am not convinced that Xyz and Yxy need a white point.
Consider the white point ```(0.31271,0.32902, 1.0)```. This represents the D65 Illuminant for 2 degree observer. If we define the white point for this Yxy, then all following will be the same. The white point is essentially meaningless here.
```
Yxy<D65,_> == Yxy<D50,_> == Yxy<A,_> == (0.31271,0.32902, 1.0)
```
But the same is not the case for the other colors.
```
Rgb<D65,_> != Rgb<D50,_> != Rgb<A,_>
```

Also, The conversions Color -> Xyz will depend only on the source white_point and Xyz -> Color will only depend on the target white point and we should not run into any issues in the conversions.

Ogeon:
> I was just thinking the same thing. I can't think of a downside to this. Is wp a good enough name or should I make it white_point (instead os _wp).

`white_point` looks nicer to me.

> then all following will be the same. The white point is essentially meaningless here

We must not mix up perceptive and numeric equality. The `Yxy` values will have the same energy values, but they will look different under different illuminants, while the `Rgb` values will look the same, but will have different energy levels under different illuminants.

Take [this snowy house](http://dreamatico.com/data_images/snow/snow-6.jpg), for example. The snow looks mostly white because the colors are adapted to sunlight, but it's blue in the shadows, where it's only illuminated by the sky. We could adapt the colors to make the shadows look white/gray, but the rest would then look yellow. Both the sunlit and shadowy snow is still "white", though, since the snow will reflect more or less equal amounts of every wavelength.

Ogeon:
I'll continue my example with some code and numbers. Let's pretend that the sky itself is a `D50` illuminant and that the sun is a `D65` illuminant. Then the shadowy snow would be `Rgb::<D50, _>::new(1.0, 1.0, 1.0)`, which looks like purest white if our eyes adapt to the `D50` light, and the sunlit snow would be `Rgb::<D65, _>::new(1.0, 1.0, 1.0)`. The camera in the photo is adapted to the sunlight, or `D65` in this example, so we have to translate the color of the shadowy snow to `D65` to know how it will look in the photo:

```rust
let snow_shadow: Rgb<D50, _> = Rgb::new(1.0, 1.0, 1.0);
let snow_shadow_xyz: Xyz = snow_shadow.into();

//This is no longer numerically the same as Rgb::new(1.0, 1.0, 1.0),
//meaning that it will look different.
let snow_shadow_photo: Rgb<D65, _> = snow_shadow_xyz.into(); 
```

Let's say that we want to know if the shadowy snow reflects the same amount of light as the sunlit snow, and therefore should look the same under the same light. That's where we have to do a chromatic adaption:

```rust
let snow_shadow: Rgb<D50, _> = Rgb::new(1.0, 1.0, 1.0);
let snow_shadow_xyz: Xyz = snow_rgb.into();

//Adapt from D50 to D65, or move snow from the shadow into the sunlight
let snow_photo_xyz: Xyz = snow_shadow_xyz.adapt::<D50, D65>();

//This should now be equal to Rgb::new(1.0, 1.0, 1.0)
let snow_photo: Rgb<D65, _> = snow_photo_xyz.into();
```

The white point _does_ matter to XYZ, but not in the same way as to RGB. It's still important for knowing how the signals are perceived, while RGB needs it to know what the signals should be. This is the first example again, but with a white point aware `Xyz`:

```rust
let snow_rgb: Rgb<D50, _> = Rgb::new(1.0, 1.0, 1.0);
let snow_xyz: Xyz<D50, _> = snow_rgb.into();

//Adapt our brain to the sunlight, so the same numbers, but different white point.
let snow_shadow_photo_xyz: Xyz<D65, _> = snow_xyz.with_white_point();

//This is no longer numerically the same as Rgb::new(1.0, 1.0, 1.0),
//meaning that it will look different.
let snow_shadow_photo: Rgb<D65, _> = snow_shadow_photo_xyz.into(); 
```

...and the second example:

```rust
let snow_shadow: Rgb<D50, _> = Rgb::new(1.0, 1.0, 1.0);
let snow_shadow_xyz: Xyz<D50, _> = snow_rgb.into();

//Adapt from D50 to D65, or move snow from the shadow into the sunlight
let snow_photo_xyz: Xyz<D65, _> = snow_shadow_xyz.adapt();

//This should now be equal to Rgb::new(1.0, 1.0, 1.0)
let snow_photo: Rgb<D65, _> = snow_photo_xyz.into();
```

It's still possible to do the same calculations without the white point on `Xyz`, but you have to keep track if it yourself. Letting `Xyz` keep track of it instead will require an explicit transition between white points, but there's not the same risk of mixing things up. I feel like it just completes the whole model.

sidred:
Yeah, Xyz also needs an illuminant. For chromatic adaptation needing to do xyz_src to xyz_dst pretty makes this necessary. I'll make the change.

Right now the Color conversions do not take chromatic adaptation into account. They only convert from and into the same white point. 

What I am concerned about needing white point annotation in conversions. For example, rgb -> hsb should just work assuming the same white point without having to specify it. White Point annotation is only necessary when chormatic adpatation is needed. 


```
// assumes default D65
let rgb = Rgb::new(1.0, 0.0,0.0);

// Should automatically use D65 white point
let hsl = rgb.into_hsl() 
let lab = rgb.into_lab() // or even rgb.into()

// Instead of Manually having to specify whtie point
let hsl: Lab<D65,_> = rgb.into();
let lab: Lab<D65,_> = rgb.into();
```

Can we use the existing FromColor and IntoColor to make this happen? 
Should I add this in this PR or in a new PR?

sidred:
just run into a usablility issue. This code fails
```
let lab = Lab::new(0.4623,-0.517,0.499);
let lab = Lab {l: 0.4623, a: -0.517, b: 0.499, white_point: PhantomData };
```
with the following error even though the default (D65 and f32) are defined for the lab space.
```
 unable to infer enough type information about `_`; type annotations or generic parameter binding required
```
 This works

```
    let lab = Lab::<D65,f64> {l: 0.4623, a: -0.517, b: 0.499, white_point: PhantomData };
```

Can we do anything to infer the default white points?

src/lab.rs
```diff
@@ -1,15 +1,20 @@
...
+use ::white_point::{WhitePoint, D65};
 
 use tristimulus::{X_N, Y_N, Z_N};
+//
+// pub type Lab<T> = Lab<D65, T>;
+// pub type Laba<T> = Laba<D65, T>;
 
 ///CIE L*a*b* (CIELAB) with an alpha component. See the [`Laba` implementation in `Alpha`](struct.Alpha.html#Laba).
-pub type Laba<T = f32> = Alpha<Lab<T>, T>;
+pub type Laba<Wp, T = f32> = Alpha<Lab<Wp, T>, T>;
```
Ogeon: Here's the problem. `Lab` has no default white point, so that's why it has to be specified somewhere. I think we can safely set `D65` as the default everywhere, just to have something. It can be changed, after all.

Ogeon:
>Right now the Color conversions do not take chromatic adaptation into account. They only convert from and into the same white point.

I think that's good. Chromatic adaption can be more destructive than normal conversion, so I think it should be done explicitly.

>Can we use the existing FromColor and IntoColor to make this happen?
>Should I add this in this PR or in a new PR?

Those conversions should just work. The white point becomes a part of the type, so it should be restricted in the same way as the float type. I wrote them out in my example to make sure it was clear, but it should only be necessary if something other than `D65` is used, just as with the float type.

>Can we do anything to infer the default white points?

I don't know. It my depend on the context. Is it just the struct syntax? Is it only when the type isn't specified anywhere else?

sidred:
> Those conversions should just work. The white point becomes a part of the type, so it should be restricted in the same way as the float type

So ```FromColor<Wp, T>``` becomes ```FromColor<Swp, Dwp, T>``` and same with the IntoColor  trait. 

> I don't know. It my depend on the context. Is it just the struct syntax? Is it only when the type isn't specified anywhere else?

You can check the last commit for all the places I has to explicitly specify the type. It's pretty much everywhere.

Ogeon:
> So FromColor<Wp, T> becomes FromColor<Swp, Dwp, T> and same with the IntoColor trait.

That's not what I meant. I think it's good that chromatic adaption is explicit and not a part of `FromColor` or `IntoColor`. It fits better as its own trait, in my opinion.

> You can check the last commit for all the places I has to explicitly specify the type. It's pretty much everywhere.

Oh, no! It looks like we have stumbled upon an [unfinished feature](https://github.com/rust-lang/rust/issues/27336). See this playpen: http://is.gd/qFU2o5

I'm not sure what to do here... This problem will also be present in the planned `Rgb` reform.

sidred:
How is HashMap default hasher impelmentation working then?

Ogeon:
Looks like `new` is only implemented for `HashMap<K, V, RandomState>`: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#506

We could do something similar: http://is.gd/AIkZ09 But the situation is different, because the white point is never passed to a function.

sidred:
We can use something like this http://is.gd/do9sEJ

Ogeon:
Hmm... All of these are weird in different ways.

 * The alias makes `new` available without any fuss, but it's an alias, so we have to deal with naming in a good way. We don't want to "punish" anyone with too long names.
 * Only implementing `new` for `D65`, without passing the `Wp` to `with_wp`, is convenient for the default case, but may be surprising and inconvenient for the general case.
 * Only implementing `new` for `D65`, and passing the `Wp` to `with_wp`, has the same pros and cons, but with the need to pass an unused parameter instead of specifying a type parameter. Not sure if better or worse, to be honest.


It looks like these two also works:

```rust
let rgb = Rgb::<D65>::with_wp(1.0, 0.0, 0.0);
let rgb: Rgb<D65> = Rgb::with_wp(1.0, 0.0, 0.0);
```

That is, only one type parameter is actually necessary.

sidred:
Went with option 2, for now. Need to add chromatic adaptation

sidred:
Not sure why tests are failing. They work fine on both stable and nightly on my computer.

Ogeon:
Looks like `png` is broken. You probably haven't `cargo update`d.

Ogeon:
Looks like it's caused by the `bitflags` crate. I submitted an issue.

sidred:
This is mostly done from my end. Just need to make sure the tests pass once the png crate is fixed and any other documentation/ clean up required.

sidred:
Also let me know if any of the trait parameters in Chromatic Adaptation/ White Points need to be associated types. I am not 100% clear when to use associated types, so I just make everything trait parameters

Ogeon:
Great! I'll take a closer look in a moment. I haven't gotten any response from the png people, which is a bit surprising. It's widely used, so we can't be the only ones affected. I'm actually considering fixing it, myself, if I can.

> I am not 100% clear when to use associated types, so I just made everything trait parameters

My rule of thumb is to use associated types when the trait is only supposed to be implemented once per type. Like in `Mix`. A color should only be mixable using one type of mix factor, while `From` uses a type parameter because you may want to convert from more than one type.

src/blend/blend.rs
```diff
@@ -28,12 +28,13 @@ pub trait Blend: Sized {
...
     ///use palette::blend::PreAlpha;
+    ///use palette::white_point::D65;
     ///
-    ///type PreRgba = PreAlpha<Rgb<f32>, f32>;
+    ///type PreRgba = PreAlpha<Rgb<D65, f32>, f32>;
     ///
     ///fn blend_mode(a: PreRgba, b: PreRgba) -> PreRgba {
     ///    PreAlpha {
-    ///        color: Rgb::new(a.red * b.green, a.green * b.blue, a.blue * b.red),
+    ///        color: Rgb::with_wp(a.red * b.green, a.green * b.blue, a.blue * b.red),
```
Ogeon: Why was this changed? `Rgb<D65, f32>` implements `new`, right?

src/chromatic_adaptation.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+
+///Converts color from one reference white point to another (Swp -> Dwp)
+pub trait ChromaticAdaptation<T, Swp, Dwp, Color>
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>
+{
+    fn chromatic_adaptation(&self, method: TransformMatrix<T, Swp, Dwp>, Swp, Dwp) -> Color;
```
Ogeon: Is it necessary to pass `Swp` and `Dwp` as arguments here, or does this also break down otherwise?

src/convert.rs
```diff
@@ -10,49 +11,51 @@ use {Alpha, Rgb, Luma, Xyz, Yxy, Lab, Lch, Hsv, Hwb, Hsl, Color};
...
     ///Convert from XYZ color space
-    fn from_xyz(Xyz<T>) -> Self;
+    fn from_xyz(Xyz<Wp, T>) -> Self;
 
     ///Convert from Yxy color space
-    fn from_yxy(inp: Yxy<T>) -> Self {
-        Self::from_xyz(inp.into_xyz())
+    fn from_yxy(inp: Yxy<Wp, T>) -> Self {
+        let xyz = <Xyz<Wp, T> as FromColor<Wp,T>>::from_yxy(inp);
+        Self::from_xyz(xyz)
```
Ogeon: Shouldn't this be `Self::from_xyz(inp.into_xyz())`, like the rest?

src/lib.rs
```diff
@@ -434,61 +501,61 @@ make_color! {
     ///Linear luminance.
     Luma {
         ///Linear luminance.
-        y(luma: T)[alpha: T] => new;
+        y(luma: T)[alpha: T] => with_wp;
```
Ogeon: Maybe we should favor D65 in this case, or add `*_with_wp` variants. Not sure what's best, but `Color` is meant to be easy to use.

src/rgb.rs
```diff
@@ -127,21 +202,25 @@ impl<T: Float> Alpha<Rgb<T>, T> {
...
+impl<Wp, T> FromColor<Wp, T> for Rgb<Wp, T>
+    where T: Float,
+        Wp: WhitePoint<T>
+{
+    fn from_xyz(xyz: Xyz<Wp, T>) -> Self {
         Rgb {
             red: xyz.x * flt(3.2404542) + xyz.y * flt(-1.5371385) + xyz.z * flt(-0.4985314),
             green: xyz.x * flt(-0.9692660) + xyz.y * flt(1.8760108) + xyz.z * flt(0.0415560),
             blue: xyz.x * flt(0.0556434) + xyz.y * flt(-0.2040259) + xyz.z * flt(1.0572252),
+            white_point: PhantomData,
```
Ogeon: Is this conversion really correct for every white point?

src/white_point.rs
```diff
@@ -0,0 +1,143 @@
...
+macro_rules! generate_white_point {
+    ($x: ident => ($p1: expr, $p2:expr, $p3:expr)) => (
+        impl<T: Float> WhitePoint<T> for $x {
+
+            fn get_yxy() -> Yxy<Self, T> {
+                Yxy::with_wp(flt($p1), flt($p2), flt($p3))
+            }
+        }
+    );
+}
```
Ogeon: Looks like this macro can be removed now.

src/white_point.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+///CIE standard illuminant F7 represents a broadband fluorescent lamp for 2° Standard Observer.
+pub struct F7;
+impl<T: Float> WhitePoint<T> for F7 {
+    fn get_xyz() -> Xyz<Self, T> {
+        Xyz::with_wp(flt(0.95041), T::one(), flt(1.08747))
+    }
+}
+///CIE standard illuminant F7 represents a narrowband fluorescent lamp for 2° Standard Observer.
```
Ogeon: This looks like the wrong description.

src/yxy.rs
```diff
@@ -103,154 +163,202 @@ impl<T: Float> Limited for Yxy<T> {
...
+    where T: Float,
+        Wp: WhitePoint<T>
+{
+    fn default() -> Yxy<Wp, T> {
         // The default for x and y are the white point x and y ( from the default D65).
         // Since Y (luma) is 0.0, this makes the default color black just like for other colors.
         // The reason for not using 0 for x and y is that this outside the usual color gamut and might
         // cause scaling issues.
-        Yxy::new(flt(D65_X), flt(D65_Y), T::zero())
+        Yxy::with_wp(flt(D65_X), flt(D65_Y), T::zero())
```
Ogeon: This is incorrect. It should be based on `Wp`.

src/chromatic_adaptation.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl<T, Swp, Dwp, Color> ChromaticAdaptation<T, Swp, Dwp, Color> for IntoColor<Swp, T>
```
Ogeon: This implements it for the trait object `IntoColor<Swp, T>` and not `C: IntoColor<Swp, T>`, as far as I know. Maybe it works anyway, though...

src/chromatic_adaptation.rs
Ogeon: It would also be nice with a short form of this, that defaults to Bradford (I _think_ that's generally considered the standard).

Ogeon:
I have looked through it and commented on a few things, but it's otherwise very good. Let's hope the png issues are resolved soon.

src/rgb.rs
sidred: This is only for Srgb D65, but without adding Rgb variants not much can be done here. This can be taken care of while adding the Rgb variants which I will implement next.

src/chromatic_adaptation.rs
sidred: I think the chromatic adaption module needs more work. Still has some compilation issues.

src/convert.rs
sidred: This works for now and will require more changes for the Rgb Variants. Once we have that implemented,  we will have a better idea how complicated color enum gets and maybe think about it then?

src/blend/blend.rs
sidred: There was a reason I changed it. I can't figure out why with out the tests working :)

src/chromatic_adaptation.rs
Ogeon: Yeah, I saw a couple of small errors in the tests, as well, but didn't comment since we can't run them right now. Any way, maybe this trait can become a part of the `IntoColor` trait. It can be implemented by default.

src/chromatic_adaptation.rs
sidred: > Is it necessary to pass Swp and Dwp as arguments here, or does this also break down otherwise?

I was thinking that this will be easier that using traits
```
let lab: Lab<D50, T> = Lab::with_wp(0.5, 0.5 ,0.5)

let yxy = lab.chromatic_adaptation(D65, D50);
```
but can also be done as 
```
let yxy: Yxy<D50, _> = lab.chromatic_adaptation();
```
I'll change it.

> It would also be nice with a short form of this, that defaults to Bradford (I think that's generally considered the standard).

That's a good idea

src/lib.rs
sidred: This works for now and will require more changes for the Rgb Variants. Once we have that implemented, we will have a better idea how complicated color enum gets and maybe think about it then?

src/rgb.rs
Ogeon: Looks like it's not so simple: http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html

It will produce incorrect results, unless the white point is separated from the primaries.

src/convert.rs
Ogeon: Did you mean to comment on the `Color` enum comment? I think it's enough to change it back to use `new` for now. It won't stop anyone from using it with any other white point, by converting.

src/lib.rs
Ogeon: I think it's enough to change it back to use `new` for now. It won't stop anyone from using it with any other white point, by converting.

src/chromatic_adaptation.rs
Ogeon: It's hard to say which one is the easiest. `let yxy: Yxy<D50, _> = lab.chromatic_adaptation();` is at least consistent with how we treat white points in other situations. I guess we'll see when we can test.

sidred: Doing this ``` let yxy = lab.chromatic_adaptation(D65, D50);``` forces the user to repeat the source white point, when the information is already in the lab variable. So this definitely needs to change.

src/lib.rs
sidred: If I remember correctly this is a compilation issue with ```  y(luma: T)[alpha: T] => new;``` with the error ```trait not implemented for <D65, T>```. It is fixed by changing it to use the white point

Ogeon: That's odd. It shouldn't complain about traits. Must be something in the way the `impl` is done. It should say `impl<T: Float> Color<D65, T>` if `new` is used.

sidred: Yeah, will have to split the  make_color macro to support this.

Ogeon:
I was thinking a bit more about the chromatic adaption trait. I don't know how practical it is to merge it with the `IntoColor` trait, but I thought that there could possibly be a `From` and `Into` variant of it:

```rust
pub trait AdaptFrom<Swp, Dwp, T, C> where
    Swp: WhitePoint,
    Dwp: WhitePoint,
    T: Float,
{
    fn adapt_from(color: C) -> Self;
    fn adapt_from_using(color: C, matrix: ...) -> Self;
}

impl<S, D, Swp, Dwp, T> AdaptFrom<Swp, Dwp, T, S> for D where
    S: IntoColor<Swp, T>,
    D: FromColor<Dwp, T>,
    Swp: WhitePoint,
    Dwp: WhitePoint,
    T: Float,
{
    fn adapt_from(color: C) -> D {
        //...
    }

    fn adapt_from_using(color: C, matrix: ...) -> D {
        //...
    }
}

pub trait AdaptInto<Swp, Dwp, T, C> where
    Swp: WhitePoint,
    Dwp: WhitePoint,
    T: Float,
{
    fn adapt_into(self) -> C;
    fn adapt_into_using(self, matrix: ...) -> C;
}

impl<S, D, Swp, Dwp, T> AdaptInto<Swp, Dwp, T, D> for S where
    S: IntoColor<Swp, T>,
    D: AdaptFrom<Swp, Dwp, T, Src>,
    Swp: WhitePoint,
    Dwp: WhitePoint,
    T: Float,
{
    fn adapt_into(self) -> C {
        C::adapt_from(self)
    }

    fn adapt_into_using(self, matrix: ...) -> C {
        C::adapt_from_using(self, matrix)
    }
}
```

I'm obviously trying to mimic `From` and `Into` here, but I think they are quite good. You can use them as both `Rgb::<D50>::adapt_from(color)` and `color.adapt_into()`, and they are useful for different situations. `adapt_from` can, for example, be used like this: `maybe_color.map(Rgb::<D50>::adapt_from)` or even  `maybe_color.map(AdaptFrom::adapt_from)`, instead of (at best) `maybe_color.map(|c| c.adapt_into())`.

What do you think about that? The names are, of course, up for discussion.

src/lib.rs
Ogeon: I'm talking about ignoring `with_wp` for now and always assume D65 in the constructors. These things will change even more, as you said, so let's make it reasonably simple for ourselves.

sidred:
FromAdapt and IntoAdapt is good. I will add it in.

Can we merge adpatation and from and into color?

With impl specialization we could have merged Adaptation with from and into traits

Have the usual fast conversions when white points are the same
```
pub trait FromColor<Wp = D65, T = f32>: Sized
    where T: Float,
        Wp: WhitePoint<T>
{
   from_xyz()
  from_xyz_using() // should be same as from_xyz, no adaptation
}
```
Have the chromatic adaptation variant when different
```
pub trait FromColor<Swp, Dwp, T >: Sized
    where T: Float,
        Swp: WhitePoint<T>,
        Dwp: WhitePoint<T>
{
   from_xyz()
   from_xyz_using()
```

I wonder how much effort it would be using macro's?

One issue I can think is that now we can support chromatic adaptation for user defined white points (will be useful for icc profiles), macro's will only implement it for the whitepoints defined in the library.

Ogeon:
I feel like it's probably best to keep them apart at the moment. It will allow implementations that are independent from `From/IntoColor`, and I expect chromatic adaption to be significantly less common than conversion within the same white point, so it's good if the usual conversion has as little overhead as possible. I would also really like to keep the support for user defined white points.

src/yxy.rs
sidred: done

src/white_point.rs
sidred: done

src/white_point.rs
sidred: done

src/lib.rs
sidred: That won't work without making making color to only use D65. i.e. Color<T> with each variant as Luma(Luma<D65, T>), Xyz(Xyz<D65,T>) etc

src/convert.rs
sidred: done

src/lib.rs
Ogeon: The constructors are currently implemented as:

```rust
impl<Wp, T> Color<Wp, T>
    where T: Float,
        Wp: WhitePoint<T>
{
    $(
        $(
            #[$ctor_comment]
            pub fn $ctor_name$(<$($ty_params : $ty_param_traits$( <$( $ty_inner_traits ),*> )*),*>)*($($ctor_field: $ctor_ty),*) -> Color<Wp, T> {
                Color::$variant($variant::$ctor_original($($ctor_field),*))
            }
        )+
    )+
}

///<span id="Colora"></span>[`Colora`](type.Colora.html) implementations.
impl<Wp, T> Alpha<Color<Wp, T>, T>
    where T: Float,
        Wp: WhitePoint<T>
{
    $(
        $(
            #[$ctor_comment]
            pub fn $ctor_name$(<$($ty_params : $ty_param_traits$( <$( $ty_inner_traits ),*> )*),*>)*($($ctor_field: $ctor_ty,)* alpha: $alpha_ty) -> Colora<Wp, T> {
                Alpha::<$variant<Wp, T>, T>::$ctor_original($($ctor_field,)* alpha).into()
            }
        )+
    )+
}
````

I see no reason why it can't be changed to this:

```rust
impl<T: Float> Color<D65, T> {
    $(
        $(
            #[$ctor_comment]
            pub fn $ctor_name$(<$($ty_params : $ty_param_traits$( <$( $ty_inner_traits ),*> )*),*>)*($($ctor_field: $ctor_ty),*) -> Color<D65, T> {
                Color::$variant($variant::$ctor_original($($ctor_field),*))
            }
        )+
    )+
}

///<span id="Colora"></span>[`Colora`](type.Colora.html) implementations.
impl<T: Float> Alpha<Color<D65, T>, T> {
    $(
        $(
            #[$ctor_comment]
            pub fn $ctor_name$(<$($ty_params : $ty_param_traits$( <$( $ty_inner_traits ),*> )*),*>)*($($ctor_field: $ctor_ty,)* alpha: $alpha_ty) -> Colora<D65, T> {
                Alpha::<$variant<D65, T>, T>::$ctor_original($($ctor_field,)* alpha).into()
            }
        )+
    )+
}
````

Comment (sidred) [inline src/lib.rs]: I was thinking of something else entirely. Made the change.

File: src/rgb.rs (hunk 6 above)
Comment (sidred) [inline src/rgb.rs]: >  unless the white point is separated from the primaries.
As discussed in #31, I was planning to implement Rgb as RGB<Variant, WhitePoint, T> and see if it makes more sense than Rgb<Primaries, T>. 

I don't see what else can be done here for now.

File: src/lib.rs (hunk 5 above)
Comment (Ogeon) [inline src/lib.rs]: Ah :) Happens to everyone.

**Comment by Ogeon** [comment]
Everything should build now if you update `image` to `0.7`. Turns out they were using an old version of `png`.

**Comment by sidred** [comment]
Adapt from is currently defined as 

```
///Converts color from one reference white point to another via chromatic adaptation (Swp -> Dwp)
pub trait AdaptFrom<S, Mat, Swp, Dwp, T>: Sized
    where T: Float,
          Swp: WhitePoint<T>,
          Dwp: WhitePoint<T>,
          Mat: TransformMatrix<Swp, Dwp, T>,
          S: IntoColor<Swp, T>,
{
    ///Convert the source color to the destination color using the bradford method
    fn adapt_from(color: S) -> Self;
    ///Convert the source color to the destination color using the specified method
    fn adapt_from_using(color: S, method: Mat) -> Self;
}

```


This works 

```
        let input_a = Xyz::<A, f32>::with_wp(0.315756, 0.162732, 0.015905);
        let computed_vonkries: Xyz<C, f32> = Xyz::adapt_from_using(input_a, Method::VonKries);
```

But this does not
```
        let computed_bradford: Xyz<C,f32> = Xyz::adapt_from(input_a);
       // errors with unable to infer enough type information about `_`; type annotations or generic parameter binding required
```
It is failing to infer information for TransformationMatrix

```
        let computed_bradford: Xyz<C, f32> = <Xyz<_,_> as AdaptFrom<_,_,_,_,_>>::adapt_from(input_a);


// Erros with

src/chromatic_adaptation.rs:264:71: 264:72 error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [E0282]
src/chromatic_adaptation.rs:264         let computed_bradford: Xyz<C, f32> = <Xyz<_,_> as AdaptFrom<_,_,_,_,_>>::adapt_from(input_a);
                                                                                                     
```

Any way to make ```  let computed_bradford: Xyz<C,f32> = Xyz::adapt_from(input_a);  ``` work directly?

**Comment by sidred** [comment]
Changed to adapt to

```
///Converts color from one reference white point to another via chromatic adaptation (Swp -> Dwp)
pub trait AdaptFrom<S, Swp, Dwp, T>: Sized
    where T: Float,
          Swp: WhitePoint<T>,
          Dwp: WhitePoint<T>,
          S: IntoColor<Swp, T>,
{
    ///Convert the source color to the destination color using the bradford method
    fn adapt_from(color: S) -> Self;
    ///Convert the source color to the destination color using the specified method
    fn adapt_from_using<Mat: TransformMatrix<Swp, Dwp, T>>(color: S, method: Mat) -> Self;
}
```

Now both these work
```
        let computed_bradford: Xyz<C,f32> = Xyz::adapt_from(input_a);
        let computed_vonkries: Xyz<C, f32> = Xyz::adapt_from_using(input_a, Method::VonKries);


```

**Comment by Ogeon** [comment]
I think that's the way to go.

File: src/chromatic_adaptation.rs [hunk 11]
```diff
@@ -0,0 +1,285 @@
...
+    }
+}
+
+
+///Converts color from one reference white point to another via chromatic adaptation (Swp -> Dwp)
+pub trait AdaptFrom<S, Swp, Dwp, T>: Sized
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+          S: IntoColor<Swp, T>,
```
Ogeon: I think we can drop this requirement in the definition.

src/chromatic_adaptation.rs
```diff
@@ -0,0 +1,285 @@
...
+    ///Convert the source color to the destination color using the specified method
+    fn adapt_into_using<M: TransformMatrix<Swp, Dwp, T>>(self, method: M) -> D;
+}
+
+impl<S, D, Swp, Dwp, T> AdaptInto<D, Swp, Dwp, T> for S
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+          S: IntoColor<Swp, T>,
+          D: AdaptFrom<S, Swp, Dwp, T> + IntoColor<Dwp,T>
```
Ogeon: Thinking about it, the only necessary requirement for `S` and `D` is that `D: AdaptFrom<S, Swp, Dwp, T>`, once the `S: IntoColor<Swp, T>` requirement has been removed from `AdaptFrom`.

src/blend/blend.rs
sidred: Should be Rgb::new. Fixed

src/chromatic_adaptation.rs
sidred: I've had to use
```
pub trait AdaptFrom<S, Swp, Dwp, T>: Sized
    where T: Float,
          Swp: WhitePoint<T>,
          Dwp: WhitePoint<T>,
{
```

because adapt from needs it
```
fn adapt_from_using<M: TransformMatrix<Swp, Dwp, T>>(color: S, method: M) -> Self;
```

src/chromatic_adaptation.rs
sidred: Fixed

sidred:
Are you ok with leaving the RGB conversion as is for now? That's the only thing remaining for now.

Ogeon:
That will leave it in an incorrect state, which I'm not super comfortable with... [That link](http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html) I posted before has an algorithm/formula for converting from RGB to XYZ, which looks like it's more or less like this:

 1. convert RGB chromaticity coordinates from `Yxy::new(1.0, x, y)` to XYZ,
 2. put the calculated values in a matrix,
 3. multiply the matrix and the white point XYZ to get some constants,
 4. put the constants and the previously calculated XYZ coordinates values in a new matrix,
 5. multiply the matrix and the RGB values.

We have all the facilities for that. The tricky part is, however, to invert it in an efficient way and I'm not sure how hard that is. Maybe there are some shortcuts. Anyway, do you think you could get _something_, with hard coded sRGB primaries, that gives the correct results? I would rather have that, and then optimize, than something that doesn't work as expected. I mean, it has to be done eventually, and I think it will simplify the RGB reform testing if it's already correct. What do you say?

Ogeon:
I will, of course, help as much as I can, and there's no hurry. I'm currently looking for a good matrix inversion algorithm.

sidred:
Can we merge this into a temporary branch for now, add in the rgb changes and the then merge it with the master then? I guess in my head the chromatic adaptation and rgb stuff are separate :)

I just want to make sure the changes so far are ok, before working on the rgb stuff. The formula's needs to invert matrices , so might need a external library for that and also I am currently not sure how the rgb spaces interact with the conversion stuff.

Let me know if this works or else I will add in the conversion for Srgb now.

Ogeon:
I would say that it's most related to the abstraction of the white point, since it's one of the components in the conversion formula. I think we can pull this off by hard coding the sRGB primaries, since that's almost what we have done already.

I think that it sounds like a good idea, in any case, to pull in a linear algebra library. These things are tedious enough as it is, so why make it worse? I don't know which library is the best, but I have used [cgmath](https://crates.io/crates/cgmath) before, and it has served me well. I have checked and it can invert 3x3 matrices without problem. It's also based on the num crate (as far as I can tell), so we don't have to change our reliance on `Float`. [nalgebra](https://crates.io/crates/nalgebra) is also popular, but it looks like it's a bit spotty when it comes to the matrix inversion. There should otherwise not be any significant differences, as far as I can tell.

I understand that this PR is becoming quite big now, so let's take it one step at the time. Let's begin by checking and polishing what's already done, to make sure it's all good, and integrate the linear algebra library. And _then_ we can decide what to do with the XYZ <-> RGB and RGB <-> Luma (yes, that's also affected) conversion.

So, let's put that problem aside, for now, and focus on what's already here. I'll take another look at the new stuff, but there shouldn't be an awful lot to fix now.

src/chromatic_adaptation.rs
```diff
@@ -0,0 +1,283 @@
...
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+          S: IntoColor<Swp, T>,
+          D: FromColor<Dwp, T>,
+
+{
+    fn adapt_from(color: S) -> D {
+        Self::adapt_from_using(color, Method::Bradford)
+    }
```
Ogeon: This can be a default implementation on the definition. I expect most of the implementations of this to be the exact same.

src/chromatic_adaptation.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+
+///Converts color from one reference white point to another via chromatic adaptation (Swp -> Dwp)
+pub trait AdaptFrom<S, Swp, Dwp, T>: Sized
+    where T: Float,
+          Swp: WhitePoint<T>,
+          Dwp: WhitePoint<T>,
+{
+    ///Convert the source color to the destination color using the bradford method
```
Ogeon: This is nit picking, but we can't guarantee that someone will implement this with an other method. Adding "by default" may be a good idea :)

src/chromatic_adaptation.rs
```diff
@@ -0,0 +1,283 @@
...
+        let computed = <TransformMatrix<D65, D50, _>>::generate_transform_matrix(&bradford);
+        for (e, c) in expected.iter().zip(computed.iter()) {
+            assert_relative_eq!(e, c, epsilon = 0.0001)
+        }
+    }
+
+    #[test]
+    fn chromatic_adaptation_from_a_to_c() {
+        let input_a = Xyz::<A, f32>::with_wp(0.315756, 0.162732, 0.015905);
+        // let original_c = Xyz::<C, f32>::with_wp(0.203465, 0.116458, 0.053125);
```
Ogeon: Just a reminder to remove these commented out colors :)

src/lch.rs
```diff
@@ -28,23 +37,59 @@ pub struct Lch<T: Float = f32> {
...
+    }
+}
+
+impl<Wp, T> Lch<Wp,T>
+    where T: Float,
+        Wp: WhitePoint<T>
+{
     ///CIE L*C*h°.
-    pub fn new(l: T, chroma: T, hue: LabHue<T>) -> Lch<T> {
+        pub fn with_wp(l: T, chroma: T, hue: LabHue<T>) -> Lch<Wp, T> {
```
Ogeon: Looks like an accidental extra indentation.

src/lib.rs
```diff
@@ -241,21 +248,25 @@ mod hwb;
 
 mod hues;
 
-mod tristimulus;
 mod convert;
 mod equality;
+///Provides the Chromatic Adpatation to convert color from one illuminant to another
+pub mod chromatic_adaptation;
+///Defines the tristimulus values of the CIE Illuminants..
```
Ogeon: I would appreciate it if these comments can be moved into the modules. That keeps everything in the same place.

src/white_point.rs
```diff
@@ -0,0 +1,147 @@
...
+
+use {Xyz, flt};
+
+///WhitePoint defines the Yxy color co-ordinates for a given white point.
+///
+///A white point (often referred to as reference white or target white in technical documents)
+///is a set of tristimulus values or chromaticity coordinates that serve to define the color
+///"white" in image capture, encoding, or reproduction.
+pub trait WhitePoint<T: Float>: Sized {
+    ///Get the Yxy chromacity co-ordinates for the white point.
```
Ogeon: A typo. `Yxy` instead of `Xyz`.

src/white_point.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn get_xyz() -> Xyz<Self, T>;
+}
+
+///CIE standard illuminant A is intended to represent typical, domestic, tungsten-filament lighting.
```
Ogeon: Would you mind adding a short "header" or "summary" paragraph to these? Something like this should do:

```rust
///CIE standard illuminant A.
///
///...
```

It will look much nicer in the docs.

src/white_point.rs
```diff
@@ -0,0 +1,147 @@
...
+///Its relative spectral power distribution is that of a Planckian radiator at a temperature of approximately 2856 K.
+///Uses the CIE 1932 2° Standard Observer
+#[derive(Copy, Clone, Debug, PartialEq)]
```
Ogeon: Might as well derive `Eq`, just because we can.

src/white_point.rs
```diff
@@ -0,0 +1,147 @@
...
+        Xyz::with_wp(flt(0.94972), T::one(), flt(1.22638))
+    }
+}
+///CIE standard illuminant E represents the equal energy radiator
+///Uses the CIE 1932 2° Standard Observer
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct E;
+impl<T: Float> WhitePoint<T> for E {
+    fn get_xyz() -> Xyz<Self, T> {
+        Xyz::with_wp(flt(1.00000), T::one(), flt(1.00000))
```
Ogeon: I think it's better to use `T::one()` for all of them.

src/white_point.rs
```diff
@@ -0,0 +1,147 @@
...
+///for 10° Standard Observer.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct D65Degree10;
+impl<T: Float> WhitePoint<T> for D65Degree10 {
+    fn get_xyz() -> Xyz<Self, T> {
+        Xyz::with_wp(flt(0.9481), T::one(), flt(1.073))
+    }
+}
+///D75 White Point is the natural daylight with a color temperature of around 7500K
+///for 2° Standard Observer.
```
Ogeon: Should be "10°".

tests/convert/lab_lch.rs
```diff
@@ -1,10 +1,10 @@
 use palette::{Lab, Lch, IntoColor};
+use palette::white_point::D65;
 
 #[test]
 fn lab_lch_green() {
-    let lab = Lab::new(0.4623,-0.517,0.499);
-    let lch = Lch::new(0.4623,	0.7185,	136.02.into());
-
+    let lab = Lab::<D65, f64>::new(0.4623,-0.517,0.499);
+    let lch = Lch::<D65, f64>::new(0.4623,	0.7185,	136.02.into());
```
Ogeon: Can these be changed back to just `*::new`?

Ogeon:
I think that was all. Nothing really big to change, and mostly typos and documentation.

sidred:
All changes done

sidred:
Here is the inverse for 3x3 matrix

```
fn invert_array(a: [f64; 9]) -> [f64; 9] {
    let d0 = a[4] * a[8] - a[5] * a[7];
    let d1 = a[3] * a[8] - a[5] * a[6];
    let d2 = a[3] * a[7] - a[4] * a[6];
    let det =  a[0] * d0 - a[1] * d1 + a[2] * d2;
    if !det.is_normal() {
        return [0.0 ; 9]
    }
    let d3 = a[1] * a[8] - a[2] * a[7];
    let d4 = a[0] * a[8] - a[2] * a[6];
    let d5 = a[0] * a[7] - a[1] * a[6];
    let d6 = a[1] * a[5] - a[2] * a[4];
    let d7 = a[0] * a[5] - a[2] * a[3];
    let d8 = a[0] * a[4] - a[1] * a[3];

    [d0/det, -d3/det, d6/det, -d1/det, d4/det, -d7/det, d2/det, -d5/det, d8/det]
}
```

Is ii better to use an external library?

sidred:
What happens when the inverse does not exist? do we return black color, original color?

Ogeon:
> I can't figure out if this is simpler or if it's better to add an external library.

We don't have to worry about correctness if we add an external library.

> What happens when the inverse does not exist? do we return black color, original color?

Good question. What does it even mean for it to not have an inverse? Can it happen with sane coordinates? Is it even ok to panic or should it return an `Option` instead of assuming?

sidred:
Option or result does not get us anything ,we will have to change the entire from and into color to use the option types (it will be odd to only have it for the rgb colors). 

Most rgb color spaces should not have an issues. Inverse should be present. This will only be an issue with user defined rgb varaints / custom icc profiles etc.
 
So we either panic or return black color. 
Trying to think of ways to make the black color work. May add an error log message. Maybe we can make is_valid for custom color spaces and return a Option/Result in the constructor of the color spaces. 

Panic will be simpler though and even the std library does it for things like divide by 0.

sidred:
I've looked at both cgmath and nalgrbra. Both perform similary to the [T; 9]. cgmath uses a struct of vec's and nalgebra uses a struct with 9 elements, but both perform roughly the same.

cgamth uses S of type BaseFloat with additional trait bounds apart from float, and nalgrbra has not trait bounds, so nalgebra is a better option.

Ogeon:
> Option or result does not get us anything ,we will have to change the entire from and into color to use the option types (it will be odd to only have it for the rgb colors).

Oh, oops, my mind was stuck somewhere in chromatic adaption for some reason... Ignore that idea :P

>Panic will be simpler though and even the std library does it for things like divide by 0.

I'm starting to think that too. I can't think of a practical reason to have such a set of primaries and expect conversion to work, so we might as well.

Ogeon:
> cgamth uses S of type BaseFloat with additional trait bounds apart from float, and nalgrbra has not trait bounds, so nalgebra is a better option.

I just realized that they don't blanket implement the traits, so it will interfere with how we do it. The only thing I'm worried about regarding nalgebra is that `Mat3` doesn't implement `Inv`. It's implemented for rotation matrices, but I'm not sure if that's enough. They may ignore scale and translation.

sidred:
>  he only thing I'm worried about regarding nalgebra is that Mat3 doesn't implement Inv. It's implemented for rotation matrices, but I'm not sure if that's enough. They may ignore scale and translation.

Not sure I understand. There is an inv() for Mat3

sidred:
Inv is implemented is here https://github.com/sebcrozet/nalgebra/blob/6fee70bd19a42454ce896f48f1b4b034565262f3/src/structs/spec/mat.rs#L64

Ogeon:
Oh, it its! The documentation must be outdated or something. Now there's just that `BaseNum` requirement...

Ogeon:
...looks like it's the same situation as with cgmath. It's custom and not blanket implemented.

sidred:
This seems to work
```
use num::Float;
use nalgebra;
use nalgebra::Inv;


pub fn test<T: Float>(a1: T, b1:T, c1:T, d1: T) -> nalgebra::Mat2<T> {
    nalgebra::Mat2::new(a1,b1,c1,d1)
}

pub fn run() {
    let out = test(2.0_f32, 3.0, 4.0, 5.0);
    let inv = out.inv();
    println!("{:?}",inv );
}


```

sidred:
cgmath errors with BaseFloat not implemented
```
pub fn test_cgamth<T: Float>(a1: T, b1:T, c1:T, d1: T) -> cgmath::Matrix2<T> {
    cgmath::Matrix2::new(a1,b1,c1,d1)
}

pub fn run() {
    let out = test(2.0_f32, 3.0, 4.0, 5.0);
    let inv = out.inv();
    println!("{:?}",inv );
}
```

So i guess we are going with nalgebra

Ogeon:
But that doesn't restrict where it's important. This is closer to our situation:

```rust
use num::Float;
use nalgebra;
use nalgebra::Inv;


pub fn test<T: Float>(a1: T, b1:T, c1:T, d1: T) -> Option<nalgebra::Mat2<T>> {
    nalgebra::Mat2::new(a1,b1,c1,d1).inv()
}

pub fn run() {
    let out = test(2.0_f32, 3.0, 4.0, 5.0);
    println!("{:?}",out );
}
```

sidred:
yeah, checked at the wrong place. blas has too much overhead I think and I don't see any other alternative. I'll add the custom function above with a few tests.

Ogeon:
I found vecmath, which [looked promising](https://github.com/PistonDevelopers/vecmath/blob/master/src/lib.rs#L1628) until I saw that they have their own special float crate. I can't find anything else that's not dynamic.

sidred:
Do you see any issues with aliasing [T; 9] as Mat9?
```
///A 9 element array representing a 3x3 matrix
pub type Mat9<T> = [T;9];

pub struct ConeResponseMatrices<T: Float> {
    ///3x3 matrix for the cone response domains
    pub ma: Mat9<T>,
    ///3x3 matrix for the inverse of the cone response domains
    pub inv_ma: Mat9<T>,
}
```

Ogeon:
I would rather call it `Mat3` or `Mat3x3`, but otherwise, no. Go for it. It communicates intention better than plain `[T; 9]`.

sidred:
I have implemented Rgb <-> Xyz conversion
- No test for Srgb D65 to D50 conversion. I could not find the transformation matrix for this.
- Is there a way to compare WP == D65. I have used a work around to compare the xyz components
- What needs to be done with the luma?

Ogeon:
The XYZ of the primaries should be the same, no matter the white point, so that implementation of `RgbPrimaries` should be generic over `Wp`. There shouldn't be a need to check what the white point is.

Luma is basically XYZ without XZ, so it's affected in the same way. Maybe it would be easiest to go RGB -> XYZ -> Luma. We can still go Luma -> RGB, though.

Oh, and there's no need to prefix paths in `use` with `::`. They are global by default.

Ogeon:
It would be good with some tests, as well. Something like this should do:

```rust
let a: Rgb<D65> = Rgb::new(1.0, 1.0, 1.0);
let b: Rgb<D50> = a.adapt_into();
assert_relative_eq!((a.red, a.green, a.blue), (b.red, b.green, b.blue));
```

and maybe a couple more for other colors than white.

sidred:
I am surprised that test actually passes. I was expecting them to be different.

Here is a check on http://www.brucelindbloom.com/index.html?Eqn_DIlluminant.html

use D65 white point and srggb space convert rgb(1,1,1)  to xyz (0.95, 1, 1.08). Now change white point to D50 with same srgb space and convert xyz to rgb which gives (0.92, 1.0, 1.14), which was what I was expecting.

Ogeon:
I don't have Java in my browser :P Sounds like strange results, anyway. It just hit me that RGB can jump between white points, since it describes the transition from primaries to white, but I'm not sure we can make such an optimization and still keep the blanket implementation of `AdaptFrom`.

src/lib.rs
```diff
@@ -241,21 +248,26 @@ mod hwb;
 
 mod hues;
 
-mod tristimulus;
 mod convert;
 mod equality;
+pub mod chromatic_adaptation;
+pub mod white_point;
+pub mod matrix;
+pub mod rgb_primaries;
```
Ogeon: I'm not sure if we should make this module public right now. It's basically an implementation detail at this stage.

sidred:
I meant this link http://www.brucelindbloom.com/ColorCalculator.html. You'll see the same on easyrgb.com

src/lib.rs
sidred: Mat9  for chromatic adaptation needs to be public for implementing custom TransfromMatrix.

Ogeon:
I'm not sure where the chromatic adaption happens in there... It affects the values, but I can't determine how. I'll dig in the source and see if it's in the JavaScript.

Ogeon:
Oh, it looks like it's adapting between reference white and the white point of the RGB model. That's good to know...

src/lib.rs
Ogeon: I guess `matrix` can be public, but the state of `rgb_primaries` is a bit too fluid.

src/gradient.rs
```diff
@@ -375,7 +375,8 @@ mod test {
 
         let v1: Vec<_> = g1.take(10).take(5).collect();
         let v2: Vec<_> = g2.take(5).collect();
-
-        assert_eq!(v1, v2);
+        for (t1, t2) in v1.iter().zip(v2.iter()){
+            assert_relative_eq!(t1, t2);
+        }
```
Ogeon: I just realized, doesn't `Vec<T: ApproxEq>` implement `ApproxEq`?

src/luma.rs
```diff
@@ -58,37 +108,68 @@ impl<T: Float> Alpha<Luma<T>, T> {
...
             luma: yxy.luma,
+            white_point: PhantomData,
         }
     }
 
-    fn from_rgb(rgb: Rgb<T>) -> Self {
+    fn from_rgb(rgb: Rgb<Wp, T>) -> Self {
         Luma {
             luma: rgb.red * flt(0.2126) + rgb.green * flt(0.7152) + rgb.blue * flt(0.0722),
+            white_point: PhantomData,
```
Ogeon: I think we can just remove the `from_rgb` implementation.

src/rgb_primaries.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn green() -> Yxy<Wp, T>;
+    ///Tristimulus values for blue
+    fn blue() -> Yxy<Wp, T>;
+}
+
+///Srgb color space with default D65 white point
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct SrgbSpace;
+
+impl<T: Float> RgbPrimaries<D65, T> for SrgbSpace {
```
Ogeon: This is where `D65` can be replaced by a generic `Wp`. The way I understand it is that the primaries are the same, no matter what the white point is.

src/rgb_primaries.rs
```diff
@@ -0,0 +1,101 @@
...
+    let b_d65: Xyz<D65, T> = SrgbSpace::blue().into_xyz();
+
+    let wp = Wp::get_xyz();
+    let d65 = D65::get_xyz();
+    let mut transform_matrix = if wp.x == d65.x && wp.y == d65.y && wp.z == d65.z {
+        mat3_from_primaries(r_d65, g_d65, b_d65)
+    } else {
+        let r_wp: Xyz<Wp, T> = r_d65.adapt_into();
+        let g_wp: Xyz<Wp, T> = g_d65.adapt_into();
+        let b_wp: Xyz<Wp, T> = b_d65.adapt_into();
```
Ogeon: These adaptions, and this whole `if` should be unnecessary one the previosly mentioned `D65` -> `Wp` change is done.

src/lib.rs
Ogeon: ...or maybe we should just reexport `Mat3` and keep `matrix` private. It doesn't really give anything more useful than that. This isn't a linear algebra library, after all.

sidred: I was thinking the same too. Done.

src/gradient.rs
sidred: It errors with

```
error: the trait `approx::ApproxEq` is not implemented for the type `collections::vec::Vec<rgb::Rgb<white_point::D65, _>>` 
```

sidred:
So there is still an issue with the chromatic adaptation test for Rgb\<D65\> to Rgb\<D50\> . They still show up a 1 for all the values

sidred:
So here is the test 

```
    #[test]
    fn d65_to_d50() {
        let a: Rgb<D65> = Rgb::new(1.0, 1.0, 1.0);
        let x: Xyz<D65> = a.into_xyz();
        let x1: Xyz<D50> = x.adapt_into();
        let y: Rgb<D50> = x1.into_rgb();
        let b: Rgb<D50> = a.adapt_into();
        println!("");
        println!("RGB D65 {:?}", a);
        println!("XYZ D65 {:?}", x);
        println!("XYZ D50 {:?}", x1);
        println!("RGB D50 {:?}", y);
        println!("RGB D50 {:?}", b);
        assert_relative_eq!(a.red, b.red);
        assert_relative_eq!(a.green, b.green);
        assert_relative_eq!(a.blue, b.blue);
    }
```
Here is the output
```
RGB D65 Rgb { red: 1, green: 1, blue: 1, white_point: PhantomData }
XYZ D65 Xyz { x: 0.95047003, y: 1, z: 1.08883, white_point: PhantomData }
XYZ D50 Xyz { x: 0.96422017, y: 1.0000001, z: 0.82521, white_point: PhantomData }
RGB D50 Rgb { red: 1.0000005, green: 1.0000001, blue: 0.9999999, white_point: PhantomData }
RGB D50 Rgb { red: 1.0000005, green: 1.0000001, blue: 0.9999999, white_point: PhantomData }

```
I think we are missing something in the Rgb-Xyz converison

src/gradient.rs
Ogeon: Alright, then :)

Ogeon:
Thinking about all of this is a bit mind bending, but the way I understand it is that XYZ and xyY are the only spaces where the values are actually different for different white points. The other spaces are based on perception, meaning that RGB(1.0, 1.0, 1.0) is white, no matter what energy levels or optical nerve signals that corresponds to. It still looks white. Adapting from `Rgb<D65>` to `Rgb<D50>` is therefore practically pointless, because white is white, and white is always RGB(1.0, 1.0, 1.0).

Ogeon:
Lindbloom's color calculator gives the same XYZ values (using sRGB, which is D65, and adapting to D50). I tried to use the easyrgb calculator as well, but it gave me an error for D50 RGB -> XYZ (probably due to too precise validation), but going XYZ -> D50 RGB gives roughly (1, 1, 1). I think all is well.

sidred:
It should be (1,1,1) in both cases. http://www.brucelindbloom.com/ColorCalculator.html shows the gamma corrected values and we are looking at the linear values. So that expalins the difference in the values seen.

Also from the link below, there are 2 types of white points ICC D65 and ASTM D65 (same for all other D series). www.brucelindbloom.com uses ASTM, ICC profiles use the ICC values. There are slight differences between them. I've used the ASTM values.

More information on this http://ninedegreesbelow.com/photography/srgb-color-space-to-profile.html

Ogeon:
> Also from the link below, there are 2 types of white points ICC D65 and ASTM D65 (same for all other D series). www.brucelindbloom.com uses ASTM, ICC profiles use the ICC values. There are slight differences between them. I've used the ASTM values.

It never ends, does it? :confounded: We should make sure we use the correct one for sRGB, and maybe provide the others as well. Maybe group them in modules. I don't know. At least make sure sRGB conforms to its standard for this PR, and maybe improve the rest later. We must draw the line somewhere.

sidred:
Does everything look good. As as far as I can see I have made all the recommended changes.

sidred:
>  I don't know. At least make sure sRGB conforms to its standard for this PR, and maybe improve the rest later. We must draw the line somewhere.

There are very small differences between, like in the 4th decimal place 0..3271 and 0.3272. We should be ok.

src/white_point.rs
Ogeon: This is still here.

Ogeon:
I suppose it's fine, then. I hope nobody thinks this is suitable for super high precision life-on-the-stake applications :wink:

I'm still checking some of the new stuff, but I'm quite sure there's nothing more. Just that typo.

Ogeon:
I can't find the `d65_to_d50` test. Did you skip it?

sidred:
Yup, not sure if it makes sense

Ogeon:
It does, as a regression test, to check that nothing odd is going on. That, and the typo fix, should be all.

sidred:
Let me merge all commit and take another look

Ogeon:
Ok, just tell me when you feel ready. I'll be busy with all sorts of other things, so no hurry.

sidred:
Fixed up a few typos and changed the docs for chromatic adaptation, Xyz( referenced only D65 ) and Yxy.

Should be good to go!

Ogeon:
Nice! I'll, however, have to postpone my final review until tomorrow.

src/xyz.rs
```diff
@@ -18,59 +20,111 @@ pub type Xyza<T = f32> = Alpha<Xyz<T>, T>;
...
-pub struct Xyz<T: Float = f32> {
+///Conversions and operations on this color space depend on the defined white point
+#[derive(Debug, PartialEq)]
+pub struct Xyz<Wp = D65, T = f32>
+    where T: Float,
+        Wp: WhitePoint<T>
+{
     ///X is the scale of what can be seen as a response curve for the cone
-    ///cells in the human eye. It goes from 0.0 to 0.95047.
+    ///cells in the human eye. It's range depends
```
Ogeon: "It's" should be "Its". Same for `z`.

src/pixel/gamma_rgb.rs
```diff
@@ -44,45 +50,104 @@ pub struct GammaRgb<T: Float = f32> {
 
     ///The decoding gamma value. Commonly 2.2.
     pub gamma: T,
+
+    ///The white point associated with the color's illuminant and observer.
+///D65 for 2 degree observer is used by default.
+pub white_point: PhantomData<Wp>,
```
Ogeon: Missing indentation.

src/lch.rs
```diff
@@ -28,23 +37,59 @@ pub struct Lch<T: Float = f32> {
...
+{
+    fn clone(&self) -> Lch<Wp, T> { *self }
 }
 
-impl<T: Float> Lch<T> {
+impl<T> Lch<D65, T>
+    where T: Float,
+{
+    ///CIE L*C*h° with white point D65.
+        pub fn new(l: T, chroma: T, hue: LabHue<T>) -> Lch<D65, T> {
```
Ogeon: Excessive indentation.

Ogeon:
I found a couple of typos that I couldn't ignore, so I took the liberty to point out a couple of strange indentations as well. It's otherwise great and I would have merged it straight away.

sidred:
Done!

Ogeon:
Great! Thank you! It's exciting how this library has grown from the original idea of just being a simple color calculations library. It's still a color calculations library, and it's still about as simple to use, but it has so much more power. And more is coming.

@homu r+

PR #57, opened 2016-02-25
[WIP] Restore the proper scale of Lab and Lch

### LINKED ISSUE #49: Lab and Lch values
Continuing discussion from #48.

Lab values are currently scaled, L by 100 and a and b by 128. Even though a and b are unbound, typical ranges are -128 to 128. 

By my limited research so far, I have not seen scaled lab values used ( unlike rgb which is usually assumed in the range 0-1 ). For example Conversion formula's, Color Difference formula's and most research papers assume unscaled lab values (I am happy to be corrected on this). 

Right now using the lab spaces implies unnecessary scaling when used practically. So I am proposing that the Lab values should be changed to be unscaled. Same for Lch values.

If the decision is to still scale the value, we should provide a new_from_unscaled and into_unscaled function, to get values in and out.

Ogeon:
This restores the scale of `Lab` and `Lch` to be the same as in all of the specifications within reach. That is:

 * `Lab::l`  and `Lch::l` goes from 0 to 100,
 * `Lab::a` and `Lab::b` goes from -128 to 127,
 * `Lch::chroma` goes from 0 to about 128-181.

This is a breaking change, since everything that uses these values directly will be wrong. Closes #49.


Ogeon:
The tests are still wrong and needs to be tweaked. Some of the inaccuracies are huge, like

```
    left = Lab { l: 91.11652, a: -51.481186, b: -14.387321 }
    right = Lab { l: 91.12, a: -48.08, b: -14.14 }
```

sidred:
I've used a tolerance of 0.5 which translates to 5 when scaled. 

I wonder if this is due to the differences in the default white point being used

```
pub const X_N: f64 = 0.95047;
pub const Y_N: f64 = 1.0;
pub const Z_N: f64 = 1.08883;
```

sidred:
Nope. Color mine uses the same values:
```
		final double REF_X = 95.047; // Observer= 2°, Illuminant= D65
		final double REF_Y = 100.000;
		final double REF_Z = 108.883;
```
But the formula in color mine is incorrect
```
		final double REF_X = 95.047; // Observer= 2°, Illuminant= D65
		final double REF_Y = 100.000;
		final double REF_Z = 108.883;

		double x = pivotXyz(xyz.X / REF_X);
		double y = pivotXyz(xyz.Y / REF_Y);
		double z = pivotXyz(xyz.Z / REF_Z);

		return new Lab(116 * y - 16, 500 * (x - y), 200 * (y - z));
```
Its not using the transformation function
```
        fn convert<T: Float>(c: T) -> T {
            let epsilon: T = (flt::<T,_>(6.0 / 29.0)).powi(3);
            let kappa: T = flt(841.0 / 108.0);
            let delta: T = flt(4.0 / 29.0);
            if c > epsilon {
                c.powf(T::one() / flt(3.0))
            } else {
                (kappa * c ) + delta
            }
        }

```

I guess we need a better set of tests

sidred:
We have a test for Xyz <-> Yxy, already from the CIE tables.

https://www.rit.edu/cos/colorscience/rc_useful_data.php has a few calculated values for Xyz, Lab and Lch  ( CIEDE200 Color Difference formula, Pointer data). But it uses a 10 degree observer.

This should cover the Cie spaces. 


We need to figure out the Rgb derived colors and their conversions.

Ogeon:
Thanks, I'll take a look at them later and see if it's possible to piece things together.

Ogeon:
Turns out the accuracy of the Yxy values were too low, so I'm currently working my way through them, adding more decimals.

sidred:
So the actual scaling part is done? Is it just the tests remaining?

Ogeon:
Yes, unless I forgot something. I'm pretty much just copy-pasting XYZ numbers to the easyrgb calculator, dividing them by 100, converting, and copy-pasting the xyY numbers back.

sidred:
Closed by #61

PR #59, opened 2016-03-06
[WIP] Add rgb variants like Srgb etc

sidred:


sidred:
To-Do
- [ ] Add RgbVariants/ Primaries /Gamma Trait
- [ ] Add Linear Rgb type
- [ ] Implement conversion and chromatic adpatation for the linear and gamma rgb types
- [ ] Add Linear HSL/HWB/HSV types and converisons

sidred:
I think you have a much better understanding of how to proceed next. So I am  closing this now.

Ogeon:
Are you sure? I mean, I may have ideas and I'm trying my best to communicate them and find a compromise with your ideas. I feel like they are very close, except for some details. Also, I'm sorry if I'm clumsy when expressing myself or if I'm too negative sometimes. English isn't my first language, as you may know, and I may start to take too much of the situation for granted. It's ok to call me out in those cases. I need to learn.

How would you like to proceed? Do you still want to do the implementation? Do you want to discuss it more?

sidred:
I think we both agree on how things look like on a high level, but in the implementation details I am not completely clear and I am heading down a few rabbit holes myself. So that's why I closed it. You seem to have a clearer view on the details.  Do you want to take a shot at it and then we'll see how things go? Does that make sense?

Ogeon:
It's easy to get snowed in when doing things like this, and we can only guess how it will actually play out when used by other people. My ideas really only are clear on some parts and very fuzzy on other parts. They are, as I have said before, based on how the color spaces are defined and how the type checker works. I'm trying to find a balance between not having to specify too many types and not introducing too many restrictions. I think there are many solutions to this problem and both of our views are valid solutions and good in their own way. We just have to manage to fuse the best parts in the best way possible.

I can most likely give it a try later today (I have some errands to run first) and we'll see how it looks. Feel free to start on some other stuff from the queue, in the meantime. Oh, and this may be too much to ask, so I'm not asking you to, but you are welcome to contribute to, or even take over, #57. It's mind-numbingly tedious and I wouldn't wish it on anyone, but I would be incredibly grateful for some help. Don't do it if you don't want to, I won't blame you at all. I'm just putting it out there :P

PR #60, opened 2016-03-10
Generalize the RGB types over RGB standards

### LINKED ISSUE #31: Common RGB struct
Have a common rgb struct 

```
struct RgbColor<RgbVariant, WP, T> {
    r: T,
    g: T,
    b: T,
    _color_space: PhantomData<RGBVariant>,
    _white_point: PhantomData<WP>,
}
```

then we can define the individual colors as 

```
const Rgb = RgbColor<LinearSrgbSpace, D65, T>
const Srgb = RgbColor<SrgbSpace, D65, T>
const AdobeRgb = RgbColor<AdobeSpace, D65, T>
```

conversions will be

```
impl<RgbVariant, WP,T: Float> From<RgbColor<RgbVariant, WP,T>> for Xyz<T> {
    fn from(rgb: Rgb<T>) -> Xyz<T> {
      let conv_matrix = RgbVariant::get_matrix();
      let wp_xyz = WP::get_xyz();


        Xyz {
           .... calculations
        }
    }
}

impl<RgbVariant, WP,T: Float> From<Xyz<T>> for RgbColor<RgbVariant, WP,T> {
    fn from(xyz: Rgb<T>) -> RgbColor<RgbVariant, WP,T> {
      let conv_matrix = RgbVariant::get_matrix();
      let wp_xyz = WP::get_xyz();


        Rgb {
           .... calculations
        }
    }
}

```

### LINKED ISSUE #58: Redesign the pixel encoding procedure
Something that has been growing in my mind for a few days is the idea of changing the way colors are encoded into pixel friendly formats. The current procedure may look like this:

``` rust
let linear = Rgb::new(1.0, 0.3, 0.6);
let srgb = Srgb::from_linear(linear);
let pixel: [u8; 3] = srgb.to_pixel();
```

or, using a shortcut:

``` rust
let linear = Rgb::new(1.0, 0.3, 0.6);
let pixel: [u8; 3] = Srgb::linear_to_pixel(linear);
```

There's also the case where a simple custom gamma exponent can be used:

``` rust
let linear = Rgb::new(1.0, 0.3, 0.6);
let pixel: [u8; 3] = GammaRgb::linear_to_pixel(linear, 2.2);
```

This is fine and all, but it's not really easy to use in a general way:
1. There's no common trait or type for encoded colors.
2. The alpha component is baked into the color type.
3. `GammaRgb` has a variable gamma exponent, which makes it very different from other encodings, and impossible to fully express as a simple type.

This problem is tightly connected to #31, where a general RGB type is proposed. The way `GammaRgb` is structures interferes with that proposal, since it has an additional `gamma` component, besides the usual `red`, `green` and `blue`. I'm therefore proposing that we redesign the way pixel encoding works to make it more friendly towards the idea of a more generalized RGB type. I would also like to extend this to other RGB related color spaces, such as HSV, HSL, Luma and HWB. These can all be seen as input types and they are, often enough, defined in terms of already encoded RGB.

The initial proposal is to keep linear and encoded types separate. The purpose of this library is to strongly encourage a linear work flow and this can easily be achieved by only implementing operations on the linear variant. There was a proposal, by @sidred, to have marker type parameters, such as `Rgb<Linear, Srgb, ...>` and `Rgb<Gamma, Srgb, ...>`, but [recent problems](https://github.com/Ogeon/palette/pull/56#issuecomment-189837611) with such marker parameters, as well as the additional complexity of extra type parameters, makes that method less attractive.

The separation of linear and encoded colors, using two different structs, will come with a bit of duplication, but the duplication will mainly be restricted to construction and fields. The linear variants will implement the operation traits, while the encoded variants will be focused on packing and converting data. Here is an example of how it may look when combined with #31:

``` rust
pub trait RgbEncoding {
    ///Encode a color component.
    fn encode<T: Float>(value: T) -> T;

    ///Decode a color component.
    fn decode<T: Float>(value: T) -> T;
}

pub trait RgbPrimaries {
    //get red, green, blue primaries, etc.
}

///Marker type, which both defines a set of primaries and an encoding
pub struct Srgb;

impl  RgbEncoding for Srgb {
    //...
}

impl  RgbPrimaries for Srgb {
    //...
}

struct LinearRgb<Prim: RgbPrimaries = Srgb, ...> {
    //..
    primaries: PhantomType<Prim>,
}

struct Rgb<Enc: RgbEncoding = Srgb, ...> {
    //..
    primaries: PhantomType<Enc>,
}
```

Encodings and primaries are often strongly connected to each other, which is why the same type can be used for both. The problem with this is that it's hard to decouple them, but that's hopefully rare enough to make this more good than bad. The only problem left is the `GammaRgb`. Its variable gamma component makes it incompatible with the above pattern, and having floating point values as type parameters is nothing but a (bad?) dream at this stage. We can, however, make more assumptions and add more escape hatches. I'm therefore proposing the addition of the probably most common gamma exponent, 2.2, as an encoding type, and the extension of the above `RgbEncoding` that makes it possible for an encoding to map to any set of primaries:

``` rust
pub trait RgbEncoding {
    type Primaries: RgbPrimaries;

    ///Encode a color component.
    fn encode<T: Float>(value: T) -> T;

    ///Decode a color component.
    fn decode<T: Float>(value: T) -> T;
}

///Gamma exponent 2.2, with any primaries.
struct Gamma22<P: RgbPrimaries>;

impl<P: RgbPrimaries> RgbEncoding for Primaries<P>
    type Primaries = P;

    //...
}

impl  RgbEncoding for Srgb {
    type Primaries = Srgb;
    //...
}

fn main() {
    let encoded = Rgb::<Gamma22<Srgb>>::new(...);
    let linear = encoded.to_linear(); //Gives LinearRgb<Srgb>
}
```

The preferred primaries can then be fetched from the associated `Primaries` type when converting from encoded to linear. Now, what about those other gamma values? That's where we need escape hatches. One such escape hatch could be to keep the old `GammaRgb` type around. It may be a bit unwieldy, but it works. An other is to just have a straight-to-pixel conversion, like the `*::linear_to_pixel` function in the current design. More suggestions are welcome here.

The last question is what encoded types are allowed to do. My proposal is "basically nothing", since I'm still strongly in favor of keeping them strictly separated from linear types. This is to minimize any risk of mistaking non-linear values for linear and forcing the user to make an explicit conversion between them. The minimal set of functions I can think of at the moment is:

``` rust
impl<E: RgbEncoding, ...> Rgb<E, ...> {
    fn new(red: T, green: T, blue: T) -> Rgb<E, ...> {...}
    fn new_u8(red: u8, green: u8, blue: u8) -> Rgb<E, ...> {...}
    fn from_pixel<P: RgbPixel<T>>(pixel: P) -> Rgb<E, ...> {...}
    fn to_pixel<P: RgbPixel<T>>(&self) -> P {...}
    fn from_linear<C: Into<LinearRgb<E::Primaries, ...>>(color: C) -> Rgb<E, ...> {...}
    fn to_linear<C: From<LinearRgb<E::Primaries, ...>>(&self) -> C {...}

    //Useful addition
    fn linear_to_pixel<C: Into<LinearRgb<E::Primaries, ...>, P: RgbPixel<T>>(color: C) -> P {
        Rgb<E, ...>::from_linear(color).to_pixel()
    }
}
```

These can, of course, be tweaked and changed. Some of these functions would be nice as a general encoding trait, but that's quite complex and may be better discussed at a later stage. 

Something this proposal hasn't touched, so far, is what to do with the `alpha` component. I think it would be best to make use of the `Alpha` type, but I haven't figured out exactly how, so that question remains open.

An other open question is type aliases for common RGB types, such as AdobeRgb. One idea is to not alias sRGB, and only implement `new` for `Rgb<Srgb, D65, T>`, similar to how it's done with only `D65` today. An other is to alias all of them and not pick a default.

### LINKED ISSUE #66: HSL in sRGB space
I may be reading things wrong in the specs, but I believe that the version of HSL that CSS supports is in fact HSL in sRGB space, not linear HSL.

It'd be nice if we had an option to store this type of HSL without hackishly converting an `Hsl` to an `Rgb` and then copying the components from the `Rgb` into an `Srgb`. Again, if I'm wrong and this is actually linear `Hsl`, feel free to ignore this use case.

Ogeon:
This changes the `Rgb` type to have a type parameter that implements the `RgbStandard` trait. This trait defines a set of primaries, white point and transfer function. It does also make the RGB types more uniform and type safe.

Closes #66, closes #31, closes #58


Ogeon:
This is a beginning. It doesn't compile, yet, because of problems with the definition of `IntoColor` and `FromColor`. These have to be redefined and I'm too tired right now to come up with a solution. The problem is that the `RgbSpace` needs to be known to convert to `Rgb`, but the whole trait can't depend on it (like in the current implementation), because then it won't be known when converting between non-RGB types, like XYZ -> Lab.

I'll write a better description later, but I decided to bundle the primaries and white point in the `RgbSpace` trait, and bundle it, together with a transfer function, as `RgbStandard`.

src/pixel/mod.rs
```diff
@@ -10,6 +10,15 @@ pub use self::gamma_rgb::GammaRgb;
 mod srgb;
 mod gamma_rgb;
 
+///A transfer function to and from linear space.
+pub trait TransferFn {
```
sidred: It will be good to add rgb_to_linear and rgb_from linear with default implementation, one the linear and gamma rgb are defined.

Ogeon: To `TransferFn` or just to the `pixel` module?

sidred: To the TransferFn. Hsl to LinearHsl, should convert though rgb. So that is the only conversion TransferFn should care about.

src/matrix.rs
```diff
@@ -76,11 +78,47 @@ pub fn matrix_inverse<T: Float>(a: &Mat3<T>) -> Mat3<T> {
     [d0/det, -d3/det, d6/det, -d1/det, d4/det, -d7/det, d2/det, -d5/det, d8/det]
 }
 
+///Geneartes to Srgb to Xyz transformation matrix for the given white point
+pub fn rgb_to_xyz_matrix<S: RgbSpace, T: Float>() -> Mat3<T> {
```
sidred: ```rgb_to_xyz_matrix``` and ``` mat3_from_primaries``` should be default implemented in the RgbSpace trait. Then we will be able to manually add the matrix for known conversions like Srgb to Xyz, without recalculating the transform matrix every time.

src/pixel/mod.rs
Ogeon: It can also apply to Luma/Grey.

src/matrix.rs
Ogeon: That's a good idea. Will do.

Ogeon:
Looks like the workaround works for now. The only problem is the occasional detour through XYZ when going from `Rgb<A, T>` to `Rgb<A, T>`. I don't think we can specialize for that.

src/pixel/mod.rs
sidred: We need a trait to convert gamma encoded colors to linear colors (and the inverse). 

```let rgb = linear_rgb.from_linear()``` currently does not work. If we do not want to add another trait, this is where these conversions will go for every type that can convert to and from linear.

Or we will need something similar to the From and Into traits that will leverage the transfer function to do the linear <-> non-linear conversions.

Comment (Ogeon) [inline src/pixel/mod.rs]: It's a bit hard to generalize that in the `TransferFn` trait, since there are other parameters involved. There's currently the `Srgb::from_linear(color)` method, which should contain all the necessary information to convert anything to `Srgb`.

`let rgb = linear_rgb.from_linear()`, which I guess is the same as `let rgb = linear_rgb.into_nonlinear()`, will always need extra information about the transfer function. That can't be changed by changing the `TransferFn` trait.

**Comment by Ogeon** [comment]
`pixel::Srgb` is now almost completely replaced. There are still some missing conversion implementations, but they are on their way. I just have to figure how to do it in a good way. Possibly another `From` vomiting macro.

Anyway, it looks like it fits right into the previous model, with a few small adjustments.

**Comment by Ogeon** [comment]
I ended up implementing a few conversions manually and modify the macros for the rest. It's now possible to freely convert `Rgb` and linear types using `From` and `Into`.

The next phase will probably be the integration of the other RGB related types. I'm just not sure how they should be presented to the user. Just as plain `Hsl<S, T>` or with accompanying type aliases, like `HslSrgb`? Adding aliases doesn't hurt, except for the clutter factor, but it's relatively low if they are limited to the `Srgb` variants.

**Comment by Ogeon** [comment]
The previously mentioned round trip when converting from `LinRgb<A, T>` to `LinRgb<A, T>` caused infinite recursion when `Hsv` joined the club. The problem was that `Hsv<A, T> -> LinRgb<A, T>` would be treated as `Hsv<B, T> -> LinRgb<A, T>`, which expands to  `Hsv<B, T> -> LinRgb<B, T> -> LinRgb<A, T> -> Hsv<A, T> -> LinRgb<A, T>`, which keeps exploding.

The solution I'm using to break the recursion is to check if `A::Primaries == B::Primaries`, by comparing their type IDs, and not do any conversion if they are the same. We already know that their white points are the same, since that's a requirement, so they should represent the same RGB space, but possibly with different actual type (e.x. `Srgb == (Srgb, D65)`). I expect this check to be optimized away after monomorphisation, and would otherwise most likely be cheaper than always going through XYZ, so I would say it's a win in any case.

File: src/rgb/mod.rs [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+///Nonlinear sRGB with an alpha component.
+pub type Srgba<T = f32> = Rgba<standards::Srgb, T>;
+
+///Linear sRGB.
+pub type LinSrgb<T = f32> = LinRgb<standards::Srgb, T>;
+///Linear sRGB with an alpha component.
+pub type LinSrgba<T = f32> = LinRgba<standards::Srgb, T>;
+
+///An RGB space and a transfer function.
+pub trait RgbStandard {
```
sidred: Should the RgbStandard have the WhitePoint type as well? The usecase I see for it is supporting non standard white points like ICC Srgb D50 \<SrgbTranfFn, SrgbSpace, D50\>.

src/rgb/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    type WhitePoint: WhitePoint;
+}
+
+impl<P: Primaries, W: WhitePoint> RgbSpace for (P, W) {
+    type Primaries = P;
+    type WhitePoint = W;
+}
+
+///Represents the red, green and blue primaries of an RGB space.
+pub trait Primaries: Any {
```
sidred: Whats the advantage of Any over Sized? I looked through the std documentation but couldn't find anything.

src/rgb/linear.rs
```diff
@@ -44,84 +47,83 @@ pub struct Rgb<Wp = D65, T = f32>
...
-impl<Wp, T> Rgb<Wp, T>
+impl<S, T> LinRgb<S, T>
     where T: Float,
-        Wp: WhitePoint<T>
+        S: RgbSpace,
 {
     ///Linear RGB.
-    pub fn with_wp(red: T, green: T, blue: T) -> Rgb<Wp, T> {
-        Rgb {
+    pub fn with_wp(red: T, green: T, blue: T) -> LinRgb<S, T> {
```
sidred: Bit of a nitpick. Needs a better name than ```with_wp``` that we can use for all colors. Maybe something like ```with_color``` or ``` with_type```.

src/rgb/mod.rs
Ogeon: That's still possible, since `RgbStandard` is implemented for `(Srgb, D50, Srgb)`.

src/rgb/mod.rs
Ogeon: `Any` is necessary for `TypeId::of` to work. I would have used `Reflect` instead, but it's still unstable.

src/rgb/linear.rs
Ogeon: Yeah, I decided to leave them like that for now, and think about a better alternative. Maybe we should even make `new` completely generic for these types. It works if they are used through type aliases.

src/rgb/standards.rs
```diff
@@ -0,0 +1,53 @@
...
+use flt;
+use white_point::{WhitePoint, D65};
+use rgb::{Primaries, RgbSpace, RgbStandard};
+use pixel::TransferFn;
+
+///The sRGB color space.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Srgb;
+
+impl Primaries for Srgb {
```
sidred: The white points listed for Srgb are using the D65 space. So the white point needs to be hard coded to use D65. This is the reason why I was suggesting the RgbStandard have a separate white point. Then we can use the chromatic adaptation to automatically convert. Not sure if there is a better way to implement this.

Here is a link that shows the transformation from Srgb to ICC Srgb D50
http://ninedegreesbelow.com/photography/srgb-color-space-to-profile.html#sRGB-profile-source-destination-white-points

Ogeon: I'm afraid I may be misunderstanding the problem. Are you talking about hard coding a transformation matrix from D65 to D50, and another for staying within D65? Are you suggesting something like this:

```rust
pub trait RgbStandard {
    type Space: RgbSpace;
    type WhitePoint: WhitePoint;
    type TransferFn: TransferFn;
}
```

or something like this:

```rust
pub trait RgbStandard {
    type Primaries: Primaries;
    type WhitePoint: WhitePoint;
    type TransferFn: TransferFn;
}
```

or something else? Would you mind giving an example?

sidred: The main issue is that the Srgb primaries definition is incorrect.

I was thinking of this
```{rust}
pub trait RgbStandard {
    type Primaries: Primaries;
    type WhitePoint: WhitePoint;
    type TransferFn: TransferFn;
}
```
The white point of the primaries have no relation to the color space white point. The primaries trait needs its own white point 
```{rust}
pub trait Primaries {
    type Wp: Whitepoint

    fn red< T: Float>() -> Yxy<Wp, T>;
    fn green<T: Float>() -> Yxy<Wp, T>;
    fn blue< T: Float>() -> Yxy<Wp, T>;
}
```

Srgb primaries will be implemented as

```{rust}
impl Primaries for Srgb {
   type Wp = D65;
    fn red< T: Float>() -> Yxy<Wp, T> {
      Yxy::with_wp(flt(0.6400), flt(0.3300), flt(0.212656));
    }
    ...
```

We could also implement the white point as follows, but it will make the implementation a bit cumbersome. 

I'd rather have a simple definition of the white point and let the library handle the adaptations where necessary
```{rust}
impl Primaries for Srgb {
   type Wp = D65;
    fn red< T: Float>() -> Yxy<Wp, T> {
      let yxy: Yxy<D60, T> = Yxy::with_wp(flt(0.6400), flt(0.3300), flt(0.212656));
      if Wp == D65 {
        yxy
      } else {
        yxy.adapt_into()
      }
    }
    ...
```

src/hsl.rs
```diff
@@ -36,60 +40,70 @@ pub struct Hsl<Wp = D65, T = f32>
...
             hue: hue,
             saturation: saturation,
             lightness: lightness,
-            white_point: PhantomData,
+            space: PhantomData,
+        }
+    }
+
+    #[inline]
+    fn reinterpret_as<Sp: RgbSpace>(self) -> Hsl<Sp, T> {
```
sidred: I don't get this. Is this function saying that ```Hsl<Srgb, T>``` = ```Hsl<AdobeRgb, T>``` with the same value for the components?

src/hsl.rs
```diff
@@ -150,15 +167,21 @@ impl<Wp, T> FromColor<Wp, T> for Hsl<Wp, T>
...
             lightness: l,
-            white_point: PhantomData,
+            space: PhantomData,
         }
     }
 
-    fn from_hsl(hsl: Hsl<Wp, T>) -> Self {
-        hsl
+    fn from_hsl<Sp: RgbSpace<WhitePoint=Wp>>(hsl: Hsl<Sp, T>) -> Self {
+        if TypeId::of::<Sp::Primaries>() == TypeId::of::<S::Primaries>() {
```
sidred: I am not sure if this is even possible, but can this be implemented using the PartialEq trait
```{rust}

 +impl<S, Wp, T> FromColor<Wp, T> for Hsl<S, T>
     where T: Float,
        S: RgbSpace<WhitePoint=Wp> + PartialEq,
        Wp: WhitePoint,
{
     ...
    fn from_hsl<Sp: RgbSpace<WhitePoint=Wp> + PartialEq>(hsl: Hsl<Sp, T>) -> Self {
    ...
}
```

Ogeon: That would need us to instantiate `Sp` and `S`, which would need `Default` or some special function that does the same thing.

src/hsl.rs
Ogeon: It's there to transfer the values from one type to the other. The thing is that `S` and `Sp` are considered different, even if they are the same types. This is similar to `transmute`, but for types, and I'm keeping it private for now, since it's a bit sketchy from a semantic perspective.

src/rgb/standards.rs
Ogeon: >I was thinking of this
>
>```rust
pub trait RgbStandard {
    type Primaries: Primaries;
    type WhitePoint: WhitePoint;
    type TransferFn: TransferFn;
}
```

That's essentially the same thing as what we already have. `Primaries` and `WhitePoint` are currently just grouped.

>The white point of the primaries have no relation to the color space white point. The primaries trait needs its own white point 

Does the white point even matter at all, in this case? I mean, yxY values are the same anyway, right? Or does sRGB + D50 imply that _the whole_ space is transformed into the D50 space, instead of just changing the white point? The details are not really clear to me...

Comment (sidred) [inline src/rgb/standards.rs]: Look at http://ninedegreesbelow.com/photography/srgb-color-space-to-profile.html#sRGB-profile-source-destination-white-points Table 6 and Table 7. The XYZ values are different.

Without the white point in the primaries, we don't know if chromatic adaptation is required or not for the primaries. Example ```Rgb<Srgb, Srgb, D65>```, no adaptation required since white point and primaries (white point) match, while ```Rgb<Srgb, Srgb, D50>``` will require adaptation.

Comment (Ogeon) [inline src/rgb/standards.rs]: I'll look at it when I get home again, or tomorrow, but if that's the case, then this makes things a lot more complicated. This births the question "how do we differ sRGB, but adapted to D50, from sRGB primaries + D50 white point?".

Comment (sidred) [inline src/rgb/standards.rs]: > "how do we differ sRGB, but adapted to D50, from sRGB primaries + D50 white point?".

They are 2 separate types

```{rust}
pub struct Srgb;

impl RgbSpace for Srgb {
    type Primaries = Srgb;
    type WhitePoint = D65;
}

pub struct IccSrgb;

impl RgbSpace for IccSrgb {
    type Primaries = Srgb;// primaries will be D65 and need to be adapted when used
    type WhitePoint = D50;
}

```

src/hsl.rs
sidred: Would it be better to add RgbSpace bound to the FromColor trait? It will remove the need for reinterpret_as function.

```
pub trait FromColor<S, Wp, T> for Hsl<S, T>
     where T: Float,
        S: RgbSpace<WhitePoint=Wp>
        Wp: WhitePoint,
```

src/rgb/standards.rs
Ogeon: >```rust
impl RgbSpace for Srgb {
    type Primaries = Srgb;
    type WhitePoint = D65;
}
```

That's still good old sRGB. What I meant was the sRGB primaries with any other white point. No adaption, just a different combination.

Comment (sidred) [inline src/rgb/standards.rs]: > What I meant was the sRGB primaries with any other white point. No adaption, just a different combination.

I am not sure I understand what you are trying to say. The Srgb primaries are defined with the D65 white point. I have not seen any other definitions for the primaries. For use with any other white point, chromatic adaptation for the primaries is required.

Comment (sidred) [inline src/rgb/standards.rs]: > how do we differ sRGB, but adapted to D50, from sRGB primaries + D50 white point?

These are the same right? What is the difference.

Comment (Ogeon) [inline src/rgb/standards.rs]: I'm thinking of the situation where the user decides to invent their own RGB space by combining the primaries from sRGB with an other white point. It's not sRGB, in the same sense that AdobeRGB isn't sRGB (even if they share two primaries and the white point).

>These are the same right? What is the difference.

No, because the primaries are defined as xy values, which, as you know, is only perceptively affected by the white point. You would only adapt them to make them _look_ the same under a different light. This is a case where that's not the goal. I'm still not sure if this is a meaningful discussion and I should really read that link before I spiral away somewhere.

Comment (sidred) [inline src/rgb/standards.rs]: Thinking more about this, there is only ```Rgb<SrgbPrim<D65>, D65>``` or ```Rgb<SrgbPrim<D50>, D50>```. ```Rgb<SrgbPrim<D65>, D50>``` does not exists, but what we are trying to say  is that we have know values of the Srgb primaries which can be converted into ```Rgb<SrgbPrim<D50>, D50>``` via chromatic adaptation. Does that make sense? When we change the white point everything including the primaries converts into the new white point is my understanding

Comment (Ogeon) [inline src/rgb/standards.rs]: Yes, that's how I understood your reasoning. That they look the same, but the virtual light source has changed. The scenario I'm thinking about is the possible invention of `Rgb<SrgbPrim<D65>, D50>` (or any other combination of existing parts) and how to express it as a type. I think that we should leave as many doors open as possible for the users to extend the set of colors. We can't encode every single possible combination, but we can allow the users to add the ones they may need.

Comment (Ogeon) [inline src/rgb/standards.rs]: Ok, I have read some of it now, but I think I get the point. It describes how to adapt the sRGB primaries to D50, to prepare them for being included in an ICC profile. The point of that is, as far as I know, to preserve their perceived colors in relation to what's white. That doesn't _necessarily_ mean that we have to do the same with `Primaries`, unless we want them to be perceptively connected to some white point. We may also choose to not do that, and only care about their numerical values (preserving xyY, but losing the perceptive values). The question is what's best when representing the primaries as a type, like this. Chromatic adaption is a bit runtime expensive, but what's the alternative if you want an ICC adapted space?

Comment (sidred) [inline src/rgb/standards.rs]: If I am not mistaken easyrgb and lindbloom calculators adapt the primaries, so I think  chromatic adaptation is the way to go.

One issue I currently see is the the AdaptFrom does not check if the source and destination white points are equal and will cause D65 -> D65 changes. I guess we should add a test for that.

Comment (Ogeon) [inline src/rgb/standards.rs]: >If I am not mistaken easyrgb and lindbloom calculators adapt the primaries, so I think chromatic adaptation is the way to go.

Yeah, we should look into that. They are a bit more specialized, though. The easyrgb is more or less a black box, but we can poke it from the outside. Lindbloom's calculator is all JavaScript, so the source can easily be inspected through the browser.

>One issue I currently see is the the AdaptFrom does not check if the source and destination white points are equal and will cause D65 -> D65 changes. I guess we should add a test for that.

We could do the `TypeId` comparison there, as well. It's may give false negatives, but only when different types represents the same values. I think that's acceptable at the moment.

Comment (Ogeon) [inline src/rgb/standards.rs]: I have finally had some time to look into how the color calculators work, and I have found that easyrgb seems to ignore the white point when converting from RGB to XYZ/yxY and then uses it when converting to Lab, etc. So, it's `Srgb<D65> -> Xyz<?> -> Lab<Wp>`. I don't know if there's any adaption involved here, but it may very well be. The way I checked was the a tried to convert RGB(1, 0, 0) with D65 and D50 and saw that the XYZ and yxY values didn't change. I changed to RGB(1, 1, 1) and saw the same result. It was locked to a single white point.

Lindbloom's calculator is easier to inspect. The JavaScript source can be found [here](http://www.brucelindbloom.com/javascript/ColorConv.js) and the most interesting function is RGB2XYZ:

```javascript
function RGB2XYZ()
{
	var R = InvCompand(RGB.R);
	var G = InvCompand(RGB.G);
	var B = InvCompand(RGB.B);
	
	XYZ.X = R * MtxRGB2XYZ.m00 + G * MtxRGB2XYZ.m10 + B * MtxRGB2XYZ.m20;
	XYZ.Y = R * MtxRGB2XYZ.m01 + G * MtxRGB2XYZ.m11 + B * MtxRGB2XYZ.m21;
	XYZ.Z = R * MtxRGB2XYZ.m02 + G * MtxRGB2XYZ.m12 + B * MtxRGB2XYZ.m22;
	
	if (AdaptationMethod != 3)
	{
		var Ad = RefWhite.X * MtxAdaptMa.m00 + RefWhite.Y * MtxAdaptMa.m10 + RefWhite.Z * MtxAdaptMa.m20;
		var Bd = RefWhite.X * MtxAdaptMa.m01 + RefWhite.Y * MtxAdaptMa.m11 + RefWhite.Z * MtxAdaptMa.m21;
		var Cd = RefWhite.X * MtxAdaptMa.m02 + RefWhite.Y * MtxAdaptMa.m12 + RefWhite.Z * MtxAdaptMa.m22;
		
		var As = RefWhiteRGB.X * MtxAdaptMa.m00 + RefWhiteRGB.Y * MtxAdaptMa.m10 + RefWhiteRGB.Z * MtxAdaptMa.m20;
		var Bs = RefWhiteRGB.X * MtxAdaptMa.m01 + RefWhiteRGB.Y * MtxAdaptMa.m11 + RefWhiteRGB.Z * MtxAdaptMa.m21;
		var Cs = RefWhiteRGB.X * MtxAdaptMa.m02 + RefWhiteRGB.Y * MtxAdaptMa.m12 + RefWhiteRGB.Z * MtxAdaptMa.m22;
		
		var X = XYZ.X * MtxAdaptMa.m00 + XYZ.Y * MtxAdaptMa.m10 + XYZ.Z * MtxAdaptMa.m20;
		var Y = XYZ.X * MtxAdaptMa.m01 + XYZ.Y * MtxAdaptMa.m11 + XYZ.Z * MtxAdaptMa.m21;
		var Z = XYZ.X * MtxAdaptMa.m02 + XYZ.Y * MtxAdaptMa.m12 + XYZ.Z * MtxAdaptMa.m22;
		
		X *= (Ad / As);
		Y *= (Bd / Bs);
		Z *= (Cd / Cs);
		
		XYZ.X = X * MtxAdaptMaI.m00 + Y * MtxAdaptMaI.m10 + Z * MtxAdaptMaI.m20;
		XYZ.Y = X * MtxAdaptMaI.m01 + Y * MtxAdaptMaI.m11 + Z * MtxAdaptMaI.m21;
		XYZ.Z = X * MtxAdaptMaI.m02 + Y * MtxAdaptMaI.m12 + Z * MtxAdaptMaI.m22;
	}
}
```

It linealizes the RGB, converts it to XYZ, and _then_ performs chromatic adaption. This is essentially the same as what easyrgb does, but with a chromatic adaption step after conversion.

Lindbloom's calculator is closes to what we are trying to do, since it supports "custom" transfer functions. The way I understand it is that it would be the same as this, with the current implementation:

```rust
let xyz = Rgb::<(Srgb, TransferFn)>::new(r, g, b).into_xyz();
let final_xyz: Xyz<Wp> = xyz.adapt_into();
```

So, the question still remains: What's the meaning of `LinRgb<(Srgb, D50)>`? Do we force chromatic adaption of the primaries or not? I'm voting "no". I think we can still solve the ICC problem, but forcing adaption will close the door to inventing new RGB spaces by combining existing primaries with another white point.

sidred: That's really surprising.

Because in this page http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html, it shows the transformation matrix for Srgb D50 which exactly matches the http://ninedegreesbelow.com/photography/srgb-color-space-to-profile.html#sRGB-profile-source-destination-white-points  table7 and also the icc profile values. Theae are calculated through chromatic adaptation of the Srgb primaries. 

I wonder if the two methods give the same result.

Ogeon: That's interesting. RGB is an interpolation between the white point and the primaries, so maybe that's just how it works. I haven't done the math, though.

Ogeon:
Hmm... this looks like it's close to being complete, judging from the diff. Just need to

- [ ] fix the conflict,
- [ ] see if the primaries/whitepoint relation is something that requires changes,
- [ ] find a better name than `with_wp`,
- [ ] and check how the docs looks.

Ogeon:
I have been making an attempt to unify `Rgb` and `LinRgb`, trying to get rid of both `LinRgb` and `GammaRgb` in the end, but I'm afraid the `Color` type stands in my way. Either that or I'm too tired and confused to realize why things don't work as they should. I'll see if I can make some progress later. Otherwise I'm saving it for another time.

Ogeon:
Seems like it's going to work, after all, so I'm adding a point to the checklist.

Ogeon:
Ok, I think I'll consider this done, at last. I skipped a couple of points from the list and decided to push them into the future.

PR #61, opened 2016-03-11
Restore the proper scale of Lab and Lch

### LINKED ISSUE #49: Lab and Lch values

sidred:
This restores the scale of Lab and Lch to be the same as in all of the specifications within reach. That is:

- Lab::l and Lch::l goes from 0 to 100,
- Lab::a and Lab::b goes from -128 to 127,
- Lch::chroma goes from 0 to about 128-181.

Also added additional tests for Lab <-> Lch and Lab<->Xyz conversions. Removed tests for lab and lch from the color mine data as it is not accurate.

This is a breaking change, since everything that uses these values directly will be wrong. Closes #49.


sidred:
I have also added tests from the color checker data http://www.babelcolor.com/colorchecker-2.htm

This has additional data for RgbSpaces to Cie tests. Once the rgb standards are added, these can be used as a comprehensive test for gamma encoded color space to cie tests.

src/lab.rs
```diff
@@ -401,22 +401,22 @@ mod test {
     #[test]
     fn red() {
         let a = Lab::from(Rgb::new(1.0, 0.0, 0.0));
-        let b = Lab::new(53.23288 / 100.0, 80.09246 / 128.0, 67.2031 / 128.0);
-        assert_relative_eq!(a, b, epsilon = 0.0001);
+        let b = Lab::new(53.23288, 80.09246, 67.2031);
+        assert_relative_eq!(a, b, epsilon = 0.01);
```
Ogeon: Does the test precision need to be this low?

Ogeon:
This is great, and look at those decimals! I think we are set for life with that precision, given that it's also correct. I'm sure it is.

src/lab.rs
sidred: I don't know the source of this data and this does not pass with 0.001. Anyway the Lab - Rgb should be taken care of in the color checker data, which is more accurate.

sidred:
Yeah, but I was hoping it would be within the default epsilon :)

src/lab.rs
Ogeon: I seem to remember that I recalculated those numbers in my attempt at this, but we can leave it as it is. It's still checked in other tests, as you said.

Ogeon:
> `epsilon = 0.000000000001`

:heart_eyes:

Sorry, it's just so refreshing after the blunt xyY numbers from color mine. Are we ready to go?

sidred:
Yes

Ogeon:
Great! @homu r+

Issue #64, opened 2016-10-31
Add raw light spectrum to XYZ conversion


tatref:
Hi!

Would it be interesting to add support for spectrum to XYZ conversions?

The spectrum can be represented as a `[f32; XXX]` (XXX depending on the sampling of the spectrum).  Where the `f32` values are the intensities (in arbitrary units), for each wavelength from ~400 to ~800 nm. Then the XYZ coordinates are a matter of simple matrix multiplication, given the XYZ weights.

The weights are available on different websites (http://www.cvrl.org/cie.htm)
There is also analytic aproximations (http://jcgt.org/published/0002/02/01/paper.pdf)

I've already started a POC, so I might submit a pull request (the code is not that clean yet though).

Tell me what you think about this.

Ogeon:
Hello, and sorry for my late response! I was at work and wanted to take my time to write a proper answer. Anyway, this sounds like something that could perhaps fit in. It could also work as a separate library. I did a similar implementation once, and I think it was relatively small, so I guess it won't be a maintenance problem, but it depends on how it will work. (Just thinking about the practical details :smile:) Speaking of which, a few things I'm going to request is that it's:

1. self contained, meaning that it keeps track of its own limits and parameters,
3. should catch user mistakes as early as possible (preferably at compile time),
2. fault tolerant, when it can't catch the mistake,
1. at least generic over the float type (preferably as generic as the XYZ type),
2. reasonably bare bones.

I realize that it's tricky and it's no big deal if the POC doesn't fulfill it. Take it one step at the time. It's a POC, after all, and I'm happy to give feedback when I have the time. The reason for this is that I want it to be as easy to use as possible, without relying om too many assumptions (it's a balance act), and I want it to be easy to overlook and maintain in the future. Part of the new order here is also that I'm trying to be more clear with my expectations, and these are the things I can think of right now that I'm usually working towards. Feel free to ask if you want me to elaborate.

Also a few question, out of curiosity: Is the idea that the input spectrum has to match the number of weights, or is there some kind of interpolation in place? Would it make sense to to combine it with the built in gradient code?

I would also love to see the POC, even if it isn't finished. I don't mind, and I have nothing against using pull requests for feedback and as a scratch space. It would also make it easier for me to make a proper decision sooner, rather than later.

tatref:
Ok so here's the POC, but I couldn't get the generic type working (I can't find a way to convert from `Float` back to `f32`)...

[...]

PR #65, opened 2016-11-01
WIP initial implementation of spectrum to xyz conversion

tatref:
WIP for issue #64 

I tried to make the conversion generic over Float, but I couldn't make it work (new to rust...)
Let me know what you think about this.


meh:
For the generic part you can use `num::cast` to convert from/to numeric types.

meh:
If you moved the table from a const to a generic function you could wrap all the numbers into `num::cast`s, the compiler should be able to optimize it away to a constant, and you could avoid converting the table each time.

tatref:
Thanks for the tips!

I don't quite get your last comment though... Do you mean a function like `get_map<T: Float>() -> [T; N_SAMPLES]`?

Also after checking a second time, it seems that all the data in the table is only `f32`. I'm not sure if using `f64` would provide any advantage?

meh:
Yes, that would be it.

The advantage is you avoid iterating over the table and converting it each time, so it's a performance and genericity win.

All the data might be `f32`, but the operations should be on the given type to avoid precision loss.

tatref:
Humm I don't understand how the function is supposed to work.

If I do the following, I get `expected type parameter, found floating-point variable`:
```
fn get_map<T: Float>() -> [[T; 3]; SPECTRUM_SAMPLES] {
    [   
    [0.000129900000,0.000003917000,0.000606100000],
    ...
```

On the other hand, if I convert each value in the get function, I'm loosing all the point, isn't it?

meh:
you wrap each one of them in a `num::cast(n).unwrap()`.

Ogeon:
Great! I have just skimmed through it, as I'm on my way to work, but it looks like a nice and simple solution. I'll take a closer look when I get home, and comment some more, but I can say that it's ok to let the table values have a specific type, and then convert them to the generic type as mentioned above. There is a function in the crate root, called `flt`, that makes it a bit shorter.

src/spectrum.rs
```diff
@@ -0,0 +1,143 @@
+/// ! Conversions from spectrum intensities to XYZ
+///
+
+use xyz::Xyz;
+use num::{Float,Zero};
+
+
+/// Data from http://www.cvrl.org/cmfs.htm, CIE 1931 2-deg, XYZ CMFs
+/// 360 nm -> 830 nm += 5nm
+const SPECTRUM_TO_XYZ_MAP: [[f32; 3]; 95] = [
```
Ogeon: I wonder if it would be better to use an `[(f32, f32, f32); 95]`. It would at least spare the compiler the work of optimizing the bounds checks away, during the multiplication. I'm not sure if it makes a difference.

meh: I'm fairly sure it makes no difference, the compiler already knows the size of the array.

Ogeon: Sure, what I meant was that indexing will always insert a bounds check AFAIK, but the optimization step can then remove it. The question is how accurate it is. I suppose it's nothing to worry about, but I like tuples or structs more for these kind situations. Structs are even nicer because of the named fields. Anyway, there's no need to change if it doesn't matter. :smile:

tatref:
I finally managed to add generic type for the spectrum. I had to add a bound to `T: AddAssign`, because otherwise, I couldn't use `+=`. Isn't the `Float` type supposed to provided that, since it already provide `Add`?

The gradient is also a good idea. I think we will have to provide a function outside of `Spectrum` with a signature like `fn to_xyz(data: BtreeMap<lambda: float,  intensity: float>) -> Xyz`

I will also add some doc + examples

meh:
`num::Float` only provides `Add`, that's why it would fail, but it's a really bad idea to add the `AddAssign` bound, just avoid using `+=`, sadly writing numerically generic code is awful in Rust.

Ogeon:
> I've finally managed to add generic type for the spectrum.

Nice! It looks good, but the convention is to import things, instead of prefixing with `::`.

> I had to add a bound to T: AddAssign, because otherwise, I couldn't use +=. Isn't the Float type supposed to provided that, since it already provide Add?

No, it's not provided, as @meh mentioned. It's thereby an unnecessary limitation to use it if we can do the same thing with `x = x + ...`, since it requires `T` to implement it. It's also not consistent with how other parts of the library looks.

> The gradient is also a good idea. I think we will have to provide a function outside of Spectrum with a signature like fn to_xyz(data: HashMap<lambda: float, intensity: float>) -> Xyz. What do you think?

Well, all of this could be compacted into a function like that :smile:, since that's all it does, which is fine. I don't know what else it needs. There could, for example, be a trait that specifies a Spectrum API, which maps wavelength to intensity. Then anything from slices to `Gradient`, to hashmaps could implement it and be used in the same `spectrum_to_xyz` (or whatever) function.

Ogeon:
Hi @tatref, it looks like I may have failed to notify you properly (I messed up and wrote the wrong usename in my first response to #70). I hope you don't mind that someone picked this up in a separate PR, and I'm sorry for letting this one stagnate. I'm still grateful for your help! Thank you!

I'll close this PR, since they are duplicates and the other one is based on your changes, but feel free to team up in #70 if you have any input or ideas.

PR #67, opened 2017-04-30
Fix typo

Joshua-Wright:
Space is needed so that GitHub markdown renders the subsection heading correctly


Ogeon:
Nice! Thank you!

PR #68, opened 2017-06-11
Fix failing Srgb doc test

jzhu98:
Just a small patch to use the approx crate's `assert_relative_eq!` macros in the failing Srgb doc test.

Because who doesn't like those green Travis badges?


Ogeon:
Hi, sorry for my delayed response. I blame having been sick and then just having too much stuff in my head.

This is really nice! Thank you! And yes, those green badges are indeed a desirable goal!

@homu r+

Ogeon:
@homu r+

Ogeon:
No Homu? I guess I have to do it myself, then.

PR #69, opened 2017-07-17
Migrate tests to Serde 1.0

jzhu98:
Since `rustc-serialize` is deprecated, I've updated the tests to use Serde 1.0 (as well as upgrading the other dev-dependencies.) Serde had one-to-one replacements for rustc-serialize's functions and #[derive] macros, so the conversion was pretty painless.

While migrating, I also fixed several csv headers, replacing some errant spaces with `_` in the `color_checker_data` data and fixing a missing `yxy_luma` header in `tests/convert/data_cie_15_2004.csv`. I think these errors were silently ignored by rustc-serialize until I switched to serde, but I'm not familiar enough with their mechanics to understand why.


Cargo.toml
```diff
@@ -29,11 +29,12 @@ version = "0.7"
 optional = true
 
 [dev-dependencies]
-image = "0.7"
-clap = "1"
-csv = "0.14"
-rustc-serialize  = "0.3"
-lazy_static = "0.1"
+image = "0.14.0"
+clap = "2.24.0"
```
Ogeon: I think we can safely set this to be just "2".

Ogeon: The same for the serde dependencies. That way there's no need to update unless we want any breaking changes.

jzhu98: Done!

jzhu98:
I've squashed the fixup commit and truncated the dependency versions. Should be good to go.

Ogeon:
Nice! Thank you!

PR #70, opened 2017-07-21
Add a Spectrum type

acolley:
This is a PR for issue #64.

I had a need for a Spectrum class for a Ray Tracer project I'm working on and I noticed the existing PR had gone stale about a year ago. I didn't know if there was a way to edit that existing PR so I made this one.

I've based most of it off of the PR by tatref: https://github.com/Ogeon/palette/pull/65. In addition I added the functions mentioned in the comments on that PR including the generic function for the mapping LUT and what I assumed was the 'gradient' method, which creates a sampling of a given input mapping wavelength to intensity using linear interpolation.

I've also added error handling using the error-chain library.

This adds two new dependencies to the library: ordered-float and error-chain.

One thing I wasn't sure of: what should the name of the 'gradient' method be? Currently it's called 'from_sparse' but I don't think that really reflects what it's doing.

The testing is also not perfect and I couldn't think of a great way of testing this sort of thing except some simple cases.


src/spectrum.rs
```diff
@@ -0,0 +1,375 @@
...
+pub struct Spectrum<T: Float> {
+    data: [T; SPECTRUM_SAMPLES],
+}
+
+impl<T: Float> Clone for Spectrum<T> {
+    fn clone(&self) -> Spectrum<T> {
+        let mut data: [T; SPECTRUM_SAMPLES] = [T::zero(); SPECTRUM_SAMPLES];
+        for (sample, cloned) in self.data.iter().zip(data.iter_mut()) {
+            *cloned = *sample;
+        }
```
acolley: Cool! Didn't know about that method, though it seems there's also [copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) which it suggests might be faster so I've used that.

src/spectrum.rs
```diff
@@ -0,0 +1,375 @@
...
+pub const SPECTRUM_MIN_LAMBDA: f32 = 360.0;
+
+/// The largest wavelength represented in the `Spectrum`
+/// data structure. This is the last value in the
+/// map's wavelength.
+#[allow(dead_code)]
+pub const SPECTRUM_MAX_LAMBDA: f32 = 830.0;
+
+pub const SPECTRUM_SAMPLES: usize = 95;
+const SPECTRUM_SAMPLE_STEP: usize = 5;
```
acolley: I've refactored the constants to use constant expressions and removed the redundant 'SPECTRUM' naming (as this is obvious given we're in the 'spectrum.rs' module).

src/spectrum.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+impl<T> Spectrum<T>
+    where T: Float + Zero + FromPrimitive
+{
+    /// Create a Spectrum from an array of spectral intensity
+    /// values assumed to be in the range 360 nm to 830 nm.
+    pub fn new(data: [T; SPECTRUM_SAMPLES]) -> Result<Self> {
+        if data.iter().any(|&intensity| intensity < T::zero()) {
+            Err(ErrorKind::SpectrumIntensityOutOfRange.into())
```
acolley: I guess if we were going with what the rest of the library does then we should use `assert!` and expect the caller to make sure there are no invalid values in the array.

I've updated the code to do this and that has removed the need for `error-chain` and the `errors.rs` module which I've also removed.

src/spectrum.rs
```diff
@@ -0,0 +1,397 @@
...
+use flt;
+use num::{Float, FromPrimitive, Zero};
+use ordered_float::OrderedFloat;
+use white_point::WhitePoint;
+use xyz::Xyz;
+
+/// The smallest wavelength represented in the `Spectrum`
+/// data structure. This is the first value in the
+/// map's wavelength.
+pub const MIN_LAMBDA: usize = 360;
```
Ogeon: I was thinking that perhaps we should name these `{MIN, MAX}_WAVELENGTH`, since that's what they are, instead of which letter is used to represent them.

src/spectrum.rs
```diff
@@ -0,0 +1,397 @@
...
+
+/// The largest wavelength represented in the `Spectrum`
+/// data structure. This is the last value in the map's
+/// wavelength.
+#[allow(dead_code)]
+pub const MAX_LAMBDA: usize = 830;
+
+const SAMPLE_STEP: usize = 5;
+
+pub const N_SAMPLES: usize = (MAX_LAMBDA - MIN_LAMBDA) / SAMPLE_STEP + 1;
```
Ogeon: It's something with the `N_` prefix... Perhaps just `SAMPLES` is good enough? Or is that ambiguous? `SECTRUM_LENGTH`? `SECTRUM_LEN`, to follow the `.len()` pattern?

src/spectrum.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// Create a `Spectrum` from a slice of (`f32`, `Float`) tuples
+    /// representing a (wavelength, intensity) mapping.
+    /// 
+    /// The more data points there are the more accurate the 
+    /// `Spectrum`'s internal representation will be.
+    ///
+    /// The intensity values must be greater than or equal to
+    /// zero and not NaN.
+    ///
+    pub fn from_samples(data: &[Sample<T>]) -> Spectrum<T> {
```
Ogeon: This could take `&[S]` where `S: Into<Sample<T>>`. And `Sample<T>` could implement `From<(f32, T)>`.

src/spectrum.rs
acolley: I had the exact same thought. Will push this change in the next commit, though I was thinking we might also want to include the units? (e.g. MAX_WAVELENGTH_NM).

src/spectrum.rs
acolley: Sounds good. I'm still learning about all the power Rust gives you for creating convenient APIs.

acolley:
Another design question I had was about whether we should allow the consumer of the API the ability to choose how many samples are stored in a Spectrum. Currently this is hard coded as 95 samples, but comparing this to [pbrt](https://github.com/mmp/pbrt-v2/blob/24e109372feac10630b13728777242c1ae8fb137/src/core/spectrum.h#L46) for example. That only uses 30 samples, presumably because this is a happy medium between accuracy and speed but when compiling that you can just set that number to be as high or low as necessary.

Obviously that wouldn't work for a library like `palette`. However I can't think of any way of allowing that sort of control at creation time short of using a `Vec` to store a dynamic number of samples. Though that would lose the flexibility of a consumer being able to choose if a `Spectrum` should be stack or heap allocated.

It would also affect the performance of the various arithmetic operators as you would have to convert the two `Spectrum`s into the same sample space to do the operation each time.

Perhaps we should consider dropping the number of samples from 95 to some lower number for the initial release of this feature to avoid the overhead of extra accuracy that the majority of users might not need?

In addition, exposing a constructor (i.e. `Spectrum::new`) that can accept a fixed-size array we are exposing the internals and wouldn't be able to adjust the samples in the future without breaking API compatibility. Perhaps we should just have the `from_samples` method that would enable any size of input data to be accepted without exposing how it's stored internally.

Sorry that was a bit of a mind dump :)

Ogeon:
I think it's fine to skip the unit. It's common to use nanometers and it can be clarified in the documentation.

As for the resolution of the spectrum, I think it's a nice idea, but the problems with combining spectra makes it less worth it. It could work if/when Rust gets type level integers. Until then we could use empty types to define different sizes (see how color spaces works), but that's not as flexible.

Do you have a working project that uses this? Would you like to do a bit of research and see if >30 makes a lot of difference? If it is a happy medium, then perhaps that's a good starting point. One thing I'm not sure about is how it works with conversion to and from XYZ. I guess each resolution would need its own curve data.

acolley:
Just an update to say I'm still working on this. Currently working on the finishing touches to a supersampling method for sampling the input data as mentioned in a previous review comment.

Ogeon:
Nice! There are still a couple of undocumented items, but it looks like I haven't added any automatic checks for that yet. Something that would be a nice addition is a small example in the examples directory. Maybe just a simple rainbow, or something. Otherwise I think it's a good first version of a spectrum type.

Ogeon:
@acolley I guess this one was forgotten. Are you still interested?

Ogeon:
I'll close this due to inactivity, but feel free to pick it up again and continue if you ever want to.

PR #76, opened 2018-01-28
Change dependency `num` to `num_traits` to shrink dependency tree

LukasKalbertodt:
This crate only uses traits from the `num` crate. These traits live in
their own crate called `num_traits` which has way fewer dependencies. By
default, `num` pulls all its children crates such as `num-bigint` in. This
in turn brings some heavy crates such as `rand` into the dependency tree.

I tried to use this crate in a wasm project and `rand` currently doesn't compile for the new wasm target. That's the way I found out. 

Changing dependencies like that is probably a breaking change? Or is it? Since `num` in turn had `num_traits` as a dependency? Maybe it's fine...

Btw: awesome crate :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ named = []
 strict = []
 
 [dependencies]
-num = "0.1"
+num-traits = "0.1"
 approx = "0.1"
 
 [dependencies.phf]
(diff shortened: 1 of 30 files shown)


Ogeon:
Wow, thanks! Optimizations are always nice!

It may not be a breaking change, but it's going to be bundled with other breaking changes, so it doesn't matter. By the way, I have been super lousy when it comes to getting this crate off the ground again, so I haven't decided if I'm just going to make an intermediate release now or put the RGB changes in there first. Either way, I don't know when I will release it. Not today, at least.

PR #103, opened 2018-05-26
Use Generics::make_where_clause helper

dtolnay:
The [`Generics::make_where_clause`](https://docs.rs/syn/0.14/syn/struct.Generics.html#method.make_where_clause) method initializes an empty where-clause if there is not one present already, similar to palette_derive's `add_missing_where_clause`. This eliminates a handful of `unwrap` calls that can never fail.


Ogeon:
Ah, nice! Thank you! There are many things that can be improved in those parts. I kind of focused on getting something going to begin with.

bors r+

PR #104, opened 2018-05-27
Update image and approx crate dependency

### LINKED ISSUE #100: Update approx dependency
This requires a bit of rewrite to do it properly, so it was not straight forward enough to do when releasing 0.4.0.

### LINKED ISSUE #101: Uppdate image dependency
Parts of it is deprecated so it wasn't done as part of releasing 0.4.0.

Veykril:
Closes #101 and closes #100. I accidentally put both pull requests together.

### DIFF
--- a/palette/src/equality.rs
+++ b/palette/src/equality.rs
@@ -22,7 +22,7 @@ macro_rules! impl_eq {
                 $( self.$element.abs_diff_eq(&other.$element, epsilon) )&&+
             }
             fn abs_diff_ne(&self, other: &Self, epsilon: T::Epsilon) -> bool {
-                $( self.$element.abs_diff_ne(&other.$element, epsilon) )&&+
+                $( self.$element.abs_diff_ne(&other.$element, epsilon) )||+
             }
         }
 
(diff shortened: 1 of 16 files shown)


palette/src/equality.rs
```diff
@@ -1,38 +1,61 @@
...
-            fn default_max_relative() -> Self::Epsilon {
-                T::default_max_relative()
+
+            fn abs_diff_eq(&self, other: &Self, epsilon: T::Epsilon) -> bool {
+                $( self.$element.abs_diff_eq(&other.$element, epsilon) )&&+
             }
-            fn default_max_ulps() -> u32 {
-                T::default_max_ulps()
+            fn abs_diff_ne(&self, other: &Self, epsilon: T::Epsilon) -> bool {
+                $( self.$element.abs_diff_ne(&other.$element, epsilon) )&&+
```
Ogeon: I think these should be delimited with `||`, since it would be the inverse of `abs_diff_eq`. Same for the other `*_ne` implementations.

Veykril: You're right that should be `||`, looks it was implemented wrong previously as well because the `*_ne` were using `&&`. I just adapted that by accident

Ogeon: Oops, good thing it was discovered then.

Ogeon:
I think that was all. Great!

bors r+

PR #106, opened 2018-06-03
Add Extended Conversion Trait 

Veykril:
This commit tries to implement a conversion trait similar to the one mentioned in #41 mimicking the `From`/`Into` traits of the `std` library.

`convert_*` clamps the resulting color to its color space limitations.
`convert_unclamped_*` simply converts the color.
`try_convert_*` converts the color and returns it in a `Result` depending on it being valid or not.

### DIFF
--- a/palette/src/convert.rs
+++ b/palette/src/convert.rs
@@ -520,7 +520,11 @@ impl<T> OutOfBounds<T> {
     }
 }
 
-impl<T: Debug> error::Error for OutOfBounds<T> {}
+impl<T: Debug> error::Error for OutOfBounds<T> {
+    fn description(&self) -> &str {
+        "Color conversion is out of bounds"
+    }
+}
 
 impl<T> Display for OutOfBounds<T> {
     fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
--- a/palette/src/convert.rs
+++ b/palette/src/convert.rs
@@ -1,12 +1,11 @@
 use num_traits::Float;
 
-use std::error;
+use std::error::Error;
 use std::fmt::{self, Debug, Display, Formatter};
 use {Component, Limited, Hsl, Hsv, Hwb, Lab, Lch, Xyz, Yxy};
 use white_point::{D65, WhitePoint};
-use rgb::{Rgb, RgbSpace, RgbStandard};
-use luma::{Luma, LumaStandard};
-use Alpha;
+use rgb::{Rgb, RgbSpace};
+use luma::Luma;
 use encoding::Linear;
 
 /// FromColor provides conversion from the colors.
@@ -510,17 +509,19 @@ pub struct OutOfBounds<T> {
 
 impl<T> OutOfBounds<T> {
     ///Create a new error wrapping a color
+    #[inline]
     fn new(color: T) -> Self {
         OutOfBounds { color }
     }
 
     ///Consume this error and return the wrapped color
+    #[inline]
     pub fn color(self) -> T {
         self.color
     }
 }
 
-impl<T: Debug> error::Error for OutOfBounds<T> {
+impl<T: Debug> Error for OutOfBounds<T> {
     fn description(&self) -> &str {
         "Color conversion is out of bounds"
     }
@@ -532,52 +533,35 @@ impl<T> Display for OutOfBounds<T> {
     }
 }
 
-///An extension for `Result` that offers a method to take out the color of `try_convert` result.
-pub trait ResultExt<T> {
-    ///Take the color out of this `Result`.
-    fn take(self) -> T;
-}
-
-impl<T> ResultExt<T> for Result<T, OutOfBounds<T>> {
-    #[inline]
-    fn take(self) -> T {
-        match self {
-            Ok(color) => color,
-            Err(color) => color.color(),
-        }
-    }
-}
-
-///A trait for converting from one color to another.
+///A trait for converting a color into another.
 ///
 ///`convert_unclamped` currently wraps the underlying `Into` implementation.
-pub trait Convert<T>: Sized + Into<T> {
-
+pub trait ConvertInto<T>: Into<T> {
     ///Convert into T with values clamped to the color defined bounds
     ///
     ///```
-    ///use palette::Convert;
+    ///use palette::ConvertInto;
     ///use palette::Limited;
     ///use palette::{Srgb, Lch};
     ///
     ///
-    ///let rgb: Srgb = Lch::new(50.0, 100.0, -175.0).convert();
+    ///let rgb: Srgb = Lch::new(50.0, 100.0, -175.0).convert_into();
     ///assert!(rgb.is_valid());
     ///```
-    fn convert(self) -> T;
+    fn convert_into(self) -> T;
 
-    ///Convert into T, the resulting color might be invalid in it's color space
+    ///Convert into T. The resulting color might be invalid in its color space
     ///
     ///```
-    ///use palette::Convert;
+    ///use palette::ConvertInto;
     ///use palette::Limited;
     ///use palette::{Srgb, Lch};
     ///
-    ///let rgb: Srgb = Lch::new(50.0, 100.0, -175.0).convert_unclamped();
+    ///let rgb: Srgb = Lch::new(50.0, 100.0, -175.0).convert_unclamped_into();
     ///assert!(!rgb.is_valid());
     ///```
     #[inline]
-    fn convert_unclamped(self) -> T {
+    fn convert_unclamped_into(self) -> T {
         self.into()
     }
 
@@ -585,10 +569,10 @@ pub trait Convert<T>: Sized + Into<T> {
     ///otherwise an `OutOfBounds` error is returned which contains the unclamped color.
     ///
     ///```
-    ///use palette::Convert;
+    ///use palette::ConvertInto;
     ///use palette::{Srgb, Hsl};
     ///
-    ///let rgb: Srgb = match Hsl::new(150.0, 1.0, 1.1).try_convert() {
+    ///let rgb: Srgb = match Hsl::new(150.0, 1.0, 1.1).try_convert_into() {
     ///    Ok(color) => color,
     ///    Err(err) => {
     ///        println!("Color is out of bounds");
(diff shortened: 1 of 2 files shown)


palette/src/convert.rs
```diff
@@ -499,6 +501,60 @@ where
...
+impl<C, T> Convert for Alpha<C, T> where C: Convert, T: Component {}
+
+///A trait for converting from one color to an arbitrary type by handing a closure the color's raw components.
+///
+///This trait makes use of the `Pixel` trait which can be unsafe depending on the implementator.
+pub trait ConvertWith<T>: Pixel<T> where T: Component
+{
+    ///Convert `self` to `T` by applying a function over its components
+    fn convert_with<F, Out>(self, f: F) -> Out where F: FnOnce(&[T]) -> Out, Out: Sized {
+        f(Pixel::into_raw_slice(&[self]))
```
Ogeon: This could be `as_raw` and the function could take `&self`. Could also use `into_raw`, but that becomes more complicated.

Veykril:
Yes, making from/the default conversion clamp to the output's valid range is probably the better approach. I believe the more common use case is only working with valid colors instead. Working with `out of bounds` colors should probably be done explicitly. Panicking is out of the question for the default behavior I'd say, especially since the standard library also says that the `From` trait should never fail. So I guess I am leaning towards default clamping as well.

That leaves the question if we even need a extra unclamped conversion, since as you mentioned we could also use the Err return value from the TryFrom implementation. One downside I see with this though is that it will be somewhat cumbersome to extract the values when you do not care about valid colors, since unwrap wont just work like that for you then. So we would probably need to implement a function + trait for Result<T, OurError<T>> that will unwrap to T in oth Ok and Err cases.

palette/src/convert.rs
Veykril: Ah yes, I had problems with coming up for a name for this function actually. `as_raw` should work here I suppose, and I didn't take self by reference because I thought of it being a conversion into something else so consuming self would've fit there. It's probably better to take it be ref nevertheless. Any Ideas on the trait name for that?

Ogeon: I have no better name suggestion at the moment.

Ogeon:
I agree about the panicking. It would just result in unexpected crashes instead of unexpected values.

Changing the default behavior to clamping will require a bit more work on top of this trait. `FromColor` and `IntoColor` will need a new design, since they are what people will reach for when `From` and `Into` isn't enough. It may be best to rename them and add clamping conversion traits with the current names. An alternative is to add more methods. The derives will also need to change a bit. I'm trying to come up with a way to make it ergonomic.

Ogeon:
Since we can't rely on `TryFrom` being a thing in the near future, I'm thinking that we should go with these traits, but maybe with a few modifications. My suggestion is to make it a first step towards making clamping the default, so let's swap the semantics of `convert` and `convert_clamped` to be `convert` (clamped) and `convert_unclamped` (better name is welcome). You will probably have to go without a default implementation for now, to avoid requiring `Limited`.

I would also appreciate some examples in the documentation. That's always nice.

The rest will need some more time and design to get the relationships between the traits right. I don't want to put all of that into this PR, to avoid turning it into some kind of monster task. It's better to split it and get this in while it's fresh.

Veykril:
Okay, I'll get to this in the next few days. So should I include the ` try_convert` function and the `as_raw` function as well? Also the ` convert` function would require `Limited` on the implementing type since it clamps according to the trait implementation. Unless you meant something different with that sentence.

Ogeon:
> Nevermind you meant as in implementing the convert trait individually for every color right?

Yes, exactly, otherwise we can't convert to every color. Only the ones that makes sense to clamp. And yes, `try_convert` should stay.

I have been thinking a bit about `convert_with` (I guess that's what you mean with `as_raw`) and I'm starting to wonder if has much of a use case anymore, or if that issue is a bit obsolete now when it's so easy to convert to `[T]` or `[T; N]`. You just have to use `AsRef` to convert it before calling the convert function, so `convert_to_something(my_color.as_ref())` instead of `my_color.convert_with(convert_to_something)`.

Veykril:
I just tried to make Convert not require `Limited` anywhere, but if I do not make `Limited` a bound on the output type then I wont be able to do any clamping ofc. Also I just noticed but doesn't every color implement `Limited`? So wouldn't that trait bound be no problem?

And I agree the raw conversion method is probably not really needed.

Ogeon:
Oh, you are right. Then I guess it would either have to be a `ConvertFrom`/`ConvertInto` pair, where `ConvertFrom` controls the implementation and `ConvertInto` just refers to it, or `Limited` have to be assumed. I was thinking on a pretty abstract level when writing that it can't be assumed, because then it doesn't have to be a hard requirement, but maybe it's a good requirement. It makes sense semantically, but a theoretically unlimited system doesn't have to be clamped. Maybe it shouldn't be called `Limited` at all. But... I guess most, if not all, color spaces have some kind of range or limit. Might as well require it until it becomes a problem.

Veykril:
I actually did manage to remove the `Limited` trait bounds by implementing the trait in a way that the output color is defined explicitly. With that we have now access to the `Limited` functions should the output color have access too. 

I also added some examples in the docs and changed the return type of `try_convert` into a `Result` with a custom error.

I'm not so sure if I like the `take` function that I added for `Result<T, OutOfBounds<T>>` though, not too sure why I thought of this to be honest.

palette/src/convert.rs
```diff
@@ -499,6 +502,173 @@ where
...
+
+impl<T: Debug> error::Error for OutOfBounds<T> {
+    fn description(&self) -> &str {
+        "Color conversion is out of bounds"
+    }
+}
+
+impl<T> Display for OutOfBounds<T> {
+    fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
+        write!(fmt, "Color conversion is out of bounds")
```
Ogeon: You could reuse the description here: `self.description().fmt(fmt)`.

palette/src/convert.rs
```diff
@@ -499,6 +502,173 @@ where
...
+    ///use palette::Limited;
+    ///use palette::{Srgb, Lch};
+    ///
+    ///
+    ///let rgb: Srgb = Lch::new(50.0, 100.0, -175.0).convert();
+    ///assert!(rgb.is_valid());
+    ///```
+    fn convert(self) -> T;
+
+    ///Convert into T, the resulting color might be invalid in it's color space
```
Ogeon: *its :slightly_smiling_face:

palette/src/convert.rs
```diff
@@ -499,6 +502,173 @@ where
...
+}
+
+impl<T> Display for OutOfBounds<T> {
+    fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
+        write!(fmt, "Color conversion is out of bounds")
+    }
+}
+
+///An extension for `Result` that offers a method to take out the color of `try_convert` result.
+pub trait ResultExt<T> {
```
Ogeon: I think we have the same gut feeling regarding this. I have a feeling that it will not be used a lot, since it has to be explicitly imported. Besides, it's a bit redundant with `convert_unclamped` basically doing the same thing. I vote for skipping it and maybe adding it later if there's ever a need. I'm in a phase where I want to remove things and boil this library down as much as I can. :smile:

palette/src/convert.rs
```diff
@@ -499,6 +502,173 @@ where
...
+    ///    Err(err) => {
+    ///        println!("Color is out of bounds");
+    ///        err.color()
+    ///    },
+    ///};
+    ///```
+    fn try_convert(self) -> Result<T, OutOfBounds<T>>;
+}
+
+macro_rules! expand_impl_convert {
```
Ogeon: I would honestly like to move away from macros if I can. They have been a hassle and I have tried to replace some of them with derives and just regular `impl`s. Do you think all of this would be possible to replace with:

```rust
impl<T: Into<U> + Limited, U> Convert<U> for T {
    // ...
}
```

It has the same pattern as the `Into` blanket implementation, so I think it should work. It's not a big deal if it's no perfect yet, and I'm sure it will have to change anyway when the rest of the library catches up and the implementation of `Into` changes. I just want a start that is somewhat in line with my longer term goals (which I maybe should note down in an issue instead of keeping in my head).

palette/src/convert.rs
Veykril: I didn't implement description at all at first because it is `soft deprecated` as its docs say, hence I didnt use it there either, but I might as well. No need to duplicate the literal there.

palette/src/convert.rs
Veykril: Yes thought the same, why use this when you could just use `convert_unclamped` in the first place.

palette/src/convert.rs
Veykril: Understandable, Macro implementations make code quite difficult to read. I was trying to get rid of the `Limited` trait-bound but that just wasnt possible with a generic `impl`. I can revert back to `impl`ementing the trait. While we are talking about the similarities to `Into`, wouldn't it make more sense to make this two traits which are `ConvertFrom` and `ConvertInto` then?

palette/src/convert.rs
```diff
@@ -499,6 +501,182 @@ where
...
+    ///        err.color()
+    ///    },
+    ///};
+    ///```
+    fn try_convert_into(self) -> Result<T, OutOfBounds<T>>;
+}
+
+///A trait for converting one color from another.
+///
+///`convert_unclamped` currently wraps the underlying `Into` implementation.
```
Ogeon: I guess this was meant to say "From".

palette/src/convert.rs
```diff
@@ -499,6 +501,182 @@ where
...
+        U::convert_from(self)
+    }
+
+    #[inline]
+    fn try_convert_into(self) -> Result<U, OutOfBounds<U>> {
+        U::try_convert_from(self)
+    }
+}
+
+/*
```
Ogeon: Also remove these. :slightly_smiling_face:

Ogeon:
Very nice! I think this will be the way to go. I'm just thinking that it may be wise to change the relationship to `ConvertInto` wrapping `ConvertFrom` to align it with `From` and `Into`. Just for consistency. After that and the small fixes, I think it's done from my point of view. :+1:

Veykril:
Alright, made `ConvertInto` wrap `ConvertFrom` in the generic implementation since it wouldnt be possible to do it on the trait level due to `ConvertInto` not requiring `ConvertFrom`.

Ogeon:
That's exactly what I expected. Looks good!

The PR title and description are going to be included in the merge commit message, so feel free to edit it a bit (at least remove "[wip]") and let me know when/if you feel like it's done.

Veykril:
I'd say it's good to go.

Ogeon:
I think you typoed the issue number. #4 is not that relevant :thinking:

Veykril:
That was supposed to be a 41, my bad. 😅

Ogeon:
Alright!

bors: r+

PR #110, opened 2018-07-09
No std support

### LINKED ISSUE #108: Possibility of supporting `no_std`?
I was wondering if it would be possible for this library to support a `#![no_std]` environment.  Because it's mostly conversion logic, I think a lot of it could be useful in embedded situations when it's necessary to e.g. write colors to a framebuffer, or control an RGB LED, or something like that.

My specific use case is in a project I'm working on with bare-metal raspberry pi development, where you can get a 24-bit color depth framebuffer from the GPU.  I would love to just use `Pixel::from_raw_slice_mut` as shown [in the docs](https://docs.rs/palette/0.4.0/palette/#1-decoding), but I was forced to roll my own `struct Pixel(u8, u8, u8)` instead.

If it's simple enough, I'd be willing to try to implement this myself with some help.  (I don't know enough about colors or about the Pixel-deriving mechanism to do anything very complicated, though.)

Thanks!

mb64:
Adds support for using palette without the standard library, as discussed in issue #108.

Fixes #108.

One tiny question: because this would be in a future release (not 0.4.0), what version should be in the README example Cargo.toml for `#![no_std]`?  I put 0.4, but was unsure about if it should be something different.

### DIFF
--- a/palette/tests/convert/lab_lch.rs
+++ b/palette/tests/convert/lab_lch.rs
@@ -7,8 +7,8 @@ fn lab_lch_green() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
+    assert_relative_eq!(lch, expect_lch, epsilon = 0.1);
 }
 
 #[test]
@@ -19,8 +19,8 @@ fn lab_lch_magenta() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
+    assert_relative_eq!(lch, expect_lch, epsilon = 0.1);
 }
 
 #[test]
@@ -31,6 +31,6 @@ fn lab_lch_blue() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
+    assert_relative_eq!(lch, expect_lch, epsilon = 0.1);
 }
--- a/palette/tests/convert/lab_lch.rs
+++ b/palette/tests/convert/lab_lch.rs
@@ -7,8 +7,8 @@ fn lab_lch_green() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.1);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
+    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
 }
 
 #[test]
@@ -19,8 +19,8 @@ fn lab_lch_magenta() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.1);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
+    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
 }
 
 #[test]
@@ -31,6 +31,6 @@ fn lab_lch_blue() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.1);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
+    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
 }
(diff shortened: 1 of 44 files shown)


Ogeon:
Excellent! I will look closer at it tomorrow (I think), but until then, leave the version as 0.4. I have a script for incrementing it automatically throughout the project when I'm preparing a release, so I will make it find that one as well.

Ogeon:
Oh, and those test results are way off! I didn't think the soft floats would differ that much...

mb64:
I ran some tests, and `mish::powf(x, 1./3.)` returns `NaN` for `x < 0.0852`.  However, the `cbrt_f32` function seems OK.  I'll push an update in a few minutes that changes it to use `.cbrt()`.

Ogeon:
Didn't know `qbrt` existed, but that sounds like a better choice in any case. Looks like you have made some really good progress with making the tests pass, and the code looks good as well! I'm just wondering if the float functions follow any specification, seeing that they are not just plain wrappers.

Ogeon:
Maybe I should have checked your last commit... Then I would have known where most some of it came from. :stuck_out_tongue:

Ogeon:
Hi! I'm just checking in to see how it's going. If you are just taking a break or on vacation, or similar, I'll leave you to enjoy that. Otherwise let me know if you are getting stuck somewhere or if anything else came in the way.

Also, I wanted to check if you have followed the recent development over at https://github.com/japaric/libm. It has gotten a ton of help during the last few weeks, so it's almost feature complete. Would be nice to only have to depend on that one library.

mb64:
Sorry I haven't been working on this for a couple weeks -- I was of town for a while, and then just didn't get back to it.  There are a few things left in this:

* To be able to say that it depends on `std`, Cargo wants a feature name for `serde` that's not the same name as the library.  Some options could be `serialize`, `serializable`, `serialize_deserialize`, `json_formatting`, etc. (Personally, I vote for `serializable`.)
* Now that there's an actual community effort for math on `no_std`, I think that `libm` will eventually include any x86/arm hard float implementations, making the potential need for user-provided `extern "C"` implementations an even more niche use-case.  I think it would probably be better (less surprising to users) to just use `libm` all the time.
* Unrelated to `no_std`, recent nightly's have ["fixed" a thing](https://github.com/rust-lang/rust/issues/50504) with proc macros that `palette_derive` was using, so now there's lots of warnings (or errors with `-Dwarnings`) about macro hygiene, etc.

Ogeon:
Alright, no hurry. :slightly_smiling_face:

 * So `serde` depends on `std` after all, or am I misunderstanding? In that case I would probably pick `serializing` as feature name.
* Sounds good! Better to reduce the complexity now and add it back later if it turns out to be required.
* Yep, see #111. I would ideally want to fix it in a 0.4.1 release, but haven't gotten around to do it yet, as I have also been quite occupied elsewhere until about now.

palette/tests/convert/lab_lch.rs
```diff
@@ -7,8 +7,8 @@ fn lab_lch_green() {
     let expect_lab = lch.into_lab();
     let expect_lch = lab.into_lch();
 
-    assert_relative_eq!(lab, expect_lab, epsilon = 0.001);
-    assert_relative_eq!(lch, expect_lch, epsilon = 0.001);
+    assert_relative_eq!(lab, expect_lab, epsilon = 0.1);
```
Ogeon: Have you tried to tighten these error margins again?

mb64: I have -- it works fine :smiley: thanks to `libm`.
I'll push those changes in a moment.

mb64:
I expect this CI run to fail, because of japaric/libm#4.  I could only get it to work by running it with `RUSTFLAGS=-Coverflow-checks=off`.  There are a few ways to deal with this:

 * Set `RUSTFLAGS` in the CI: this is unideal because it disables other overflow checks that you might want the CI to catch
 * Make lots of [PR's](https://github.com/japaric/libm/pull/142) to `libm` and wait for them to be released: this will take a while and potentially create merge conflicts with other changes, etc.
 * Both?

palette/tests/convert/lab_lch.rs
Ogeon: Great! :smile:

Ogeon:
I appreciate that you are already on the case with PRs! Hopefully we can get away with just a few of those, because I would really like to wait for it to be fixed in `libm`. To me, the alternative feels like just hiding the problem (and other potential problems, as you mention). Waiting is not a problem for me, but the conflicts could be annoying if the branches diverges too much. Maybe the whole thing could be split into two phases. The first one prepares for `no_std` by removing the current usages of `std`. The second one waits for the current issues in `libm` to be fixed and implements the final changes. I'm happy with the general implementation, so building it in debug mode is pretty much all that's left from my point of view.

What's your view on it? You are putting in a lot of work to get this through (which I once again appreciate a lot) and I don't want it to turn into a burden for you, rather than something enjoyable.

mb64:
The only necessary PR's to libm are for atan2 and sqrt; the others work fine.  Here's an idea: 
 * Add a commit to this PR that disables "really" using no_std based on a not-in-Cargo.toml feature (something like `"libm_works"`).  This will make it at least mergeable.
 * Make another PR to Palette that undoes the temporary feature; wait to merge it until `libm` is OK.

Ogeon:
Sounds like a plan!

Ogeon:
By the way, I have a PR up with a quick fix for the hygiene warning and will make a release within the next day or so. Then I'll also merge those changes into `master`. :slightly_smiling_face:

Ogeon:
The fix is in `master` now.

mb64:
I just pushed a merge from `master` -- it should pass CI now.  The `no_std` float trait that uses `libm` is hidden behind a never-enabled feature flag.

I think it's OK to merge, waiting on reverting the float trait once the libm PR's are in and released.

mb64:
(As a side note -- the docs link in `palette/README.md` points to the `0.4.0` docs, not the `0.4.1` docs)

Ogeon:
Thanks! I saw that one of the travis jobs timed out, so I restarted it, but I'll merge it when it's all green. If you have your revert ready, you can upload it and open a PR that can stay open until it's time. Otherwise I'll make an issue for it so it's not forgotten.

> (As a side note -- the docs link in palette/README.md points to the 0.4.0 docs, not the 0.4.1 docs)

Oh, you're right. It's supposed to be symlinked to the root README.md, but I must have messed something up...

Ogeon:
bors: r+

PR #117, opened 2018-09-10
Round to nearest instead of down when converting components to integers.

PeterHatch:
A pretty simple change, this makes it so when converting components to an integer format, we round it to nearest first, instead of just doing the conversion which will round towards zero.

Not happy about using the destination type's `LIMITED` to determine if it will be an integer, since that's not really appropriate semantics, but it is currently set if and only if the type is an integer, and every integer type would need it, and I'm not sure why there'd every be a limited non-integer type.  Could rename it, or add a separate property, but I'm not sure it's important enough for that.


Ogeon:
Thanks, I appreciate it a lot! I have been meaning to adjust this whenever I pass by it, but I have had other things on my mind for a while... Anyway, rounding to nearest will make it slightly off in a different way. 0 and `MAX` will be smaller than the rest of the "bins", for example. A more correct formula would be `(x * (MAX + 1)).min(MAX)`. The only time it would hit `MAX + 1` is if it's larger than or equal to 1 and that bias will be negligible compared to floating point errors. Here are some more detailed explanations: https://stackoverflow.com/questions/1914115/converting-color-value-from-float-0-1-to-byte-0-255

As for the `LIMITED` constant, nothing prevents a third party from implementing the trait for a non-integer value. The formula I suggested clashed with the semantics as well, so I'm not sure what to do about it right now. I'm open for suggestions!

PeterHatch:
I actually agree with the second answer at the link you posted, that the way I did it is more accurate. :)

Fundamentally, I think reducing error is the goal, not having equal sized bins.  And I think converting from an int to a float is accurately done by simply dividing by `MAX`.  Not having equal sized bins when converting to a fixed range is a result of the desire for the fixed range to be able to exactly represent the end points - if not for that, the proper formula for converting from int to float would be, I think, `(x * 2 + 1)/((MAX +1) * 2)`, and then the formula that you gave would both have equal sized bins and minimize error.  Does that make sense?

Um, still thinking about the `LIMITED` thing.  How important is backwards compatibility?  Is it okay if I find a solution that removes that from the public API?

Ogeon:
It's fine to break things if you have a better setup in mind. :+1:

> I actually agree with the second answer at the link you posted, that the way I did it is more accurate.

Do you mean the `b=f*255.999` one? I guess it accomplishes more or less the same thing. I don't trust the input enough to skip the clamping, so multiplying with 256 or 255.999 would practically be the same thing. I should have written

```rust
let scaled = x * (cast(T::max_intensity()) + 1.0);
cast(clamp(scaled, 0.0, cast(T::max_intensity())))
```

as that's what it would end up being in our case.

The way the accepted answer from the link, and this formula, works is the same as `b=f*255.999`, but it leverages the fact that only 1.0 and above will be out of rage. That way it gets as close to the full range as possible while only being biased towards 255 by practically one value. Multiplying by 255.999 achieves the goal by trying to avoid getting too close to the upper limit, but by an arbitrary value, so the scale is still off (even if it's just by a fraction). Using 255.999 would probably be fine if that was all, but since we are still clamping we might as well go all the way. We are not really saving much computing power. Only one addition, as far as I can tell.

> - if not for that, the proper formula for converting from int to float would be, I think, (x * 2 + 1)/((MAX +1) * 2), and then the formula that you gave would both have equal sized bins and minimize error.

I think you will have to explain that one... :thinking:

PeterHatch:
Um, the answer I meant is the one that advocates for `round(f * 255.0)` - https://stackoverflow.com/questions/1914115/converting-color-value-from-float-0-1-to-byte-0-255/46575472#46575472

Okay, to explain more - say you're converting to and from just 2 bits, as that's easier to explain.  There's a couple methods you could use.  Method A sets the four available values to 1/8, 3/8, 5/8, and 7/8 - this means that when converting, you're never off by more than 1/8, but you can't exactly represent something at 0 or 1 - full white or black would be impossible to represent if this method was used for colors.  Method B has the available values at 0, 1/3, 2/3, and 1, which means that conversion could be off by 1/6, but you can exactly represent 0 or 1.

Method A means that conversion will have the same size bins for each value; that is how it minimizes the maximum possible error.  Method B does not have that property - the bins will be half size for the minimum and maximum values.  Nonetheless, colors stored as discrete values should use method B, so as to be able to properly represent full white and black.

Does that make sense now?

Ogeon:
I think I get it now. Especially with the answer you really meant in mind. Reading that and the explanation again, I'm starting to think that you may be right. My main requirements are that

* Values close to 0.0 should map to 0u8, and 0u8 should map to 0.0.
* Values close to 1.0 should map to 255u8, and 255u8 should map to 1.0.

which both methods fulfill. Then I assumed rounding down was preferred because of familiarity, but I'm starting to wonder if it's important and if it's more important to avoid conversion errors. I want Quality Colors™ so maybe a low error rate is the way to go after all. As long as it doesn't create too "weird" colors compared to other software, because that may actually be worse than conversion errors. If you don't mind, you can try to add a test in palette/tests/convert for converting to u8. There doesn't seem to be one right now, but that would at least check against what's probably generally expected.

Ogeon:
By the way, do you think reversing the relationship, so it's `fn convert_from<T: Component>(other: T) -> Self` would help with the semantics? Then only the target type need to care about rounding. I'm not sure why I didn't do that from the beginning.

Ogeon:
Hi! Are you still interested in continuing this PR?

PeterHatch:
Yes!  I actually did work on it back in September, just need to get back to it.  Sorry about dropping it for so long and not communicating.

Ogeon:
Don't worry about it! I have also had my mind elsewhere for a while. It's
nice to see that you are still on it. No stress, and let me know if you run
into any roadblocks.

Ogeon:
Nice! It becomes a bit of a combinatoric monster, but I think it's a reasonable case for using macros to remove repetition. I'm super tired right now, but I will have another look tomorrow. Until then I have at least a couple of comments:

> Oh, also I removed conversion from f32 to u32, as I don't think that makes sense, and removed u64 as a component type entirely. In general I think the point of using an integer type is to reduce the storage size; my though was if you're trying to convert from f32 to u32, you're probably making a mistake. But if I'm missing something, let me know.

The larger integer types are there to support formats with more range/precision. 16 bit components are not that unusual in photography, but 32 bit and even 64 bit may appear. Either way, I see no big downside with supporting them, other than the repetition (but we have macros).

> Also want to update this to be based off the latest master branch, and maybe combine the two commits into one, but I haven't done that often enough to remember how, and didn't want to mess with it right now. But I'm aware it's something to be done; I may wait to see if there's other feedback to act on first, though.

You might as well wait with combining them until it's all green, but I would recommend rebasing your branch onto `master` to bring in the latest and greatest. Let me know if you need any help and I can either point you towards some resources or give you some useful commands. :slightly_smiling_face:

palette/src/lib.rs
```diff
@@ -825,134 +825,430 @@ pub trait ComponentWise {
...
     /// The highest displayable value this component type can reach. Higher
     /// values are allowed, but they may be lowered to this before
     /// converting to another format.
     fn max_intensity() -> Self;
+}
+
+/// Trait for converting between color components; component-specific version
+/// of std::convert::From.
+pub trait FromComponent<T: Component>: Component {
+    fn from_component(T) -> Self;
```
Ogeon: This needs a documentation comment.

Ogeon:
Hi! Long time, no follow-up from me (sorry). I just want to check what the status for this is, as it hasn't moved in a while, and maybe suggest rewinding it to only include the first commit. I would definitely appreciate to have the first one merged. The rest can be split out to a separate project, if it is to be continued, as it involves a bit more figuring out.

Ogeon:
I have created a separate PR with only the first commit: #135. I'm going to close this one as inactive when that one passes the tests. Feel free to open a new PR if you decide to continue the trait project!

PR #119, opened 2018-10-02
Remove the color enum

### LINKED ISSUE #72: Remove the Color type
It's unwieldy to implement, hides things unnecessarily, and I suspect that it's not as useful as I initially thought it would be. It feels like it does more harm than good.

Veykril:
This PR removes the color type enum as described in #72. I do not know if this has any side effects tough, everything compiles fine on my end, tests and examples compile and run as well. The only difference i noticed is in the `readme_examples/readme_manipulation` example part. The latter two colors arent the same but im not sure why.

Closes #72.


palette/examples/color_scheme.rs
```diff
@@ -93,8 +93,8 @@ fn main() {
         .and_then(|r| r.parse().ok())
         .expect("the blue channel must be a number in the range [0-255]");
 
-    let primary: Color = Srgb::new(red, green, blue)
-        .into_format()
+    let primary: Hsv = Srgb::new(red, green, blue)
```
Ogeon: This should be `Lch` to achieve the same results.

palette/examples/readme_examples.rs
```diff
@@ -27,13 +27,13 @@ mod color_spaces {
...
-    use palette::{Color, Saturate, Shade, Srgb};
+    use palette::{Saturate, Shade, Srgb, Hsv};
     use display_colors;
 
     pub fn run() {
-        let color: Color = Srgb::new(0.8, 0.2, 0.1).into_linear().into();
+        let color = Srgb::new(0.8, 0.2, 0.1).into_linear();
         let lighter = color.lighten(0.1);
-        let desaturated = color.desaturate(0.5);
+        let desaturated = Hsv::from(color).desaturate(0.5);
```
Ogeon: This should also be `Lch` to achieve the same results.

palette/examples/color_scheme.rs
Veykril: Ah yes, I already thought that it had to be the color type I used

Veykril:
Okay, the example now has the correct color and I've updated the Readme 👌

palette/examples/color_scheme.rs
Ogeon: This kind of illustrates one of the reasons why I wanted to remove it. The selling point for `Color` was that it would pick the most suitable color space for the task. The downside was that it would hide this choice and could produce surprising results.

Ogeon:
Nice! This is the best kind of pull request, where added/removed counter goes negative! My final request is that you mention "fixes #72" or "closes #72" in your PR description. That will go into the merge commit and automatically close the issue. :slightly_smiling_face:

Ogeon:
Thanks!

bors: r+

PR #120, opened 2018-10-07
Fix grammar in readme

soenkehahn:
> The problem is that these formats __don't__ represent the actual intensities [...]


Ogeon:
Hah, thank you! Englishing is not always easy. :smile:

soenkehahn:
Thanks for merging!

PR #127, opened 2019-03-20
fix no_std

### LINKED ISSUE #125: no_std support is broken
I tried using this crate in a no_std context and it doesn't work because the inclusion of `num_traits::Float` isn't behind a feature flag, and evidently, the stuff behind the `libm_works` feature flag doesn't "work".

evq:
fixes #125

### DIFF
--- a/palette/src/convert.rs
+++ b/palette/src/convert.rs
@@ -391,7 +391,7 @@ where
 ///         0.7353569830524495,
 ///         0.5370987304831942,
 ///     ];
-///     let hsv = Bgr::from_raw_slice(&buffer)[1].into();
+///     let hsv: Hsv<palette::encoding::Srgb, _> = Bgr::from_raw_slice(&buffer)[1].into();
 ///
 ///     assert_relative_eq!(hsv, Hsv::new(90.0, 1.0, 0.5));
 /// }
@@ -441,7 +441,7 @@ where
 ///         blue: 255,
 ///         alpha: 0.3,
 ///     };
-///     let color = css_color.into();
+///     let color: LinSrgba = css_color.into();
 ///
 ///     assert_relative_eq!(color, LinSrgba::new(0.496933, 0.0, 1.0, 0.3));
 /// }
--- a/palette/src/convert.rs
+++ b/palette/src/convert.rs
@@ -352,9 +352,11 @@ where
 ///
 /// use palette::{Component, Hsv, IntoColor, Pixel, Srgb};
 /// use palette::rgb::{Rgb, RgbSpace};
-/// use palette::encoding::Linear;
+/// use palette::encoding::{Linear, self};
 /// use palette::white_point::D65;
 /// use palette::float::Float;
+///
+/// type Hsv64 = Hsv<encoding::Srgb, f64>;
 ///
 /// /// sRGB, but with a reversed memory layout.
 /// #[derive(Copy, Clone, IntoColor, Pixel)]
@@ -391,7 +393,7 @@ where
 ///         0.7353569830524495,
 ///         0.5370987304831942,
 ///     ];
-///     let hsv: Hsv<palette::encoding::Srgb, _> = Bgr::from_raw_slice(&buffer)[1].into();
+///     let hsv: Hsv64 = Bgr::from_raw_slice(&buffer)[1].into();
 ///
 ///     assert_relative_eq!(hsv, Hsv::new(90.0, 1.0, 0.5));
 /// }
--- a/palette/src/blend/blend.rs
+++ b/palette/src/blend/blend.rs
@@ -1,6 +1,6 @@
 use num_traits::{One, Zero};
 use float::Float;
-#[cfg(feature = "libm_works")]
+#[cfg(not(feature = "std"))]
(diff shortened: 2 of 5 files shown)


Ogeon:
Cool, thank you for this! Is libm in a state where it can be used for the things we need now? That would be excellent! If this is all that's needed, it's definitely approved once the tests pass.

As for the test results, that overflow seems strange. I hope it's not something too nasty... Otherwise, I guess they had to tweak some inference rules since last time it ran.

evq:
hey @Ogeon - looks like the overflow is due to the longstanding issue https://github.com/rust-lang-nursery/libm/issues/4 which was previously discussed in https://github.com/Ogeon/palette/pull/110. so to your question - I suppose libm is not yet fixed.

that libm issue is very interesting - it appears the overflow is intentional however it does not use `Wrapping` or `wrapped_add` and as such panics only in debug mode. meaning release mode is perfectly functional.

IMO having working `no_std` support with the caveat that there may be occasional spurious failures in debug mode far outweighs not having `no_std` support at all (current state). What are your thoughts on updating the documentation to highlight this caveat and move forward without waiting on libm?

Ogeon:
Ah, that one. I more or less agree with you, but it requires a workaround that will also work for the tests. I'm not a fan of the thought of disabling overflow checks for the whole library, for example, but maybe that's what we have to do... We can't wait for libm to move if we want this within a foreseeable future, so unless a better alternative has appeared since last time i checked, we have to do what we can with what we have.

palette/src/convert.rs
```diff
@@ -391,7 +391,7 @@ where
 ///         0.7353569830524495,
 ///         0.5370987304831942,
 ///     ];
-///     let hsv = Bgr::from_raw_slice(&buffer)[1].into();
+///     let hsv: Hsv<palette::encoding::Srgb, _> = Bgr::from_raw_slice(&buffer)[1].into();
```
Ogeon: Wasn't just `Hsv` enough? `Srgb` should be the default.

evq: yeah strangely enough, just `Hsv` didn't seem to be enough :(

Ogeon: That's odd. Oh well. In that case can you add `Srgb` to the import at the top of the example? That would make this part a bit less noisy... I don't know if you tried it, but you may be able to skip the `_` at least, unless I'm forgetting something.

evq: That would conflict with `palette::Srgb` imported above - do you have a preferred alias for the `Srgb` encoding? (e.g. `eSrgb` or similar?)

I have tried without the `_` and that doesn't seem to be enough.

evq:
What do you think about creating multiple travis jobs? One runs tests with `std` as a required feature and the other runs tests with `--release` and would not have `std` as a required feature.

palette/src/convert.rs
Ogeon: Ugh, good job on me for not reading it all. It's also `f64`, not the default `f32`, so that's probably why `_` has to be there.

Uhm... I usually prefer to spell it out if I can... You could put a `self` in the encoding import and at least shorten it down to `encoding::Srgb`, I guess. :) If that makes a difference. I'm just trying to put the focus on the right hand expression, but it's not that important.

Ogeon:
There's one script (test_features.sh) that runs all the features one by one. I suppose that one can run with `--release` instead. Then it will test with both in the same job. It's just too bad it's going to increase the build time, but let's try and see how big the impact is.

palette/src/convert.rs
Ogeon: Oh, you can create an alias like `type Hsv64 = Hsv<encoding::Srgb, f64>;`!

evq:
If we only run with `--release` we might miss other overflow checks though. If we create two travis jobs  they can run in parallel so we can both get overflow checks except for `no_std` and test `no_std`.

Ogeon:
I don't think it's going to be a problem, really. If you look at https://github.com/Ogeon/palette/blob/master/.travis.yml, it starts out by compiling and running with the default features. Then it runs the script and most of that script runs with `std` disabled. That's why we would not lose much on changing the script to run with `--release`. The alternative could be to change the script to run everything with `std` enabled, and then run a final test pass with everything disabled and `--release` enabled.

Ogeon:
Also, I'm afraid we may be maxing out on parallel runners already, so I don't think we will save any time on that unless they increased the limit.

Ogeon:
Sorry, I accidentally clicked the close and comment button.

evq:
Ah gotcha! So we'd hit the default features in debug mode and individual features in release mode. That sounds good to me - will adjust

Ogeon:
I think that will cover enough of the cases to give a good enough result. The optional features are mostly just utilities.

evq:
@Ogeon I think this should be good to go now

Ogeon:
Nice! I didn't have the time to check it until now, but it looks good! I'll try to get a release out this weekend. I would just like to check if the rest of the code is in a good state first.

Ogeon:
bors: r+

Ogeon:
Oh, and once again, thanks for the help!

Issue #121, opened 2018-10-20
Support for BT.709


HeroicKatora:
## Requested features

It would be awesome to have support for the ITU-R Recommendations family of color representations. The primary ones of concern for today are `BT.709`, `BT.2020` and `BT.2100` for HD, 4k and high-dynamic range respectively. 

## Motivation

These are in use by video codes and other media alike, most prominently mpeg using the `YCbCr` encoding of `BT.709`. This would lay the ground work for implementing correct consumers and producers of these raw video codecs in rust. Additionally, it would improve the state of interoperability with external libraries such as `ffmpeg` without sacrificing performance (such as copying/converting image data with another ffi).

## Open questions

There standards are defined for various different bit depths. It is an open question on how this is best mapped to the api. Staying in floating point representation is not enough, as the quantization is not defined as a simply scaling.

## References

<https://en.wikipedia.org/wiki/Rec._709>
<http://www.itu.int/rec/R-REC-BT.709-6-201506-I/en>
<https://en.wikipedia.org/wiki/Rec._2020>
<https://en.wikipedia.org/wiki/Rec._2100>

Ogeon:
That sounds useful. Thanks for the detailed suggestion! I suppose the bit depths can be solved by having a more opaque version that doesn't expose individual channels as properties, but I'm not familiar enough with them to say for sure. If they are only meant for encoding they may not have to be as flexible as RGB, for example. They should at least be implemented as close to specification as possible.

HeroicKatora:
I've started with `601` and `709` for now expecting `2020` to also be included. I'm not going to atempt `Rec.2100` though since I'm not completely sure about the patent situation with regards to both transfer functions and encoded color spaces. Furthermore, it adds to the electro-optical transfer function and its inverse an additional optic-optical transfer function which can either appear before the encoding (applied to linear scene luminance), or after decoding (applied to intended display luminance) depending the chosen model. This seems a bit too complex for the first contribution :)

Ogeon:
Excellent! And yes, take it in small steps. It's better to iterate than trying to swallow it whole if it's a large effort. Start with the simple ones. You can always submit an early PR if you want feedback!

Keavon:
[...]

PR #131, opened 2019-05-19
Add non-quantized support for Rec.601 and Rec.709

HeroicKatora:
Adds the color spaces and transfer functions for the color spaces defined in 
ITU-R BT601 and ITU-R BT709 (aka. Rec.601 and Rec.709). Also adds the
notion of quantization functions for these two standards but only implements 
8-bit quantization for the moment. The available implementation is partially
intended as reference, with future additions adding more comprehensive
support for conversion.

Addresses pieces of #121 and lays groundwork for the other recommendations.

### DIFF
--- a/palette/src/yuv/mod.rs
+++ b/palette/src/yuv/mod.rs
@@ -1,0 +1,73 @@
+//! YUV types, spaces and standards.
+use float::Float;
+
+use encoding::{TransferFn};
+use rgb::RgbSpace;
+use {Component};
+
+mod yuv;
+
+/// A YUV standard with digital quantization function.
+pub trait YuvStandard {
+    /// Underlying color space of the RGB signal.
+    type RgbSpace: RgbSpace;
+
+    /// The transfer function from linear RGB space.
+    type TransferFn: TransferFn;
+
+    /// The normalized color difference space.
+    type Differences: Differences;
+}
+
+/// Gives the YUV space values of each primary.
+pub trait Differences {
+    /// The weights of the luminance transform.
+    ///
+    /// The linear transform is assumed to happen after the opto-electric transfer function is
+    /// applied to each color value. This is true for all ITU-R standards. Nevertheless, A
+    /// different form of encoding exists, called YcCbcCbr or constant luminance, which calculates
+    /// the luminance value from the linear RGB values instead to optimize the accuracy of its
+    /// result.
+    fn luminance<T: Float>() -> [T; 3];
+
+    /// Normalize the difference of luminance and blue channel.
+    fn norm_blue<T: Float>(denorm: T) -> T;
+
+    /// Denormalize the difference of luminance and blue channel.
+    fn denorm_blue<T: Float>(norm: T) -> T;
+
+    /// Normalize the difference of luminance and red channel.
+    fn norm_red<T: Float>(denorm: T) -> T;
+
+    /// Denormalize the difference of luminance and red channel.
+    fn denorm_red<T: Float>(norm: T) -> T;
+}
+
+/// A digital encoding of a YUV color model.
+///
+/// While the difference conversion is mostly performed in an analog signal space free of
+/// quantization errors, the final digital output is quantized to some number of bits defined in
+/// individual standards.
+///
+/// The direct conversion of digitally quantized, gamma pre-corrected RGB is also possible. This
+/// yields minor differences compared to a conversion to analog signals and quantization. A strict
+/// integer arithmetic quantization is available as well where performance concerns make the
+/// floating point conversion less reasonable.
+pub trait QuantizationFn<Y: YuvStandard> {
+    type Output: Component;
+
+    /// Quantize an analog yuv pixel.
+    fn quantize<F: Component + Float>(yuv: [F; 3]) -> [Self::Output; 3];
+
+    /// Quantize from an rgb value directly.
+    fn quantize_direct<F: Component + Float>(rgb: [F; 3]) -> [Self::Output; 3];
+
+    /// Transfer a quantized color value.
+    fn quantized_rgb(rgb: [Self::Output; 3]) -> [Self::Output; 3];
+}
+
+impl<R: RgbSpace, T: TransferFn, D: Differences> YuvStandard for (R, T, D) {
+    type RgbSpace = R;
+    type TransferFn = T;
+    type Differences = D;
+}
--- a/palette/src/yuv/mod.rs
+++ b/palette/src/yuv/mod.rs
@@ -40,25 +40,30 @@ pub trait DifferenceFn {
     fn luminance<T: Float>() -> [T; 3];
 
     /// Normalize the difference of luminance and blue channel.
-    fn norm_blue<T: Float>(denorm: T) -> T;
+    fn normalize_blue<T: Float>(denorm: T) -> T;
 
     /// Denormalize the difference of luminance and blue channel.
-    fn denorm_blue<T: Float>(norm: T) -> T;
+    fn denormalize_blue<T: Float>(norm: T) -> T;
 
     /// Normalize the difference of luminance and red channel.
-    fn norm_red<T: Float>(denorm: T) -> T;
+    fn normalize_red<T: Float>(denorm: T) -> T;
 
     /// Denormalize the difference of luminance and red channel.
-    fn denorm_red<T: Float>(norm: T) -> T;
+    fn denormalize_red<T: Float>(norm: T) -> T;
 }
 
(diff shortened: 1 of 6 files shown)


palette/src/yuv/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn denorm_blue<T: Float>(norm: T) -> T;
+
+    /// Normalize the difference of luminance and red channel.
+    fn norm_red<T: Float>(denorm: T) -> T;
+
+    /// Denormalize the difference of luminance and red channel.
+    fn denorm_red<T: Float>(norm: T) -> T;
+}
+
+/// A digital encoding of a YUV color model.
```
Ogeon: So this is different from the conversation from other float to int format conversation, right? I see the place where it's used doesn't look like a straight rescale and rounding. Might want to make that clear for the uninitiated, so they know the difference compared to the `into_format` method. 🙂

palette/src/yuv/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// While the difference conversion is mostly performed in an analog signal space free of
+/// quantization errors, the final digital output is quantized to some number of bits defined in
+/// individual standards.
+///
+// TODO:
```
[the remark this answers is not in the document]
HeroicKatora: Sounds good. I'll open individual ones and reference the corresponding numbers.

palette/src/yuv/mod.rs
HeroicKatora: The standard leaves some headroom and footroom, so this is not straightup scaling. I intended to do at least one pure documentation pass and will add some additional context here. Also, code examples overall probably.

HeroicKatora:
>  Do you think it would be possible to add a few tests, just to make sure it's not broken in the future?

Can do. I'll see what official (or unofficial) test vectors I can find.

palette/src/yuv/mod.rs
Ogeon: Ok, that sounds good. :+1: It doesn't have to be a signal processing tutorial, just something that prevents mistakes.

palette/src/yuv/mod.rs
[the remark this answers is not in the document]
Ogeon: You can add a checklist to the original issue, unless it's too unrelated.

palette/src/yuv/quant.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    fn quantize_rgb<F: Component + Float>(rgb: [F; 3]) -> [u8; 3] {
+        Self::quantize_rgb(rgb)
+    }
+}
+
+/// Round to 8-bit integer in valid signal range.
+fn int_u8<F: Component + Float>(value: F) -> u8 {
+    // Note: signal level below 1 and the level 255 and above are reserved.
+    clamp(value.round(), cast(1.), cast(254.)).convert()
```
Ogeon: Is `convert` the right thing to use here? It may still produce the correct value, but after some unnecessary scaling back and forth. I think you can just round it and cast it.

palette/src/yuv/quant.rs
```diff
@@ -0,0 +1,55 @@
...
+    // Representation of 254.75 with 2 fractional bits.
+    const UPPER_MAX: u16 = 4*254 + 3; // = 1019
+    // Representation of 1.0 with 2 fraction bits.
+    const LOWER_MIN: u16 = 4;
+
+    // Add two fractional bits.
+    let value = value*cast(4.);
+
+    // Note: signal level below 1 and the level 255 and above are reserved.
+    clamp(value.round(), cast(LOWER_MIN), cast(UPPER_MAX)).convert()
```
Ogeon: This `convert` is also questionable...

Ogeon: And the comment mentions 255. :slightly_smiling_face:

HeroicKatora: Is `convert` the scaling one? I suppose both this and the one in `u8` should instead be `NumCast::from`

Ogeon:
The technical parts are going to be

Ogeon:
Nice work! The most technical parts are going to be a bit hard for me to get into (need to study the documents and dig up some of my old signal processing knowledge), but your code is pretty straight forwards if I don't try to verify the actual numbers.

palette/src/yuv/quant.rs
Ogeon: it's meant to scale between a 0.0 - 1.0 scale and any unsigned integer (where it will span the full value range), so it's going to nomalize it to 0.0 - 0.1 and then scale it back to 0 - 255. You can just use `NumCast::from` or `cast` (which is basically the same idea), but you may also want to round it. I don't know if it's supposed to be rounded or floored in this case.

Ogeon: Or it may not even normalize it, since it's already a float. It may actually just output garbage here. :astonished: It's going to run [this one](https://github.com/Ogeon/palette/blob/master/palette/src/lib.rs#L549) or the one for f64, if I'm reading it correctly, so it multiplies with 255.

HeroicKatora: Ah, then `NumCast::from` is the one to take. The rounding is supposed to already have been applied to the `F: Float` itself, and it should be an actual rounding with `.5` upwards (that's actually not completely true for IEEE floats afaik but good enough in this range since the half will never be the last bit of the number).

Ogeon: (Never mind about the rounding, I somehow missed that you are already doing that)

Ogeon: But yes, you want to just cast it in this case. The naming could be better ("convert" is very unspecific), but I think I had a hard time coming up with something that didn't make it worse. :laughing:

HeroicKatora:
I somehow don't find any large enough collection of test vectors. Mostly blog posts, stackoverflow, and seldomly referencing the standard directly. Any guidance on this? The tests could be on a higher level such as testing that the value ranges are not exceed or something.

Ogeon:
I'm totally ok with something simple, like a handful of colors (white, gray, black, red, green, blue, for example), or some range tests if you have a plan for that. That should cover most of it. 🙂

HeroicKatora:
Found some tables for `BT.601` in the standard that I had previously overlooked. Also added some implementations for `approx::*` as far as I could see useful.

Ogeon:
Nice! And good catch with the inversion! It looks like [the build fails on nightly](https://travis-ci.org/Ogeon/palette/jobs/543539816) because they are deprecating trait objects without `dyn`. It's only highlighting a handful of places, so it doesn't look too bad. Do you think you could just add those in too?

Otherwise I think it looks like a good first step. I was just thinking, maybe it would be a good idea to merge this into a separate branch for now. That way it's less problematic to merge a work in progress. Or maybe I should just release what's in master before merging this. It's about time for that anyway.

HeroicKatora:
Good to know that nightly is actually enforcing this now, must be very recent since I remember auto-fixing those missing `dyn` annotations in another library but just as cleanup not because something broke.

Wouldn't it be better to fix `dyn` on nightly in a seperate PR though and retest this one afterwards? That change should probably land before a new release independent of this one.

Ogeon:
True, that would be better.

Ogeon:
It's on its way in #132.

HeroicKatora:
@Ogeon Ping. Is there anything else to address?

Ogeon:
Wow, I totally missed your merge! I would usually get an email when there's a new commit but not for that one... Sorry about that.

Considering this is just the first part, I have decided I would prefer to keep this out of master until the remaining parts in your checklist are done. It makes it easier to merge and release other changes independently in the meantime. I'll create a branch for you if you want to merge this PR. The alternative is to keep adding to this one.

How does that sound?

HeroicKatora:
The mail(s) may have been overshadowed by the failing builds but no worries. If you want to not yet merge this to `master` until more parts to checklist are cleared I would prefer a temporary feature branch as a more stable point of reference of *finished* parts of the checklist in contrast to a constant work-in-progress PR.

Ogeon:
Alright, I have created the `itu-r` branch for you, assuming that's an ok umbrella name for this. :slightly_smiling_face: I think you can just edit this PR, at the top of this page, and pick that one as the target without losing any of history.

HeroicKatora:
Yup. Assuming I don't hit a `http 500` from github but apparently trying again was enough.

Ogeon:
Nice! I think we are done here for now. Sorry, once again, about not seeing the merge. This could have been a lot smoother otherwise. Looking forward to seeing the next part!

PR #137, opened 2019-07-13
Add some missing `From` impls between `Srgb` and `LinSrgb` types

mitchmindtree:
I ran into an issue downstream when attempting to write a generic
function that would accept any type of colour that may be converted to
`Srgb<T>`. I noticed implementations seemed to exist for all of the
colour types I cared about apart from `LinSrgb<T>`, so I thought I'd
implement the missing conversions and open a PR!

Let me know if there's a reason that these have been omitted that I'm
overlooking, or if you have a better way in mind for adding these
conversions!  Unfortunately, conversions between entirely generic
`Rgb<S, T>` and `Rgb<St, U>` are impossible due to the existing blanket
`impl From<T> for T`, but adding individual implementations for the RGB
standards we care about doesn't seem so bad :)

### DIFF
--- a/palette/src/rgb/mod.rs
+++ b/palette/src/rgb/mod.rs
@@ -1,6 +1,5 @@
 //!RGB types, spaces and standards.
 
-use alpha::Alpha;
 use float::Float;
 use core::any::Any;
 
@@ -100,9 +99,9 @@ where
     U: Component,
 {
     fn from(lin_srgb: LinSrgb<T>) -> Self {
-        let color = lin_srgb.into();
-        let alpha = U::max_intensity();
-        Alpha { color, alpha }
+        let non_lin = Srgb::<T>::from_linear(lin_srgb);
+        let new_fmt = Srgb::<U>::from_format(non_lin);
+        new_fmt.into()
     }
 }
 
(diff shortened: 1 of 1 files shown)


Ogeon:
Hi! There were no particular reasons for not adding these, other than that I tend to not add too many convenience things from the start. There are a lot of possible combinations that can be made and it becomes hard to keep track of all of them after a while. But these are pretty standard so I see no harm in having them.

palette/src/rgb/mod.rs
```diff
@@ -70,3 +71,75 @@ pub trait Primaries: Any {
...
+
+impl<T, U> From<LinSrgb<T>> for Srgba<U>
+where
+    T: Component + Float,
+    U: Component,
+{
+    fn from(lin_srgb: LinSrgb<T>) -> Self {
+        let color = lin_srgb.into();
+        let alpha = U::max_intensity();
+        Alpha { color, alpha }
```
Ogeon: Hmm... You should not have to construct the `Alpha` like this. You should be able to use `from_linear` and `into_format` on these too. Otherwise that's also a hole in the API.

Ogeon:
Hi! Just checking to see if you are still interested in picking this up again. My only reason for not merging straight away is that I would like to know if it can be implemented in terms if `from`, `{from,into}_linear` and `{from,into}_format`. They should be available for every `Alpha<Rgb>` type.

mitchmindtree:
Sorry about the delay @Ogeon, I think I've now addressed the issues you mentioned!

Ogeon:
No worries, this looks great! Thanks!

PR #140, opened 2019-08-11
Add a basic test crate for no_std

### LINKED ISSUE #139: Add a test crate for testing no_std builds
As mentioned in https://github.com/Ogeon/palette/pull/138#issuecomment-519881908, testing the `no_std` support seem to require a whole separate crate to avoid leaks from the dev dependencies. The current tests does not catch any compile errors from not having `std`, so this is necessary for knowing if that feature even works.

Also, if there is a better way, then that is of course preferred.

Ogeon:
It's very bare bones, but will hopefully get the job done.

Closes #139

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,12 @@ script:
   - cargo test -v --features strict
   - bash ../scripts/test_features.sh
 
+  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
+      cd ../no_std_test;
+      cargo build -v;
+    fi
+
   - cd ..
-  - cargo doc
+  - cargo doc --all --exclude no_std_test
 after_success:
   - sh scripts/upload_doc.sh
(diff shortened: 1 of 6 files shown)


CryZe:
Whoops I missed this PR. You definitely need to build against a target that doesn't have std to properly check for it as no_std crates can depend on std crates (for whatever reason), so you don't know if all your dependencies are properly no_std unless you compile against a target that doesn't even have it.

Ogeon:
Ouch, alright, I though it worked because it failed to compile (duplicate symbols) when I enabled std. Maybe that's not reliable enough. Do you have an example of how to do it in a better way?

CryZe:
I guess what you have is fine, but I'd additionally also compile it for `--target thumbv6m-none-eabi` (and then probably just not execute it, unless you want to go through the effort with qemu)

Ogeon:
I'm fine with not executing it for now. That's already how it's done here. I will try adding a test for `thumbv6m-none-eabi` too. Thanks for pointing it out!

PR #141, opened 2019-08-11
Also test no_std with a target without std support

Ogeon:
As suggested in https://github.com/Ogeon/palette/pull/140#issuecomment-520235759.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,11 @@ script:
       cargo build -v;
     fi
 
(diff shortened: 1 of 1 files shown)


Ogeon:
Looks like it worked, but it became really hard to read. I'll split it into two separate commands.

Also CC @CryZe, just in case you didn't see it yet. :slightly_smiling_face:

CryZe:
nice 👍

PR #146, opened 2019-09-23
bump syn, quote and proc-macro2 to version 1.0

Veykril:
Just a simple version bump for the dependencies, and a few line fixes where the api changed. This currently does not get rid of the old versions in the `Cargo.lock` due to `num-derive`'s newest version not having upgraded to the 1.0 releases yet. From what I've seen that will be fixed once https://github.com/rust-num/num-derive/pull/29 lands there.

### DIFF
--- a/palette_derive/Cargo.toml
+++ b/palette_derive/Cargo.toml
@@ -14,9 +14,9 @@ license = "MIT OR Apache-2.0"
 proc-macro = true
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


Ogeon:
Nice, thanks! The num-derive situation may have been resolved by the time I release the next version anyway, so that's fine. I'm hoping for a new release of num-traits with some no-std improvements.

Ogeon:
bord r+

PR #154, opened 2019-11-26
Add DoubleEndedIterator impl for gradient::Take

### LINKED ISSUE #153: impl std::iter::DoubleEndedIterator for palette::gradient::Take
Would allow doing `gradient.take(n).rev()` without collecting. My main reasoning for that is basically because for lower values of `n` (1 or 2), `gradient.take(n)` is a bit skewed to the first colors of the gradient.

jansegre:
`gradient.take(n).rev()` is now possible, closes #153

### DIFF
--- a/palette/src/gradient.rs
+++ b/palette/src/gradient.rs
@@ -101,7 +101,8 @@ impl<C: Mix + Clone> Gradient<C> {
             from: min,
             diff: max - min,
             len: n,
-            current: 0,
+            from_head: 0,
+            from_end: 0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


Ogeon:
Looks like a fine solution for preserving the current behavior. :+1: An alternative could have been to use start position, end position and step size, but I don't know if it improves anything.

The skew towards the beginning of the range should also be fixed at some point, as I mentioned, but it may be better to do it in a separate PR. It's technically a breaking change, while this doesn't have to be.

The formatting is fine, but you can always just stage and commit the file you have changed, and reset the rest if you want. Either way, I would appreciate it a lot if you can write a short description and include the "closes ..." part in the PR description/initial comment. That makes it show up nicely in the merge commit. :slightly_smiling_face: I should really add a PR template for this...

jansegre:
OK, I added the "closes #<issue>" to the title. But I guess it's better on the commit message body alongside a short description. I'll do that and amend this commit.

Ogeon:
Just to be clear, I'm talking about your first comment in this thread, where you mentioned `cargo fmt`. That goes into the merge commit.

jansegre:
Oh, sorry I didn't realize the PR title and description ended up in the merge commit. Almost hit push here.

jansegre:
Is this good?

Ogeon:
That's excellent, thanks! It's not easy to know, but quite nice IMO. I just need to make that template, so it's more obvious...

Ogeon:
I just took the liberty to remove it from the title, so it's not double.

PR #157, opened 2019-11-30
implement FromStr method for Srgb<f32>.

### LINKED ISSUE #148: Working with Hex codes
Are there any plans for adding support for hex format, specifically something like rgb::new(hex) ?

mswift42:
This enables converting hex colors like '#00ddff' or '#fff' to Rgb. 

this closes #148.

### DIFF
--- a/palette/src/rgb/rgb.rs
+++ b/palette/src/rgb/rgb.rs
@@ -2,6 +2,8 @@ use core::any::TypeId;
 use core::fmt;
 use core::marker::PhantomData;
 use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Sub, SubAssign};
+use core::str::FromStr;
+use core::num::ParseIntError;
 
 use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 use float::Float;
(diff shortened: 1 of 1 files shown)


palette/src/rgb/rgb.rs
```diff
@@ -2,6 +2,7 @@ use core::any::TypeId;
 use core::fmt;
 use core::marker::PhantomData;
 use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Sub, SubAssign};
+use std::str::FromStr;
```
Ogeon: This need to be changed to get it from `core` instead, for the `#[no_std]` support to work.

```suggestion
use core::str::FromStr;
```

palette/src/rgb/rgb.rs
```diff
@@ -953,14 +954,50 @@ where
     }
 }
 
+impl FromStr for Rgb {
```
Ogeon: While this may be the more convenient option for many cases, I would still prefer to implement it for `Rgb<Srgb, u8>`. I think that's less surprising than only having it for `f32`. Ideally both would be supported, but then what about if it's a `u16` hex string? Or even larger? So, in the end I think it's better to start small, as you have, but for `Rgb<Srgb, u8>`.

```suggestion
impl FromStr for Rgb<Srgb, u8> {
```

palette/src/rgb/rgb.rs
```diff
@@ -953,14 +954,50 @@ where
...
+    type Err = Box<dyn std::error::Error>;
```
Ogeon: This will have to change to something that isn't `Box`ed, due to the `#[no_std]` support requirement. A simple error type like

```rust
enum FromHexError {
    ParseIntError(ParseIntError),
    HexFormatError(&'static str), // Like "invalid hex code length" or "unexpected hex code prefix"
}
```

with all the impls it would need, would do the trick.

palette/src/rgb/rgb.rs
```diff
@@ -953,14 +954,50 @@ where
...
+
+    // parses a color hex code of format '#ff00bb' or '#abc' into a
```
Ogeon: It mentions "#" being part of the format, but never checks that it's there. Better to be strict and relax the rules later, rather than the opposite. Besides, it would also be technically possible to support strings without the "#". Maybe that should be added too?

Also, capital "P" in "parses". :slightly_smiling_face:

palette/src/rgb/rgb.rs
Ogeon: Thinking some more about this, I don't think there's any reason to limit it to only `Srgb`.

```suggestion
impl<S: RgbStandard> FromStr for Rgb<S, u8> {
```

It would also be great to have one for `Alpha<Rgb<S, u8>>`.

palette/src/rgb/rgb.rs
mswift42: I struggle with this. Sorry, I am fairly new to rust, but there is no Error type in core only in std.

Do you have a suggestion how to do this ?

Ogeon: Ah, ok. No worries! :) You are right, it seems like the `Error` trait is only in `std`. In that case we will have to make use of conditional compilation if we want to use it (it's good practice to implement `Error`, so we should). `#[no_std]` is a particularly prickly feature. If we use `std` anywhere, unconditionally, it's all out the window, and it's hard to notice unless it's used on a platform that doesn't support the `std` library. I have set up a CI test to detect it, and that's what made this PR fail the tests.

For this case we have to give up the convenience of having the boxed trait in either case, since `Box` is also not available, so the first step would be to add an enum like the one I suggested and change this line like this:

```suggestion
    type Err = FromHexError;
```

That way, the parsing will work both with and without `std`.

Then the `FromHexError` type will need some traits (I would look at `ParseIntError` for inspiration), including `Error`. That's where we will have to check if the library user is allowing the `std` library:

```rust
#[cfg(feature = "std")] // Checks if the "std" feature from Cargo.toml is enabled
impl std::error::Error for FromHexError {
    // `std` is safe to use in this `impl` item, because of the `cfg` attribute.
    // It will not be included at all if the "std" feature is disabled.
    //
    // ...
}
```

The same goes for any other trait that is not in `core`.

mswift42:
awesome, thanks.

palette/src/rgb/rgb.rs
```diff
@@ -953,14 +955,82 @@ where
...
+
+impl<S: RgbStandard> FromStr for Rgb<S, u8> {
+    type Err = FromHexError;
+
+    // Parses a color hex code of format '#ff00bb' or '#abc' into a
+    // Rgb<S, u8> instance.
+    fn from_str(hex_code: &str) -> Result<Self, Self::Err> {
+        match hex_code.len() {
+            3 | 4 | 6 | 7 => {
+                let (red, green, blue, factor) = if hex_code.len() == 3 {
```
Ogeon: I'm just stopping by to see how it's going. Looks promising! I would just recommend to either drop this if-else chain and only use the match, or the other way around. Both would work. As it is now, you are checking the length twice, at least. :slightly_smiling_face:

Ogeon: An additional idea, to halve the amount of cases, is to check for the existence of `#` at the beginning first. If it's there, you slice it away. That way, you are down to only two possible lengths when parsing the components.

mswift42: Yes, I've done it this way. (December is a really busy month, :)

Ogeon: Indeed, it is. But don't feel like you have to rush it or anything. :)

Ogeon:
I think this looks really good! The only addition I think would be good to have would be to implement `Error` for `FromHexError`, as we talked about. :slightly_smiling_face: After that I'm definitely prepared to merge this, if you are happy with it. But, once again, take your time and let me know if anything is still unclear.

palette/src/rgb/rgb.rs
```diff
@@ -953,14 +955,92 @@ where
...
+}
+impl core::fmt::Display for FromHexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &*self {
+            FromHexError::ParseIntError(e) => {
+                write!(f, "ParseIntError: {}", e)
+            }
+            FromHexError::HexFormatError(s) => {
+                write!(f, "HexFormatError: {}, please use format '#fff', 'fff', '#ffffff' or\
+                'ffffff'.", s)
```
Ogeon: I think it would be better to skip the `ParseIntError:` and `HexFormatError:` prefixes. I don't think they will be that interesting to someone who would read the error message, to be honest. They just want to know what went wrong and how to fix it. :slightly_smiling_face:

Ogeon:
This looks really nice! Well done! There's just that detail with the error messages. Also, I think the common convention in general is to just have a plain sentence without an initial capital letter, so there's that too.

mswift42:
Ok, done, and a happy new year to you :champagne:

Ogeon:
Thanks and happy new year to you too! Oh, I forgot, you may want to squash those commits. Just to keep the history a bit shorter.

Ogeon:
Sorry, I didn't see the notification for the squash. It must have gotten lost somewhere...

Thanks you!

bors r+

Issue #156, opened 2019-11-26
Redesign and improve the Gradient type and friends


Ogeon:
It should have about the same API as today, where it still makes sense, but overall modernized. Also fulfill this:

* [ ] make it `#[no_std]` enabled
* [ ] support more than just linear interpolation
* [x] the `Take` iterator should be inclusive, meaning both ends are part of the output for n > 1
* [ ] support slicing with at least [`x..=y`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html), but other ranges may also make sense

The first two points may go hand in hand, by replacing the internal `Vec` with some input collection.

okaneco:
To learn more about [no_std], I looked into making gradient [no_std] enabled and [built it successfully](https://travis-ci.org/okaneco/palette/jobs/627020703#L1852) for `thumbv6m-none-eabi`. Since the [`alloc` crate](https://doc.rust-lang.org/alloc/index.html) has been stabilized, you can use some collections like `Vec` with no_std. That required changing instances of `std::vec` to `core::vec` in `gradient.rs`, configuring the `alloc` crate behind a feature flag in `lib.rs`, and adding some helper code in the `no_std_test.rs` like the alloc error handler. For data structures and no_std, it seems like the choices are use  `alloc` and make users specify their own memory allocator in their crates/runtimes, or bring in a crate like [`heapless`](https://docs.rs/heapless/0.5.1/heapless/) for your internal data structures (I didn't want to add any new dependencies). I'm very inexperienced with this though, this is what I can gather from the current information I could search.

The no_std_test [fails to compile](https://travis-ci.org/okaneco/palette/jobs/627020703#L1802) for `x86_64-unknown-linux-gnu` target in the same CI build. It doesn't seem like there's a solution currently for the `undefined reference to '_Unwind_Resume'`. It seems to be an issue with the linker and I couldn't get the rust-ld linker to work as I believe you'd have to specify library locations in the VM image. `eh_unwind_resume` doesn't seem to matter for helping linux-gnu build and thumbv6m didn't need it to build. I don't have Linux installed at the moment so can't test it locally, but I can build it locally for `x86_64-apple-darwin` on nightly.

[Here](https://github.com/okaneco/palette/pull/2/files) are the changed files.

Ogeon:
[...]

PR #158, opened 2019-12-12
Make `Take` iterator for gradient inclusive of both end colors, add tests

okaneco:
As mentioned in the third point of #156, this PR includes the bounds of the gradient along with more equal distribution of colors returned from the iterator.

In the current version of the iterator, `(self.from_head) / (self.len)` will approach 1 but never equal 1 which results in the upper gradient bound being excluded. The current behavior of `take()` is surprising if one expects the last color step to be the max color of the gradient. This PR increases step distance so the user receives an iterator of colors that evenly traverses from the minimum bound of the gradient to the maximum. The step is increased by subtracting 1 from `self.len` in the calculation of `i`. The calculation for `i` will then include
- `0` when `self.from_head` is `0`
- `1` when `self.from_head == self.len - 1`. 

The assignment of `i` has a conditional for when `self.len` is `1`. This avoids division by 0 which results in NaN parameters for the iterator color returned in the case of `take(1)`. The check needed to be added to the `DoubleEndedIterator`, otherwise it would NaN on `1` as well. The behavior of `take(1)` before this PR is to supply the minimum color in the gradient.

Pictured below is an example of the difference between current behavior and this PR in 3 sets of Lch gradients. The 1st, 3rd, and 5th rows are the current implementation of the iterator and the others are from the PR with the final step being inclusive of the maximum gradient color specified. The gradients are no longer skewed towards the beginning. The penultimate steps are very close to the current behavior's final step.
image of 6 stepped gradients

---

`gradient::test::simple_slice` fails as a result of this PR.
```
---- gradient::test::simple_slice stdout ----
thread 'gradient::test::simple_slice' panicked at 'assert_relative_eq!(t1, t2)

    left  = Rgb { red: 0.8888888888888888, green: 0.0, blue: 0.1111111111111111, standard: PhantomData }
    right = Rgb { red: 0.875, green: 0.0, blue: 0.125, standard: PhantomData }

', palette/src/gradient.rs:453:13
```
```rust
#[test]
    fn simple_slice() {
        let g1 = Gradient::new(vec![
            LinSrgb::new(1.0, 0.0, 0.0),
            LinSrgb::new(0.0, 0.0, 1.0),
        ]);
        let g2 = g1.slice(..0.5);

        let v1: Vec<_> = g1.take(10).take(5).collect();
        let v2: Vec<_> = g2.take(5).collect();
        for (t1, t2) in v1.iter().zip(v2.iter()) {
            assert_relative_eq!(t1, t2);
        }
    }
```


okaneco:
The simple slice test passes when
```rust
let v1: Vec<_> = g1.take(10).take(5).collect();
```
is changed to
```rust
let v1: Vec<_> = g1.take(9).take(5).collect();
```

The size of the slice has increased so the test failed. Here are g1.take(10).take(5), g2.take(5).

```
g1.take(10).take(5)
Rgb { red: 1.0, green: 0.0, blue: 0.0, standard: PhantomData }
Rgb { red: 0.8888888888888888, green: 0.0, blue: 0.1111111111111111, standard: PhantomData }
Rgb { red: 0.7777777777777778, green: 0.0, blue: 0.2222222222222222, standard: PhantomData }
Rgb { red: 0.6666666666666667, green: 0.0, blue: 0.3333333333333333, standard: PhantomData }
Rgb { red: 0.5555555555555556, green: 0.0, blue: 0.4444444444444444, standard: PhantomData }
g2.take(5)
Rgb { red: 1.0, green: 0.0, blue: 0.0, standard: PhantomData }
Rgb { red: 0.875, green: 0.0, blue: 0.125, standard: PhantomData }
Rgb { red: 0.75, green: 0.0, blue: 0.25, standard: PhantomData }
Rgb { red: 0.625, green: 0.0, blue: 0.375, standard: PhantomData }
Rgb { red: 0.5, green: 0.0, blue: 0.5, standard: PhantomData }
simple slice test
```

`g1.take(9).take(5)` produces the same results as `g2.take(5)` and passes the test.
```
g1.take(9)
Rgb { red: 1.0, green: 0.0, blue: 0.0, standard: PhantomData }
Rgb { red: 0.875, green: 0.0, blue: 0.125, standard: PhantomData }
Rgb { red: 0.75, green: 0.0, blue: 0.25, standard: PhantomData }
Rgb { red: 0.625, green: 0.0, blue: 0.375, standard: PhantomData }
Rgb { red: 0.5, green: 0.0, blue: 0.5, standard: PhantomData }
Rgb { red: 0.375, green: 0.0, blue: 0.625, standard: PhantomData }
Rgb { red: 0.25, green: 0.0, blue: 0.75, standard: PhantomData }
Rgb { red: 0.125, green: 0.0, blue: 0.875, standard: PhantomData }
Rgb { red: 0.0, green: 0.0, blue: 1.0, standard: PhantomData }
g1.take(9).take(5)
Rgb { red: 1.0, green: 0.0, blue: 0.0, standard: PhantomData }
Rgb { red: 0.875, green: 0.0, blue: 0.125, standard: PhantomData }
Rgb { red: 0.75, green: 0.0, blue: 0.25, standard: PhantomData }
Rgb { red: 0.625, green: 0.0, blue: 0.375, standard: PhantomData }
Rgb { red: 0.5, green: 0.0, blue: 0.5, standard: PhantomData }
```

I've amended the `simple_slice_test`.

palette/src/gradient.rs
```diff
@@ -142,7 +142,11 @@ impl<'a, C: Mix + Clone> Iterator for Take<'a, C> {
 
     fn next(&mut self) -> Option<C> {
         if self.from_head + self.from_end < self.len {
-            let i = self.from + (self.diff / cast(self.len)) * cast(self.from_head);
+            let i = if self.len > 1 {
+                self.from + (self.diff / cast(self.len - 1)) * cast(self.from_head)
+            } else {
+                self.from + (self.diff / cast(self.len)) * cast(self.from_head)
```
Ogeon: I suppose it's possible to get here if `len`, `from_head` and `from_end` are all 0. That wouldn't be great. I think we can make special cases for 0 and 1 length iterators. 0 gives `None` on every call and `1` give the lower bound color (to be consistent with the old behavior) if `from_head` and `from_end` are both 0 (otherwise `None`). That may simplify the rest of the logic.

palette/src/gradient.rs
```diff
@@ -439,7 +447,7 @@ mod test {
         ]);
         let g2 = g1.slice(..0.5);
 
-        let v1: Vec<_> = g1.take(10).take(5).collect();
+        let v1: Vec<_> = g1.take(9).take(5).collect();
```
Ogeon: 9 items is 8 "jumps" and 5 items is 4 "jumps". 5 ends at the middle of 9. It's slightly mind bending (for my post work Friday brain), but it checks out.

okaneco: Yes, I had to use pen and paper to reason about it. My thinking was the interval between each color step is 1/(n-1) for n > 1. So 9 is 1/8, 10 is 1/9, etc. 

When you take 9, you're getting 7 intermediate colors that are 8 "jumps" apart. At first, this can be confusing when reasoning about what to take if you want a specific interval. I wanted to add a short but hopefully clear explanation of this in the docs so future people can avoid bending minds.

okaneco:
Thanks for the feedback. I've updated the documentation in `gradient.rs`, let me know if there are any other places to mention or corrections to be made. Terminology around explaining the change feels very clunky but did my best to streamline it.

I hope I understood your comment. Now, the logic is more explicit about the special cases and the interpolation is isolated. The "let if" was opaque and not ideal.

okaneco:
I updated the readme to mention an example in an updated /gfx/readme_gradients.png. However I wasn't sure if the readme_examples were automatically built at some point. I copied the relevant gradient portion into a local test file to work out, then I reinserted and ran the --features strict test to make sure it built.

Gradient Comparison

palette/src/gradient.rs
```diff
@@ -17,9 +17,9 @@ use Mix;
 ///It's used to smoothly transition between a series of colors, that can be
 ///either evenly spaced or have customized positions. The gradient is
 ///continuous between the control points, but it's possible to iterate over a
-///number of evenly spaced points using the `take` method. Any point outside
-///the domain of the gradient will have the same color as the closest control
-///point.
+///number of evenly spaced points using the `take` method. The `take` method is
+///now inclusive of both ends of the gradient. Any point outside the domain of
+///the gradient will have the same color as the closest control point.
```
Ogeon: The "The `take` method is now inclusive of both ends of the gradient" part kind of reads like a release notes entry. I think we can safely skip that sentence.

palette/src/gradient.rs
```diff
@@ -92,7 +92,14 @@ impl<C: Mix + Clone> Gradient<C> {
...
 
-    ///Take `n` evenly spaced colors from the gradient, as an iterator.
+    ///Take `n` evenly spaced colors from the gradient, as an iterator. Now
+    ///inclsuive of both ends of the gradient. Points are spaced 1/(n-1)
+    ///distance apart for n > 1. There are n-2 intermediate points
+    ///between start and end for n > 2.
+    ///
+    ///For example, `take(5)` will include point 0.0 of the gradient, three
+    ///intermediate colors, and point 1.0 spaced apart at 1/4 the distance
+    ///between colors 0.0 and 1.0 on the gradient.
```
Ogeon: I would suggest shortening the first paragraph to something like

> Take `n` evenly spaced colors from the gradient, as an iterator. The iterator includes both ends of the gradient, for `n > 1`, or just the lower end of the gradient for `n = 0`.

And then add a small example that illustrates the second paragraph:

```rust
let gradient = Gradient::new(vec![
    LinSrgb::new(1.0, 1.0, 0.0),
    LinSrgb::new(0.0, 0.0, 1.0),
]);

let colors = gradient.take(5).collect();
assert_relative_eq!(colors, vec![
    LinSrgb::new(1.0, 1.0, 0.0),
    LinSrgb::new(0.75, 0.75, 0.25),
    LinSrgb::new(0.5, 0.5, 0.5),
    LinSrgb::new(0.25, 0.25, 0.75),
    LinSrgb::new(0.0, 0.0, 1.0),
]);
```

palette/src/gradient.rs
```diff
@@ -142,9 +149,14 @@ impl<'a, C: Mix + Clone> Iterator for Take<'a, C> {
 
     fn next(&mut self) -> Option<C> {
         if self.from_head + self.from_end < self.len {
-            let i = self.from + (self.diff / cast(self.len)) * cast(self.from_head);
-            self.from_head += 1;
-            Some(self.gradient.get(i))
+            if self.len == 1 {
+                self.from_head += 1;
+                Some(self.gradient.get(cast(0)))
```
Ogeon: Shouldn't this rather be `Some(self.gradient.get(self.from))`? The gradient may be a slice in the middle of the parent gradient.

palette/src/gradient.rs
```diff
@@ -160,9 +172,14 @@ impl<'a, C: Mix + Clone> ExactSizeIterator for Take<'a, C> {}
 impl<'a, C: Mix + Clone> DoubleEndedIterator for Take<'a, C> {
     fn next_back(&mut self) -> Option<Self::Item> {
         if self.from_head + self.from_end < self.len {
-            let i = self.from + (self.diff / cast(self.len)) * cast(self.len - self.from_end - 1);
-            self.from_end += 1;
-            Some(self.gradient.get(i))
+            if self.len == 1 {
+                self.from_end += 1;
+                Some(self.gradient.get(cast(0)))
```
Ogeon: `Some(self.gradient.get(self.from))` here too.

okaneco:
I had some trouble getting the macro for assert_relative_eq to work in the doc test example. It ended up looking more like the unit tests.

Ogeon:
That's fine. I just didn't remember if the required trait was implemented for `Vec`, so I took a chance. 

I think this look good to go now! Do you want to squash any of the fixups before it's merged? Otherwise I'll order it to be merged as soon as I get to it tomorrow. :slightly_smiling_face:

Ogeon:
(sorry, clicked the wrong button there)

okaneco:
Great 👍

Do you want to edit the PR title or original comment to mention anything?

Ogeon:
No, I think it's fine the way it is.

bors r+

PR #159, opened 2019-12-15
Convert remaining float asserts to approx

### LINKED ISSUE #21: Avoid equality tests in examples, and use approx otherwise
Floating point numbers are crazy and should not be trusted. They should therefore not be compared in the examples (for aesthetic reasons) and any unit test that compares them should make use of the `approx` crate.

okaneco:
Closes #21. 

Converted `assert_eq!` to `assert_relative_eq!` for float comparisons not involving serialization in examples and tests.

Removed redundant `extern crate palette;` from example tests.

### DIFF
--- a/palette/src/convert.rs
+++ b/palette/src/convert.rs
@@ -100,8 +100,6 @@ use encoding::Linear;
 ///
 /// ```rust
 /// #[macro_use]
-/// extern crate palette;
-/// #[macro_use]
 /// extern crate approx;
 ///
 /// use palette::{Component, FromColor, Hsv, Pixel, Srgb};
@@ -160,7 +158,7 @@ use encoding::Linear;
 ///
 /// ```rust
 /// #[macro_use]
-/// extern crate palette;
+/// extern crate approx;
 ///
 /// use palette::{FromColor, LinSrgba, Srgb};
 /// use palette::rgb::{Rgb, RgbSpace};
@@ -346,8 +344,6 @@ where
 ///
 /// ```rust
 /// #[macro_use]
-/// extern crate palette;
-/// #[macro_use]
 /// extern crate approx;
 ///
 /// use palette::{Component, Hsv, IntoColor, Pixel, Srgb};
@@ -402,8 +398,6 @@ where
 /// With alpha component:
 ///
 /// ```rust
-/// #[macro_use]
-/// extern crate palette;
 /// #[macro_use]
 /// extern crate approx;
 ///
--- a/palette/src/convert.rs
+++ b/palette/src/convert.rs
@@ -157,9 +157,6 @@ use encoding::Linear;
 /// With alpha component:
 ///
 /// ```rust
-/// #[macro_use]
-/// extern crate approx;
-///
 /// use palette::{FromColor, LinSrgba, Srgb};
 /// use palette::rgb::{Rgb, RgbSpace};
 /// use palette::encoding::Linear;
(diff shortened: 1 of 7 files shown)


palette/src/convert.rs
```diff
@@ -201,15 +199,10 @@ use encoding::Linear;
...
-///             red: 188,
-///             green: 0,
-///             blue: 255,
-///             alpha: 0.3,
-///         }
-///     );
+///     assert_eq!(css_color.red, 188);
+///     assert_eq!(css_color.green, 0);
+///     assert_eq!(css_color.blue, 255);
+///     assert_relative_eq!(css_color.alpha, 0.3);
```
Ogeon: I think this one is safe to make an exception for. It's not going to alter the alpha value in any way.

palette/src/encoding/pixel/mod.rs
```diff
@@ -52,23 +52,18 @@ mod raw;
...
-///             cyan: 0.1,
-///             magenta: 0.2,
-///             yellow: 0.3,
-///             key: 0.4,
-///         }
-///     );
+///     assert_relative_eq!(color.cyan, 0.1);
+///     assert_relative_eq!(color.magenta, 0.2);
+///     assert_relative_eq!(color.yellow, 0.3);
+///     assert_relative_eq!(color.key, 0.4);
```
Ogeon: This one and the other one in the same file can also use `assert_eq`. They will just copy the data.

Ogeon:
I'll stop commenting on individual changes, but anything that is related to (de)serializing should probably still be `assert_eq`, since they should preserve the original value. The issue for this is old and didn't take that into account.

okaneco:
Ok, that makes sense. I reverted those changes.

palette/src/macros.rs
```diff
@@ -43,10 +43,10 @@ macro_rules! raw_pixel_conversion_tests {
 
         let color_alpha: Alpha<$name<$($ty_param,)+ $float>, $float> = *Alpha::<$name<$($ty_param,)+ $float>, $float>::from_raw(&raw_plus_1);
 
-        assert_eq!(color, $name::new($($component),+));
-        assert_eq!(color_long, $name::new($($component),+));
+        assert_relative_eq!(color, $name::new($($component),+));
+        assert_relative_eq!(color_long, $name::new($($component),+));
 
-        assert_eq!(color_alpha, Alpha::<$name<$($ty_param,)+ $float>, $float>::new($($component,)+ alpha));
+        assert_relative_eq!(color_alpha, Alpha::<$name<$($ty_param,)+ $float>, $float>::new($($component,)+ alpha));
```
Ogeon: Pretty sure these should be reverted too.

palette/src/macros.rs
```diff
@@ -76,11 +76,11 @@ macro_rules! raw_pixel_conversion_tests {
...
 
-        assert_eq!(color, $name::new($($component),+));
-        assert_eq!(color_long, $name::new($($component),+));
+        assert_relative_eq!(color, $name::new($($component),+));
+        assert_relative_eq!(color_long, $name::new($($component),+));
 
-        assert_eq!(color_alpha, Alpha::<$name<$($ty_param,)+ $float>, $float>::new($($component,)+ alpha));
-        assert_eq!(color_alpha_long, Alpha::<$name<$($ty_param,)+ $float>, $float>::new($($component,)+ alpha));
+        assert_relative_eq!(color_alpha, Alpha::<$name<$($ty_param,)+ $float>, $float>::new($($component,)+ alpha));
+        assert_relative_eq!(color_alpha_long, Alpha::<$name<$($ty_param,)+ $float>, $float>::new($($component,)+ alpha));
```
Ogeon: ...and these.

Ogeon:
Great! I'll put this in the queue. Thank you!

bors r+

PR #163, opened 2020-01-22
Correcting a word in README.md

damiancarrillo:
Using "featured" instead of "features" in

    It's not a fully featured image manipulation library.


Ogeon:
Good catch! That one must have been there since forever, almost. Thanks!

bors  r+

PR #164, opened 2020-01-25
Implement WCAG contrast ratio calculations

okaneco:
Add RelativeContrast trait for Luma, add tests

Implements contrast ratio calculation for W3C Web Content Accessibility Guidelines (WCAG). Success Criterion 1.4.3 Contrast (Minimum) and Success Criterion 1.4.6 Contrast (Enhanced) suggest text color and background color contrast ratios accessible for people with low vision or color deficiencies. Success Criterion 1.4.11 Non-text Contrast suggests a minimum contrast ratio for adjacent colors graphical elements such as UI components.

[Success Criterion 1.4.3 Contrast \(Minimum\)](https://www.w3.org/TR/WCAG/#contrast-minimum)

[Success Criterion 1.4.6 Contrast \(Enhanced\)](https://www.w3.org/TR/WCAG/#contrast-enhanced)

[Success Criterion 1.4.11 Non-text Contrast](https://www.w3.org/TR/WCAG/#non-text-contrast)

[Contrast ratio](https://www.w3.org/TR/WCAG/#dfn-contrast-ratio)

[Relative luminance](https://www.w3.org/TR/WCAG/#dfn-relative-luminance)

### DIFF
--- a/palette/src/luma/luma.rs
+++ b/palette/src/luma/luma.rs
@@ -5,16 +5,16 @@ use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Sub, SubAssign};
 use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 
 use blend::PreAlpha;
-use clamp;
 use encoding::linear::LinearFn;
 use encoding::pixel::RawPixel;
 use encoding::{Linear, Srgb, TransferFn};
 use luma::LumaStandard;
 use white_point::WhitePoint;
+use {clamp, contrast_ratio};
 use {Alpha, Xyz, Yxy};
 use {
     Blend, Component, ComponentWise, FloatComponent, FromColor, FromComponent, IntoColor, Limited,
-    Mix, Pixel, Shade,
+    Mix, Pixel, RelativeContrast, Shade,
 };
 
 /// Luminance with an alpha component. See the [`Lumaa` implementation
@@ -722,6 +722,21 @@ where
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let size = f.width().unwrap_or(::core::mem::size_of::<T>() * 2);
         write!(f, "{:0width$X}", self.luma, width = size)
+    }
+}
+
+impl<S, T> RelativeContrast for Luma<S, T>
+where
+    T: FloatComponent,
+    S: LumaStandard,
+{
+    type Scalar = T;
+
+    fn get_contrast_ratio(&self, other: &Self) -> T {
+        let luma1 = self.into_luma();
+        let luma2 = other.into_luma();
+
+        contrast_ratio(luma1.luma, luma2.luma)
     }
 }
 
--- a/palette/src/relative_contrast.rs
+++ b/palette/src/relative_contrast.rs
@@ -1,0 +1,147 @@
+use component::Component;
+use core::ops::{Add, Div};
+use {from_f64, FromF64};
+
+/// A trait for calculating relative contrast between two colors.
+///
+/// W3C's Web Content Accessibility Guidelines (WCAG) 2.1 suggest a method
+/// to calculate accessible contrast ratios of text and background colors for
+/// those with low vision or color deficiencies, and for contrast of colors used
+/// in user interface graphics objects.
+///
+/// These criteria are recommendations, not hard and fast rules. Most
+/// importantly, look at the colors in action and make sure they're clear and
+/// comfortable to read. A pair of colors may pass contrast guidelines but still
+/// be uncomfortable to look at. Favor readability over only satisfying the
+/// contrast ratio metric. It is recommended to verify the contrast ratio
+/// in the output format of the colors and not to assume the contrast ratio
+/// remains exactly the same across color formats. The following example checks
+/// the contrast ratio of two colors in RGB format.
+///
+/// ```rust
+/// #[macro_use]
+/// extern crate approx;
+///
+/// use std::str::FromStr;
+/// use palette::{Srgb, RelativeContrast};
+///
+/// fn main() {
+///     // the rustdoc "DARK" theme background and text colors
+///     let my_background_rgb: Srgb<f32> = Srgb::from_str("#353535").unwrap().into_format();
+///     let my_foreground_rgb = Srgb::from_str("#ddd").unwrap().into_format();
+///
+///     assert!(my_background_rgb.has_enhanced_contrast_text(&my_foreground_rgb));
+/// }
+/// ```
+///
+/// The possible range of contrast ratios is from 1:1 to 21:1. There is a
+/// Success Criterion for Contrast (Minimum) and a Success Criterion for
+/// Contrast (Enhanced), SC 1.4.3 and SC 1.4.6 respectively, which are concerned
+/// with text and images of text. SC 1.4.11 is a Success Criterion for "non-text
+/// contrast" such as user interface components and other graphics. The relative
+/// contrast is calculated by `(L1 + 0.05) / (L2 + 0.05)`, where `L1` is the
+/// luminance of the brighter color and `L2` is the luminance of the darker
+/// color both in sRGB linear space. A higher contrast ratio is generally
+/// desireable.
+///
+/// For more details, visit the following links:
+///
+/// [Success Criterion 1.4.3 Contrast (Minimum) (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum)
+///
+/// [Success Criterion 1.4.6 Contrast (Enhanced) (Level AAA)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced)
+///
+/// [Success Criterion 1.4.11 Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)
+
+pub trait RelativeContrast {
+    /// The type of the contrast ratio.
+    type Scalar: FromF64 + PartialOrd;
+
+    /// Calculate contrast ratio between two colors.
+    fn get_contrast_ratio(&self, other: &Self) -> Self::Scalar;
+    /// Verify the contrast between two colors satisfies SC 1.4.3. Contrast
+    /// is at least 4.5:1 (Level AA).
+    fn has_min_contrast_text(&self, other: &Self) -> bool {
+        self.get_contrast_ratio(other) >= from_f64(4.5)
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.3 for large
+    /// text. Contrast is at least 3:1 (Level AA).
+    fn has_min_contrast_large_text(&self, other: &Self) -> bool {
+        self.get_contrast_ratio(other) >= from_f64(3.0)
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.6. Contrast
+    /// is at least 7:1 (Level AAA).
+    fn has_enhanced_contrast_text(&self, other: &Self) -> bool {
+        self.get_contrast_ratio(other) >= from_f64(7.0)
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.6 for large
+    /// text. Contrast is at least 4.5:1 (Level AAA).
+    fn has_enhanced_contrast_large_text(&self, other: &Self) -> bool {
+        self.has_min_contrast_text(other)
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.11 for graphical
+    /// objects. Contrast is at least 3:1 (Level AA).
+    fn has_min_contrast_graphics(&self, other: &Self) -> bool {
+        self.has_min_contrast_large_text(other)
+    }
+}
+
+/// Calculate a ratio between two `luma` values.
+pub fn contrast_ratio<T>(luma1: T, luma2: T) -> T
+where
+    T: Add<Output = T>,
+    T: Div<Output = T>,
+    T: FromF64,
+    T: Component,
+{
+    if luma1 > luma2 {
+        (luma1 + from_f64(0.05)) / (luma2 + from_f64(0.05))
+    } else {
+        (luma2 + from_f64(0.05)) / (luma1 + from_f64(0.05))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use core::str::FromStr;
+    use RelativeContrast;
+    use Srgb;
+
+    #[test]
+    fn relative_contrast() {
+        let white = Srgb::new(1.0, 1.0, 1.0);
+        let black = Srgb::new(0.0, 0.0, 0.0);
+
+        assert_relative_eq!(white.get_contrast_ratio(&white), 1.0);
+        assert_relative_eq!(white.get_contrast_ratio(&black), 21.0);
+        assert_relative_eq!(
+            white.get_contrast_ratio(&black),
+            black.get_contrast_ratio(&white)
+        );
+
+        let c1 = Srgb::from_str("#600").unwrap().into_format();
+
+        assert_relative_eq!(c1.get_contrast_ratio(&white), 13.41, epsilon = 0.01);
+        assert_relative_eq!(c1.get_contrast_ratio(&black), 1.56, epsilon = 0.01);
+
+        assert!(c1.has_min_contrast_text(&white));
+        assert!(c1.has_min_contrast_large_text(&white));
+        assert!(c1.has_enhanced_contrast_text(&white));
+        assert!(c1.has_enhanced_contrast_large_text(&white));
+        assert!(c1.has_min_contrast_graphics(&white));
+        assert!(c1.has_min_contrast_text(&black) == false);
+        assert!(c1.has_min_contrast_large_text(&black) == false);
+        assert!(c1.has_enhanced_contrast_text(&black) == false);
+        assert!(c1.has_enhanced_contrast_large_text(&black) == false);
+        assert!(c1.has_min_contrast_graphics(&black) == false);
+
+        let c1 = Srgb::from_str("#066").unwrap().into_format();
+
+        assert_relative_eq!(c1.get_contrast_ratio(&white), 6.79, epsilon = 0.01);
+        assert_relative_eq!(c1.get_contrast_ratio(&black), 3.09, epsilon = 0.01);
+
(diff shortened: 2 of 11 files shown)


okaneco:
This saves users from having to implement it themselves and is useful for generating color palettes with acceptable contrast between text and background colors. The calculation is easy because of `palette` but can be improperly implemented if Luma is used for contrast calculation instead of Luma\<Linear\>.

palette/src/luma/relative_contrast.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// Higher contrast ratio is generally desireable.
+    pub trait RelativeContrast {
+        /// The return type for the contrast ratio.
+        type Scalar: Component;
+
+        /// Calculate contrast ratio between two colors.
+        fn get_contrast_ratio(&self, other: &Self) -> Self::Scalar;
+        /// Verify the contrast between two colors satisfies SC 1.4.3. Contrast
+        /// is at least 4.5:1 (Level AA).
+        fn is_min_contrast(&self, other: &Self) -> bool;
```
Ogeon: I think this and the ones below can have default implementations. They are going to be the same, regardless of the color representation, right?

palette/src/luma/relative_contrast.rs
```diff
@@ -0,0 +1,34 @@
+pub mod wcag {
```
Ogeon: Not sure if the double wrapping is necessary, unless you have plans for suggesting more ways of implementing contrast ratio. :slightly_smiling_face:

palette/src/luma/luma.rs
```diff
@@ -725,10 +726,44 @@ where
     }
 }
 
+impl<S, T> RelativeContrast for Luma<S, T>
+where
+    T: FloatComponent,
+    S: LumaStandard,
```
Ogeon: An alternative could be to force this to be linear. But I'm not sure that's the way to go. Maybe it's better to go the opposite way this time and implement this trait for anything that can be converted to linear `Luma`. I think a blanket implementation would be possible.

Ogeon:
There's apparently some subtleties when it comes to non-text elements: https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html
May be good to somehow clarify that it's for text and also maybe add the 3:1 ratio for other graphical elements.

palette/src/luma/relative_contrast.rs
okaneco: I did this preemptively to explore other contrast types later. After you pointed it out and I looked it up, I realized there weren't really other ways.

okaneco:
I added Success Criterion 1.4.11 Non-text Contrast, tried to clarify the methods were for text and non-text, and implemented the trait on types that can convert to linear luma. 

For the original commit, I imagined the primary users would be people using RGB/HSV/HSL/HWB for picking out colors, but it can definitely be convenient for other types too. My concern was that in some color models the contrast ratio calculated isn't guaranteed to be the same when the colors are clipped during conversion to sRGB. I'm not sure how to stop people from assuming contrast will remain the same or if it's that much of a problem.

---

For the `cargo fmt` run, `cargo +nightly fmt` is needed in order to get comment wrapping as specified in the `rustfmt.toml`. I believe there are 14 files that need formatting still.

palette/src/luma/relative_contrast.rs
Ogeon: I think we can solve that problem later if it appears.

Ogeon:
These criteria are more guidelines and recommendations than hard rules, as I understand it, so one will have to use manual tests either way. I suppose all we can do there is to recommend that it's used on a color space that is the intended output space. So even if the input is HSV (still in the RGB family), the conversion to RGB is going to be lossy in some way, so it's probably best to convert to RGB before checking the contrast. This does also become an ergonomics v.s. correctness issue that's connected to the whole conversion system. The most correct way would perhaps be to always return a `Result` when converting, but that's not the convenient way. It's not exactly a straight forward choice.

The fact that we can't know the output format for sure (we can only guess that it's probably sRGB, using statistics) leaves us with very few options. I think restricting it to only be available for `Luma` will only add friction. What it could do, at best, is to make people aware that some conversion is required, but every color space outside the RGB family can be converted to `Luma` without going through RGB. That, together with the fact that we don't know what people will do with their colors, makes me lean towards a more enabling than restricting approach. That's why I suggested a blanket implementation like `impl<T: IntoColor> RelativeContrast for T`, to make it included "for free".

The rest is a matter of education. The trait documentation can have examples that clearly demonstrate the best practices. It can use variable naming to emphasize that it's the output color, for example, and the example scenario can put it in a context where it's even more clear.

> For the cargo fmt run, cargo +nightly fmt is needed in order to get comment wrapping as specified in the rustfmt.toml. I believe there are 14 files that need formatting still.

I forgot that there may be problems if someone is developing on the nightly channel, although that's not recommended for compatibility reasons. That config file doesn't do much good really...

palette/src/luma/relative_contrast.rs
```diff
@@ -0,0 +1,57 @@
...
+        self.get_contrast_ratio(other) >= from_f64(7.0)
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.6 for large
+    /// text. Contrast is at least 4.5:1 (Level AAA).
+    fn is_enhanced_contrast_large_text(&self, other: &Self) -> bool {
+        self.is_min_contrast_text(other)
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.11 for graphical
+    /// objects. Contrast is at least 3:1 (Level AA).
+    fn is_non_text_contrast(&self, other: &Self) -> bool {
```
Ogeon: Or `is_min_contrast_non_text`? Or just `is_min_contrast`? Maybe `is_min_contrast_graphics`?

Besides, isn't it more correct to say that the color _has_ some contrast? As in `has_min_contrast(_for)_text`. I'm not a native speaker, so I may definitely be wrong.

palette/src/luma/relative_contrast.rs
```diff
@@ -0,0 +1,57 @@
...
+/// For more details of the guidelines:
+///
+/// [Success Criterion 1.4.3 Contrast (Minimum) (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum)
+///
+/// [Success Criterion 1.4.6 Contrast (Enhanced) (Level AAA)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced)
+///
+/// [Success Criterion 1.4.11 Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)
+pub trait RelativeContrast {
+    /// The return type for the contrast ratio.
+    type Scalar: FloatComponent;
```
Ogeon: I think we can tone this requirement down a bit. :slightly_smiling_face: All we need is `PartialOrd + FromF64`, as far as I can tell, and it doesn't _have_ to be useful as a color component. The rest is implementation details. Part of why I'm commenting on this is that I'm considering reducing the trait requirements further, if possible. `Float`, for example, implies a lot of additional mathematical properties!

palette/src/luma/relative_contrast.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+    /// Verify the contrast between two colors satisfies SC 1.4.11 for graphical
+    /// objects. Contrast is at least 3:1 (Level AA).
+    fn is_min_contrast_graphics(&self, other: &Self) -> bool {
+        self.is_min_contrast_large_text(other)
+    }
+}
+
+impl<T: IntoColor + Copy> RelativeContrast for T {
+    type Scalar = f32;
```
Ogeon: Sorry, I was a bit sloppy when I suggested the blanket implementation. You can make this generic, but I think `IntoColor` defaults to `f32`.

palette/src/luma/relative_contrast.rs
okaneco: Any of those names are fine, I actually prefer `is_min_contrast_graphics`. The heading for that recommendation is called `Success Criterion 1.4.11 Non-text Contrast`. My reasoning for not including `min_contrast` in that particular function name was because `Success Criterion 1.4.3 Contrast (Minimum)` has minimum in the name and 1.4.11 is a separate criterion which does not. There is no minimum or enhanced contrast for graphical/non-text objects, only the recommendation of 3:1, so I didn't want it to be confused with 1.4.3. I'm not attached to any of the names so changing that is fine as 1.4.11 does fall in the same Level AA of Compliance as 1.4.3 which is the "minimum" for contrast levels (but still not Level A which seems more like a fundamental level for content).

I chose `is_*_contrast*` thinking of it like "is this (compliant with) recommendation criteria". I think `has_min_contrast` is more clear about the idea that "this color has a minimum contrast with other color" so `has_min_contrast_graphics` might be the way to go. I think I should rename the other functions now that I think about it.

Ogeon: Sounds good. I thought it sounded weird to think of it as "is this color a (non-text) contrast"? Other than that, I'm trying to align them in some way. It's probably ok to go a step further and use `_min_`, because that seem to be the intention. As in "anything under this is not recommend".

palette/src/luma/relative_contrast.rs
okaneco: What would this look like? I saw that `IntoColor` defaults to `f32` but I'm not sure how to use that to satisfy the trait bounds.

Ogeon: The defaults are just what the type parameters take when they are omitted, so all you have to do is to add them to the `impl`:

```rust
impl<C, Wp, T> RelativeContrast for C
where
    C: IntoColor<Wp, T>,
    Wp: WhitePoint,
    T: FloatComponent
```

Then you will have To available for the scalar type.

Ogeon:
Interesting with API design on the fly. I hope you don't feel like I'm pushing too much. I have just been thinking a lot about how to improve the trait situation. Hence my mind dump earlier. I have some ideas regarding how things may become more convenient. Some of it involves doing more "under the hood". For example trying to combine component conversion with color space conversion, so it becomes possible to go from `Srgb<u8>` to `Lch<f32>` in a single method call. But there's still some distance to go before it's there yet...

As for this, I think a final thing that probably should be done is to move this trait out of the luma module. It's no longer a special thing for Luma, after all.

okaneco:
It's fine, no worries. For this PR and the color difference PR I wasn't sure where they exactly fit in because they extend functionality in different ways than most of the operations already in the  library. I expected a lot of reworking.

> For example trying to combine component conversion with color space conversion

That would be excellent.

Ogeon:
Good to hear I'm not overly annoying! Things like these are not always straight forward. There's always that balance I mentioned and other factors. In the end we are all trying to improve this thing as much as we can.

palette/src/relative_contrast.rs
```diff
@@ -0,0 +1,137 @@
...
+/// desireable.
+///
+/// For more details of the guidelines:
+///
+/// [Success Criterion 1.4.3 Contrast (Minimum) (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum)
+///
+/// [Success Criterion 1.4.6 Contrast (Enhanced) (Level AAA)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced)
+///
+/// [Success Criterion 1.4.11 Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)
+pub trait RelativeContrast<C, Wp, T>
```
Ogeon: None of these should be necessary here... If the blanket implementation complains, then that's the problem we have to solve.

okaneco: If I remove them then I get:

```rust
error[E0207]: the type parameter `Wp` is not constrained by the impl trait, self type, or predicates
  --> palette/src/relative_contrast.rs:64:9
   |
64 | impl<C, Wp, T> RelativeContrast for C
   |         ^^ unconstrained type parameter

error[E0207]: the type parameter `T` is not constrained by the impl trait, self type, or predicates
  --> palette/src/relative_contrast.rs:64:13
   |
64 | impl<C, Wp, T> RelativeContrast for C
   |             ^ unconstrained type parameter

error: aborting due to 2 previous errors
```

Ogeon: Oh no... That's a bummer. I can think of three options that do not involve a lot more figuring out:

1. Make `RelativeContrast` have a `<Wp, T>` parameter and remove `Scalar`. Not the best option, unless we want a many-to-many relationship between types and `RelativeContrast` implementations.
2. Give `RelativeContrast` an additional `WhitePoint` associated type and implement it for `C: IntoColor<C::WhitePoint, C::Scalar>` (I think this will work). Not fun to have a seemingly unrelated associated type. It does at least mean that there is only one implementation per scalar and white point pair.
3. Go back to implementing it for every type. Tedious but should at least work.

It may also be possible to use other traits to get the required types, but I'm not sure how to do it in a good way yet. Something like a universal `trait Color {type WhitePoint; type Component;}`, but I'm not sure if it's better or worse.

okaneco: 1. If I go for the approach of 1 with removing the `Scalar` and having the `<Wp, T>` parameter, I get this which I'm not sure how to solve.
```rust
error[E0284]: type annotations needed: cannot resolve `<f32 as std::ops::Add<_>>::Output == _`
  --> palette/src/relative_contrast.rs:78:24
   |
78 |             (this.luma + from_f64(0.05)) / (other.luma + from_f64(0.05))
   |                        ^
```

2. I'm not sure how to make 2 work.
```rust
error[E0391]: cycle detected when computing the bounds for type parameter `C`
  --> palette/src/relative_contrast.rs:66:18
   |
66 |     C: IntoColor<C::WhitePoint, C::Scalar> + Copy,
   |                  ^^^^^^^^^^^^^
   |
   = note: ...which again requires computing the bounds for type parameter `C`, completing the cycle
note: cycle used when processing `relative_contrast::<impl at palette/src/relative_contrast.rs:64:1: 83:2>`
```

I can implement for every type, I was hoping not to.

Ogeon: It's hard to tell without seeing the trait bounds, but the first one seem really odd. It should be the same you have, but without the `C` parameter on `RelativeContrast`. An extended variant of that is to have the trait require `IntoColor<Wp, T>` and be all default implementations. But that's kind of getting a bit weird.

I'm starting to think that implementing it for each of them is the way to go. It's tedious, and I know you already almost did before, but it's simple and it's going to be the approach that requires the least restrictions as it is today. The trait doesn't have to be aware of anything more than its output type.

What do you think?

okaneco: Sorry, I should have included gists of what I tried. I can't recreate the issue I had with `1.` but I came back with a clearer head to try `1.` again and it worked.

Including a gist of what I tried to `2.` https://gist.github.com/okaneco/2b9d8c9f1ddd2d59ea9a61404806cece

Ogeon: Alright, seems like it's not possible to use the associated types like that.

To add to my reasoning, one of the downsides I see with option 1 and possibly 2 is that one would have to pick a white point to use it in trait bounds.

okaneco: I agree and favor the more permissive way as it was implemented for each type before. 

I cleaned up some of the imports to be more consistent in separating the functions, structs, and traits across the files I interacted with.

okaneco:
I submitted a rough draft of documentation to educate about the usage of the trait. I just wasn't sure about what the doc test should explicitly show.

> The rest is a matter of education. The trait documentation can have examples that clearly demonstrate the best practices. It can use variable naming to emphasize that it's the output color, for example, and the example scenario can put it in a context where it's even more clear.

palette/src/relative_contrast.rs
```diff
@@ -0,0 +1,141 @@
...
+/// `L1` is the luminance of the brighter color and `L2` is the luminance of the
+/// darker color both in sRGB linear space. Higher contrast ratio is generally
+/// desireable.
+///
+/// These criteria are recommendations, not hard and fast rules. Manual tests
+/// should be used to check that contrast ratios are compliant in the output
+/// color space. For example, if the input colors are in HSV and the output
+/// colors are in RGB, the contrast should be checked again since the output
+/// color model differs. Conversion of color from HSV to RGB is going to be
+/// lossy in some way so the contrast ratio of the RGB should be re-checked.
```
Ogeon: I think this is a good summary, but I feel like I need to clarify a couple of things:

* With "manual tests" I meant actually looking at the result and verifying that it's clearly and comfortably readable, even with low vision or color deficiencies. Nothing beats that, so far.

* The "checking the contrast of the output space" part should probably clarify that it's recommended to basically run `my_background_rgb.has_enhanced_contrast_text(&my_foreground_rgb)` before outputting them, if RGB is the output format (and it usually is).

I would suggest editing it to clarify that and move it up above the more technical part. I think we will have more users than people implementing it. Then I think the example can also be simplified a lot to something that only shows checking for enhanced contrast of a single color pair. Nothing prevents us from having examples for each method too, but I think it's best to have each example as small and on point as possible. It's also more aligned with the rest of the examples if you try to use the types from the crate root, such as `palette::Srgb`. It's the same as `Rgb::<Srgb, _>`, but with less syntax.

Ogeon:
Or all the other color types could just call the method on Luma. Converting to it is always going to be as cheap as possible, anyway. 🤷🏻‍♂️

Ogeon:
Nice! I think that gets the message across. Let me know when you feel like it's ready to be merged.

okaneco:
Twists and turns indeed. 

I liked the idea of moving the calculation into the `relative_contrast` module which simplifies all the calls and should help make it more maintainable. It is quite obvious in retrospect but I completely missed it as well while trying to handle the blanket impl.

I've fixed some minor grammar issues I just noticed but I believe this is finished and I can squash after your approval.

Ogeon:
Looks great! Thanks!

There was a new Rust release just yesterday, so there may have been a regression there. Or just high load on their servers. I'll have a look at the numbers in a moment and merge after that.

Ogeon:
The Mac runners can be slow sometimes and this seem to be within the ballpark, so it may have been just bad luck. But we'll have to see if it stays consistent.

Either way, time to merge!

bors r+

Ogeon:
Seems like Bors isn't feeling well today.

Ogeon:
Let's see if this works now.

bors r+

PR #166, opened 2020-01-31
Migrate to Rust 2018 edition

okaneco:
Update `palette` to 2018 edition
Update `palette_derive` and `no_std_test` to 2018 edition
Apply selected cargo clippy lints
- `#[cfg_attr(rustfmt, rustfmt_skip)]` to `#[rustfmt::skip]`
- remove redundant names in struct assignment for variables with same name: `value: value,` becomes `value,`

Remove `extern crate` when applicable
Replace `#[macro_use]` with `use {crate}::{macro}`
Remove needless `fn main()` in doctests
Fix spacing and wrapping in comments
Change 4 expects to unwrap_or_else in build/named.rs
Remove rustfmt.toml until rustfmt 2.0 stabilizes wrapping comments

### DIFF
--- a/palette/src/alpha.rs
+++ b/palette/src/alpha.rs
@@ -1,25 +1,26 @@
 use core::fmt;
 use core::ops::{Add, AddAssign, Deref, DerefMut, Div, DivAssign, Mul, MulAssign, Sub, SubAssign};
 
-use float::Float;
-
 use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 
-use blend::PreAlpha;
-use encoding::pixel::RawPixel;
-use {clamp, Blend, Component, ComponentWise, GetHue, Hue, Limited, Mix, Pixel, Saturate, Shade};
+use crate::blend::PreAlpha;
+use crate::encoding::pixel::RawPixel;
+use crate::float::Float;
+use crate::{
+    clamp, Blend, Component, ComponentWise, GetHue, Hue, Limited, Mix, Pixel, Saturate, Shade,
+};
 
-///An alpha component wrapper for colors.
+/// An alpha component wrapper for colors.
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[cfg_attr(feature = "serializing", derive(Serialize, Deserialize))]
 #[repr(C)]
 pub struct Alpha<C, T> {
-    ///The color.
+    /// The color.
     #[cfg_attr(feature = "serializing", serde(flatten))]
     pub color: C,
 
-    ///The transparency component. 0.0 is fully transparent and 1.0 is fully
-    ///opaque.
+    /// The transparency component. 0.0 is fully transparent and 1.0 is fully
+    /// opaque.
     pub alpha: T,
 }
 
@@ -382,7 +383,7 @@ where
 impl<C, T: Component> From<C> for Alpha<C, T> {
     fn from(color: C) -> Alpha<C, T> {
         Alpha {
-            color: color,
+            color,
             alpha: T::max_intensity(),
         }
     }
@@ -424,8 +425,8 @@ where
 
 #[cfg(test)]
 mod test {
-    use encoding::Srgb;
-    use rgb::Rgba;
+    use crate::encoding::Srgb;
+    use crate::rgb::Rgba;
 
     #[test]
     fn lower_hex() {
--- a/palette/src/blend/blend.rs
+++ b/palette/src/blend/blend.rs
@@ -1,53 +1,53 @@
-use float::Float;
 use num_traits::{One, Zero};
 
-use blend::{BlendFunction, PreAlpha};
-use {clamp, ComponentWise};
+use crate::blend::{BlendFunction, PreAlpha};
+use crate::float::Float;
+use crate::{clamp, ComponentWise};
 
-///A trait for colors that can be blended together.
+/// A trait for colors that can be blended together.
 ///
-///Blending can either be performed through the predefined blend modes, or a
-///custom blend functions.
+/// Blending can either be performed through the predefined blend modes, or a
+/// custom blend functions.
 ///
-///_Note: The default implementations of the blend modes are meant for color
-///components in the range [0.0, 1.0] and may otherwise produce strange
-///results._
+/// _Note: The default implementations of the blend modes are meant for color
+/// components in the range [0.0, 1.0] and may otherwise produce strange
+/// results._
 pub trait Blend: Sized
 where
     <Self::Color as ComponentWise>::Scalar: Float,
 {
-    ///The core color type. Typically `Self` for color types without alpha.
+    /// The core color type. Typically `Self` for color types without alpha.
     type Color: Blend<Color = Self::Color> + ComponentWise;
 
-    ///Convert the color to premultiplied alpha.
+    /// Convert the color to premultiplied alpha.
     fn into_premultiplied(self) -> PreAlpha<Self::Color, <Self::Color as ComponentWise>::Scalar>;
 
-    ///Convert the color from premultiplied alpha.
+    /// Convert the color from premultiplied alpha.
     fn from_premultiplied(
         color: PreAlpha<Self::Color, <Self::Color as ComponentWise>::Scalar>,
     ) -> Self;
 
-    ///Blend self, as the source color, with `destination`, using
-    ///`blend_function`. Anything that implements `BlendFunction` is
-    ///acceptable, including functions and closures.
+    /// Blend self, as the source color, with `destination`, using
+    /// `blend_function`. Anything that implements `BlendFunction` is
+    /// acceptable, including functions and closures.
     ///
-    ///```
-    ///use palette::{LinSrgb, LinSrgba, Blend};
-    ///use palette::blend::PreAlpha;
+    /// ```
+    /// use palette::{LinSrgb, LinSrgba, Blend};
+    /// use palette::blend::PreAlpha;
     ///
-    ///type PreRgba = PreAlpha<LinSrgb<f32>, f32>;
+    /// type PreRgba = PreAlpha<LinSrgb<f32>, f32>;
     ///
-    ///fn blend_mode(a: PreRgba, b: PreRgba) -> PreRgba {
+    /// fn blend_mode(a: PreRgba, b: PreRgba) -> PreRgba {
     ///    PreAlpha {
     ///        color: LinSrgb::new(a.red * b.green, a.green * b.blue, a.blue * b.red),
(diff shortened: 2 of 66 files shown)


okaneco:
After you mentioned migration to 2018, it caught my interest to see what work was involved. It's not a simple `cargo fix --edition` and required some hands on conversion for resolving errors/warnings. I kept the commits split in two because I wasn't sure if you wanted to touch the other crates or had more changes in mind you wanted to handle yourself. I used Rust some before 2018 edition but I'm not sure of all the changes 2018 brings aside from some of the module handling and keyword specifics. I remember NLL being a big deal but it's hard for me to tell what was and wasn't backported to 2015 edition.

https://doc.rust-lang.org/edition-guide/rust-2018/edition-changes.html

palette/src/alpha.rs
```diff
@@ -394,7 +396,7 @@ where
     C: fmt::LowerHex,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let size = f.width().unwrap_or(::core::mem::size_of::<T>() * 2);
+        let size = f.width().unwrap_or_else(|| ::core::mem::size_of::<T>() * 2);
```
Ogeon: How come this was changed? I would expect `size_of` to be constant and not need a function wrapper.

palette/src/convert.rs
```diff
@@ -1,9 +1,9 @@
+use crate::encoding::Linear;
+use crate::luma::Luma;
+use crate::rgb::{Rgb, RgbSpace};
+use crate::white_point::{WhitePoint, D65};
+use crate::{FloatComponent, Hsl, Hsv, Hwb, Lab, Lch, Limited, Xyz, Yxy};
```
Ogeon: My module grouping convention is

```rust
// Standard libs first
use core::...;
use std::...;

// Then third party stuff in one or more groups
use some_lib::...;

// And finally internal imports
use crate::...;
```

palette/src/lch.rs
```diff
@@ -174,7 +174,7 @@ where
         Lch {
             l: color.l,
             chroma: (color.a * color.a + color.b * color.b).sqrt(),
-            hue: color.get_hue().unwrap_or(LabHue::from(T::zero())),
+            hue: color.get_hue().unwrap_or_else(|| LabHue::from(T::zero())),
```
Ogeon: Not sure this is necessary either.

palette/src/convert.rs
okaneco: That was `cargo fmt` since the lines were touching. Fixing that now.

Ogeon: I suspected that was the case. 🙂 I forgot to elaborate, but I just meant to show how I usually do it. Formatting that isn't enforced programmatically isn't a big deal.

okaneco:
Yes, the `unwrap_or_else` were clippy lints. They suggest that change for `unwrap_or` which have a function call that could be labor-intensive but you're right that they're probably constant and should be fine. I don't have any benchmarks or other justification other than to hear your thoughts about it. I've tried looking up performance writeups but haven't found much on the topic other than people recommending it. I know in the source it doesn't call the function unless it has to and there's not much going on with the function calls involved. I know clippy is just a recommendation and sometimes the recommendations aren't positives.

```rust
warning: use of `unwrap_or` followed by a function call
   --> palette\src\lch.rs:177:34
    |
177 |             hue: color.get_hue().unwrap_or(LabHue::from(T::zero())),
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| LabHue::from(T::zero()))`
    |
    = note: `#[warn(clippy::or_fun_call)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
```
In other cases of this lint, if you run `cargo clippy` in `palette/palette` you'll get it 4 times for `palette/build/named.rs` in an instance which I hadn't seen before. This seems more like the type of place it'd make an impact if this were especially performance critical.

```rust
warning: use of `expect` followed by a function call
  --> palette\build\named.rs:22:14
   |
22 |             .expect(&format!("couldn't get color for {}", name))
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| panic!("couldn't get color for {}", name))`
   |
   = note: `#[warn(clippy::expect_fun_call)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
```
 ---
The main lints remaining are for "unreadable literals" asking for separators like 000_000 between numbers and all the `fn main()`s that used to be needed for the doc tests. The only other one that possibly seems actionable are these two which just require documentation. 

```rust
warning: unsafe function's docs miss `# Safety` section
  --> palette\src\encoding\pixel\raw.rs:35:5
   |
35 |     unsafe fn from_raw_parts<'a>(pointer: *const T, length: usize) -> &'a Self;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::missing_safety_doc)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc

warning: unsafe function's docs miss `# Safety` section
  --> palette\src\encoding\pixel\raw.rs:38:5
   |
38 |     unsafe fn from_raw_parts_mut<'a>(pointer: *mut T, length: usize) -> &'a mut Self;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```
---
I believe I've addressed the major issues. Only thing left is that I can run the `cargo +nightly fmt` to wrap the comments at this point but as you've stated before it's a somewhat Sisyphean endeavor.

Ogeon:
Thanks for explaining. Thinking about it, the compiler may be able to optimize the wrapper functions away and produce the same result, but that's a guess. It's hard to say much without benchmarks or inspecting the assembly output, which I haven't.

The other improvements you pointed out seem good to address, on the other hand.

And feel free to remove the config file for rustfmt if you want to. It's only making things more inconsistent at the moment. 🙂

okaneco:
I'll probably use some variation of that `rustfmt.toml` in my own projects as it's nice. But as you said it causes more annoyance than benefit at this time so I've removed it for now. I did some research and unstable features don't seem like they'll be stabilized until `rustfmt 2.0` which seems a ways off. https://github.com/rust-lang/rustfmt/projects/2

I was able to remove all the `fn main()` from doctests and went through with the changes in `named.rs` since I assumed you meant that was a good change, if not I can revert.

palette/src/blend/blend.rs
```diff
@@ -32,22 +32,22 @@ where
     ///acceptable, including functions and closures.
     ///
     ///```
-    ///use palette::{LinSrgb, LinSrgba, Blend};
-    ///use palette::blend::PreAlpha;
+    /// use palette::{LinSrgb, LinSrgba, Blend};
```
Ogeon: Shouldn't the space at the beginning be applied for the whole comment block? That's how it is for newer files. I see you made similar changes elsewhere.

palette/src/rgb/mod.rs
```diff
@@ -2,30 +2,30 @@
 
 use core::any::Any;
 
-use white_point::WhitePoint;
-use {Component, FloatComponent, FromComponent, Yxy};
-
-use encoding::{Linear, TransferFn};
+use crate::encoding;
+use crate::encoding::{Gamma, Linear, TransferFn};
```
Ogeon: Just FYI, this is possible to write as just `use crate::encoding::{self, Gamma, Linear, TransferFn};` if one would want to have fewer lines. :slightly_smiling_face: It's perfectly fine like this too.

Ogeon:
A well made migration with both cream and cherry on top! Judging by your research, it looks like removing the config is a good idea. I think the defaults are otherwise good, so it's no longer useful. It served its purpose back when it didn't require the nightly version.

I don't have anything more I feel need to be done here, other than looking into those spaces in the comments. I don't know how the markdown parser will react to those...

Did you check if it's possible to remove most of the remaining `extern crate`s by importing their macros? I know some used to call other macros, which made it much harder, but I don't know if that has changed. You probably tried, but I'm asking in case it was forgotten. :slightly_smiling_face:

okaneco:
For the remaining `extern crate`s, there's a note about some sysroot crates being exceptions to the rules. It specifically called out `proc_macro` as a crate so I left that and realized that's why `extern crate libc` is needed in `no_std_test`. I left anything behind a feature gate and I don't think the `#[macro_use] extern crate palette_derive` can be removed easily.  
https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#an-exception   
https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html

For formatting, rustfmt doesn't seem that concerned about every comment having whitespace. I checked the files with `cargo doc --open` and it looked fine so I think the markdown parser can handle it. I believe this is an issue with rustfmt currently in that it's not as consistent as it could be.

----
As a last bit of enhancement, I saw the keywords in the cargo manifest and did some research on crates.io. According to the documentation you can have up to 5 keywords.
These are the current keywords
```toml
keywords = ["color", "colour", "space", "linear"]
```
That said, searching isn't great on the site with how it separates exact matches, keywords, and categories. I've not had great experiences trying to find things on there in the past. Searching "palette" in the keywords, it doesn't come up but palette_derive does. 
https://crates.io/keywords/palette
Some things like https://crates.io/keywords/space don't seem to be a good use of keyword. I think it pulls some from the readme since searching for srgb, palette and *_derive came up 2nd and 3rd.

It might make more sense to include a sentence of all the color models that palette works with in the description so it'll get hits in the search.

Ogeon:
Yes, some `extern crate`s are still necessary, but that's fine. It has come as close as we can get it now, I think, so I'm ready to merge it unless you want to change anything else.

Regarding the keywords, I'm open for suggestions. I'm pretty bad at making them up. Although, I think that deserves its own separate discussion. Let's not cram too much into this. :smile:

okaneco:
Yes, I think this is a good stopping point. 

I can start an issue for that, I agree.

Ogeon:
Thanks. I think some kind of squashing would be good, though.

okaneco:
I was in the process of that. :+1: 

I just wanted to make sure the Rust nightly passed.

Ogeon:
Nice, thanks a lot!

bors r+

Issue #167, opened 2020-02-01
Crates.io search improvements


okaneco:
This came up while working on #166 

These are the current keywords in the cargo manifest:
```toml
keywords = ["color", "colour", "space", "linear"]
```

### Goal

Improve search of `palette` on crates.io

### Issue

According to the documentation you can have up to 5 keywords. That said, crate discovery isn't great on the site with how it handles searching and separating exact matches, keywords, and categories. I've not had pleasant experiences trying to find things on there in the past or at least some type of "canonical" crate for a task.

Searching "palette" in the keywords, it doesn't come up but palette_derive does so that might be a good keyword to add: https://crates.io/keywords/palette. Others like like https://crates.io/keywords/space don't initially seem to be a good use of keyword because of how "dumb" the search is, however "colo[u]r space" brings up the crate as a top hit so keywords combine in your favor. I think it pulls some from the readme since searching for "srgb", palette and *_derive came up 2nd and 3rd, along with other crates being surfaced in similar color space/model searches.

### Suggestions

Add at least one more keyword. I think "**pixel**" may be the best candidate over "palette", since the crate doesn't show up while searching it https://crates.io/search?q=pixel and that's very relevant to what a lot of us want to do. Ideally search wouldn't differentiate color/colour so "colour" is probably still worth a keyword instead of replacing it with "palette".

It might make more sense to include a sentence of all the color models that `palette` works with and a sentence of some features people might be looking for in the description so it'll get hits in the search. An example being if you search "hwb", palette doesn't appear in the results. `palette` generally does well showing up at the top for queries like "srgb". Whatever you think the most important features are that can fit in 2-3 sentences probably makes the most sense. In this research, I looked up "lab" and `palette` doesn't surface however two CIEDE2000 color difference crates do that I didn't know about (funny enough, they avoid the test case that gave us errors until we switched to f64).

These are some of the words that I think should be in the description based on lack of presence in searches:
- Alpha composition/alpha compositing
- Luma, luminance
- color difference, contrast ratio
- HSL, HSV, HWB, Lab, Lch (HCL), RGB, Xyz, Yxy

### Motivation
[...]

PR #169, opened 2020-02-18
Update cargo manifest keywords, categories

okaneco:
Change `cargo.toml` keywords: `keywords = ["color", "conversion", "linear", "pixel", "rgb"]`
Add cateogries to `cargo.toml`: `categories = ["graphics", "multimedia::images", "no-std"]`
Reduce dev-dependencies features for crates used in `palette` examples, results in reduction from 131 to 82 crates built for `cargo test --features strict`
Add docs.rs badge to README.md

This addresses some parts of #167.

### DIFF
--- a/palette/Cargo.toml
+++ b/palette/Cargo.toml
@@ -7,9 +7,10 @@ description = "Makes linear color calculations and conversion easy and accessibl
 documentation = "https://docs.rs/palette/0.5.0/palette/"
 repository = "https://github.com/Ogeon/palette"
 readme = "README.md"
-keywords = ["color", "colour", "space", "linear"]
+keywords = ["color", "conversion", "linear", "pixel", "rgb"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
+categories = ["graphics", "multimedia::images", "no-std"]
 
 build = "build/main.rs"
 
@@ -39,13 +40,19 @@ features = ["serde_derive"]
 optional = true
 
 [dev-dependencies]
-image = "0.22"
-clap = "2"
 csv = "1"
+lazy_static = "1"
 serde = "1"
 serde_derive = "1"
-lazy_static = "1"
 serde_json = "1"
+
+[dev-dependencies.clap]
+version = "2"
+default-features = false
+
+[dev-dependencies.image]
+version = "0.22"
+default-features = false
 
 [build-dependencies.phf_codegen]
 version = "0.8"
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Build Status](https://travis-ci.org/Ogeon/palette.svg?branch=master)](https://travis-ci.org/Ogeon/palette)
 [![Crates.io](https://img.shields.io/crates/v/palette.svg)](https://crates.io/crates/palette/)
+[![Docs.rs](https://docs.rs/palette/badge.svg)](https://docs.rs/palette)
 
 A Rust library that makes linear color calculations and conversion easy and accessible for anyone. It provides both precision tools that lets you work in exactly the color space you want to, as well as a general color type that abstracts away some of the technical details.
 
--- a/palette_derive/Cargo.toml
+++ b/palette_derive/Cargo.toml
@@ -20,6 +20,5 @@ quote = "^1.0"
 proc-macro2 = "^1.0"
 
 [features]
-
 #internal
 strict = []


okaneco:
I believe the remaining parts to improve are:
- updating the description in the manifest, if needed
- adding a feature description to the README, which is indexed by crates.io
- resolving whether there should be separate readmes for repo root/crates.io

I wasn't sure if any of those should be part of this PR.

----
For changing the dev-dependencies, I didn't see a reason to keep building the extra 49 crates for the examples. That time adds up, especially on a slower computer I use sometimes to work on `palette`, if I ran cargo clean or used a new nightly. I don't think much more can be trimmed or would be worth changing.

palette/Cargo.toml
```diff
@@ -7,12 +7,16 @@ description = "Makes linear color calculations and conversion easy and accessibl
...
-keywords = ["color", "colour", "space", "linear"]
+keywords = ["color", "conversion", "linear", "pixel", "rgb"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
+categories = ["graphics", "multimedia::images", "no-std"]
 
 build = "build/main.rs"
 
+[badges]
+travis-ci = { repository = "Ogeon/palette" }
```
Ogeon: Considering it doesn't show the CI state of the released code (which should always pass), I'm not sure how useful this badge is, to be honest. I have always wondered. I find most status badges to be odd, really, for this reason. The maintenance badge has the reverse problem of not being live, while the true maintenance status of the lib may change from one day to another. :man_shrugging: Not sure if I'm misunderstanding something, though...

palette/Cargo.toml
```diff
@@ -39,14 +43,20 @@ features = ["serde_derive"]
...
 serde_json = "1"
 
+[dev-dependencies.clap]
+version = "2"
+default-features = false
+
+[dev-dependencies.image]
+version = "0.22"
+default-features = false
+
```
Ogeon: Good call! It's easy to forget that there may be features to disable. This may be something to look into for the rest of the dependencies, as I haven't done that in quite a while.

palette/Cargo.toml
okaneco: I agree about the badges and find them mostly to be noise. Good point about not showing the released version's build, which should always pass, so this shouldn't be necessary. I don't think much of a passing badge but a failing badge can be concerning even if it's not reason to be. 

I can strike this out. This was only included because some other projects use it and makes it show up as a badge on the search next to the version number.

palette/Cargo.toml
okaneco: ### [palette](https://github.com/Ogeon/palette/blob/master/palette/Cargo.toml#L16) cargo.toml
[image](https://github.com/image-rs/image/blob/master/Cargo.toml) was definitely the heaviest: 
```toml
[features]
default = ["gif_codec", "jpeg", "ico", "png_codec", "pnm", "tga", "tiff", "webp", "bmp", "hdr", "dxt", "jpeg_rayon"]
```
[clap](https://github.com/clap-rs/clap/blob/v2-master/Cargo.toml) not as much since I realize master is on the v3 beta, this is v2.33
```toml
[features]
default = ["suggestions", "color", "vec_map"]
```
[serde](https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml) and [serde_json](https://github.com/serde-rs/json/blob/master/Cargo.toml) not much to change since this is also feature gated on `palette`.
```toml
[features]
default = ["std"]
```
[phf](https://github.com/sfackler/rust-phf/blob/master/phf/Cargo.toml), similar to above
```toml
[features]
default = ["std"]
std = ["phf_shared/std"]
```
No default:
[lazy_static](https://github.com/rust-lang-nursery/lazy-static.rs/blob/master/Cargo.toml) 
[csv](https://github.com/BurntSushi/rust-csv/blob/master/Cargo.toml)
[serde_derive](https://github.com/serde-rs/serde/blob/master/serde_derive/Cargo.toml)
[phf_codegen](https://github.com/sfackler/rust-phf/blob/master/phf_codegen/Cargo.toml)

### [palette_derive](https://github.com/Ogeon/palette/blob/master/palette_derive/Cargo.toml#L17) cargo.toml
```toml
[dependencies]
syn = { version = "^1.0", features = ["extra-traits"] }
quote = "^1.0"
proc-macro2 = "^1.0"
```

[syn](https://github.com/dtolnay/syn/blob/master/Cargo.toml)
```toml
[features]
default = ["derive", "parsing", "printing", "clone-impls", "proc-macro"]
```
[quote](https://github.com/dtolnay/quote/blob/master/Cargo.toml)
```toml
[features]
default = ["proc-macro"]
# Disabling the proc-macro feature removes the dynamic library dependency on
# libproc_macro in the rustc compiler.
proc-macro = ["proc-macro2/proc-macro"]
```
[proc-macro2](https://github.com/alexcrichton/proc-macro2/blob/master/Cargo.toml)
```toml
[features]
proc-macro = []
default = ["proc-macro"]
```

----
I can remove `syn`'s `"printing",  "clone-impls"` features which don't reduce crate count but might improve compile time.

Ogeon: Nice, seems like it's in a good shape now, after your change. The syn changes have to be measured to know the impact, but I guess it doesn't hurt to remove them if they are not used.

palette/Cargo.toml
Ogeon: They can be fun, but I would rather skip the badges for now, unless they are directly related to the published versions. :slightly_smiling_face: 

The ones in the readme are a whole other business, but there I'm leaning more and more towards the idea of having different readmes for the repo and published crate. That's a separate topic.

Ogeon:
My speed test experiment didn't show much, by the way. (36 vs 32 minutes) But I don't think those particular dependencies are rebuilt a lot.

palette/Cargo.toml
okaneco: Looks like I'll have to revert the syn changes, when I tried building locally I don't think it built cleanly.

okaneco:
Yes, I believe they get used as cached results so it won't be that large of an effect and I'm not sure how much of that is in the CI runner noise.

palette/Cargo.toml
Ogeon: Ah, oh well. I think it's good anyway. I know there are other tricks, like not optimizing particular dependencies, but I don't think we need to go down that rabbit hole in this PR.

Ogeon:
Looking at the history, I think one can expect it to vary with as much as 3 minutes, if not more. And it makes sense that they would only be built once, since they are not toggled by features. It's not a good benchmark in that sense. I just noticed the built time was much shorter than the previous PR and wanted to know why, but they more likely just fixed the regressions from before.

Ogeon:
Oh, I looked at the wrong time again. They show the time without the waiting in a different place and they were even closer, but still, not a good benchmark for this.

okaneco:
Yes, I'm not sure of how to benchmark it properly, aside from seeing how many crates are being built on a clean build.

I know you can check https://perf.rust-lang.org/ which can help correlate past build runs to performance at the time. I believe there was a change within the past week or two which mostly affected `clap` instruction counts with a 2-5% reduction in compile time. You can see if you ctrl-F clap.

Ogeon:
Yes, looks like something improved there, and that would have been between now and the previous PR. No matter what Travis machines show, I think it's a good change. I'll merge it and we can take the rest separately.

bors r+

PR #170, opened 2020-02-22
Add `{into,from}_u32` methods for RGB/A, Packed struct for u32 representations of RGBA

### LINKED ISSUE #144: Packed formats
It's common to store sRGB values packed in a `u32` (eg. 0xAARRGGBB) - it would be helpful to provide functions to convert to this format.

I need to convert from linear, floating-point RGB to a packed sRGB u32, and the way I'm currently doing it is like this:
```rust
fn to_srgb(color: Vector3<f64>) -> u32 {
    let src = LinSrgb::new(color[0], color[1], color[2]);
    let dst: Srgb<u8> = Srgb::from_format(Srgb::from_linear(src));
    0xFF000000 | ((dst.red as u32) << 16) | ((dst.green as u32) << 8) | (dst.blue as u32)
}
```

I'm not sure if this is the best way to do it.

okaneco:
Add `into_u32` and `from_u32` methods for `Rgb` and `Rgba`.
Add `Packed` struct which stores Rgb/a colors as a packed u32 with support for multiple channel orders: ABGR, ARGB, BGRA, and RGBA.
Add `RgbChannels` trait for splitting and combining of Rgba components in the previously mentioned channel orderings.
Add corresponding `From` impls for `Packed`, `Rgb/a`, `u32`

Closes #144

### DIFF
--- a/palette/src/rgb/packed.rs
+++ b/palette/src/rgb/packed.rs
@@ -1,0 +1,339 @@
+pub mod channels;
+
+use core::marker::PhantomData;
+
+use crate::rgb::{Rgb, RgbStandard, Rgba};
+use crate::Pixel;
+
+/// RGBA color packed into a 32-bit unsigned integer. Defaults to ARGB
+/// ordering for `Rgb` types and RGBA ordering for `Rgba` types.
+///
+/// Packed integer type represented in `u32`. Two hexadecimal digits (8-bits)
+/// express each value of the Red, Green, Blue, and Alpha components in the
+/// RGBA color.
+///
+/// Note that conversion from float to integer component types in `palette` maps
+/// the floating point value to the integer's value range and then rounds the
+/// result before casting to the integer type. An example of the difference in
+/// conversion methods is shown below.
+///
+/// ```
+/// use approx::assert_relative_eq;
+/// use palette::{Packed, Srgb, Srgba};
+/// use palette::rgb::channels::{Argb, Rgba};
+///
+/// let packed: Packed = Srgb::new(0.5, 0.0, 0.5).into_format().into();
+/// assert_eq!(0xFF80_0080, packed.color);
+///
+/// let unpacked: Srgba<u8> = Packed::<Rgba>::from(0x80FF_FF80).into();
+/// assert_relative_eq!(
+///     Srgba::new(0.5, 1.0, 1.0, 0.5),
+///     unpacked.into_format(),
+///     epsilon = 0.01
+/// );
+///
+/// // By default, `Packed` uses `Argb` order for creating `Rgb` colors to make
+/// // entering 6-digit hex numbers more convenient
+/// let rgb = Srgb::from(0xFF8000);
+/// assert_eq!(Srgb::new(0xFF, 0x80, 0x00), rgb);
+///
+/// let rgba = Srgba::from(0xFF80007F);
+/// assert_eq!(Srgba::new(0xFF, 0x80, 0x00, 0x7F), rgba);
+///
+/// // The second assert is essentially how library components are converted
+/// let float = 0.5f32 * 255.0;
+/// assert_eq!(0x7F, float as u32);
+/// assert_eq!(0x80, float.round() as u32);
+/// ```
+///
+/// When an `Rgb` type is packed, the alpha value will be `0xFF` in the
+/// corresponding `u32`. Converting from a packed color type back to an `Rgb`
+/// type will disregard the alpha value.
+///
+/// `Packed` implements [Pixel](encoding/pixel/trait.Pixel.html) and can be
+/// constructed from a slice of `&[u32]`.
+///
+/// ```
+/// use palette::{Packed, Pixel};
+/// use palette::rgb::channels::Argb;
+///
+/// let raw = &[0x7F0080u32, 0x60BBCC];
+/// let colors = Packed::<Argb>::from_raw_slice(raw);
+///
+/// assert_eq!(colors.len(), 2);
+/// assert_eq!(colors[0].color, 0x7F0080);
+/// assert_eq!(colors[1].color, 0x60BBCC);
+/// ```
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Pixel)]
+#[palette_internal]
+#[repr(C)]
+pub struct Packed<C: RgbChannels = channels::Argb> {
+    /// The sRGB color packed into a `u32`.
+    pub color: u32,
+
+    /// The channel ordering for red, green, blue, and alpha components in the
+    /// packed integer; can be `Abgr`, `Argb`, `Bgra`, or `Rgba`. See
+    /// [RgbChannels](trait.RgbChannels.html).
+    #[palette_unsafe_zero_sized]
+    pub channel_order: PhantomData<C>,
+}
+
+/// Splits and combines RGB(A) types with some channel ordering. Channels may be
+/// ordered as `Abgr`, `Argb`, `Bgra`, or `Rgba`.
+pub trait RgbChannels {
+    /// Split RGBA components into a `(u8, u8, u8, u8)` tuple.
+    fn split_rgb<S: RgbStandard>(rgb: Rgba<S, u8>) -> (u8, u8, u8, u8);
+    /// Create an RGBA color from a `(u8, u8, u8, u8)` tuple.
+    fn combine_rgb<S: RgbStandard>(channels: (u8, u8, u8, u8)) -> Rgba<S, u8>;
+}
+
+impl<S: RgbStandard> From<Rgb<S, u8>> for u32 {
+    fn from(color: Rgb<S, u8>) -> Self {
+        Rgb::into_u32::<channels::Argb>(color)
+    }
+}
+
+impl<S: RgbStandard> From<Rgba<S, u8>> for u32 {
+    fn from(color: Rgba<S, u8>) -> Self {
+        Rgba::into_u32::<channels::Rgba>(color)
+    }
+}
+
+impl<C: RgbChannels> From<u32> for Packed<C> {
+    fn from(color: u32) -> Self {
+        Packed {
+            color,
+            channel_order: PhantomData,
+        }
+    }
+}
+
+impl<S, C> From<Rgb<S, u8>> for Packed<C>
+where
+    S: RgbStandard,
+    C: RgbChannels,
+{
+    fn from(color: Rgb<S, u8>) -> Self {
+        Self::from(Rgba::from(color))
+    }
+}
+
+impl<S, C> From<Rgba<S, u8>> for Packed<C>
+where
+    S: RgbStandard,
+    C: RgbChannels,
+{
+    fn from(color: Rgba<S, u8>) -> Self {
+        let bytes = C::split_rgb(color);
+        Packed {
+            color: u32::from_be_bytes([bytes.0, bytes.1, bytes.2, bytes.3]),
+            channel_order: PhantomData,
+        }
+    }
+}
+
+impl<S: RgbStandard> From<u32> for Rgb<S, u8> {
+    fn from(color: u32) -> Self {
+        Self::from_u32::<channels::Argb>(color)
+    }
+}
+
+impl<S, C> From<Packed<C>> for Rgb<S, u8>
+where
+    S: RgbStandard,
+    C: RgbChannels,
+{
+    fn from(packed: Packed<C>) -> Self {
+        Rgba::from(packed).color
+    }
+}
+
+impl<S: RgbStandard> From<u32> for Rgba<S, u8> {
+    fn from(color: u32) -> Self {
+        Self::from_u32::<channels::Rgba>(color)
+    }
+}
+
+impl<S, C> From<Packed<C>> for Rgba<S, u8>
+where
+    S: RgbStandard,
+    C: RgbChannels,
+{
+    fn from(packed: Packed<C>) -> Self {
+        let bytes = packed.color.to_be_bytes();
+        C::combine_rgb((bytes[0], bytes[1], bytes[2], bytes[3]))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    // use crate::rgb::{Abgr, Argb, Bgra, Rgba};
+    use crate::rgb::packed::channels::{Abgr, Argb, Bgra, Rgba};
+    use crate::{Packed, Srgb, Srgba};
+
+    #[test]
+    fn rgba() {
+        let a1: Packed<Rgba> = Srgb::new(0.5, 0.0, 0.0).into_format().into();
+        let a2: Packed<Rgba> = Srgb::new(0.0, 1.0, 0.0).into_format().into();
+        let a3: Packed<Rgba> = Srgb::new(0.0, 0.0, 0.5).into_format().into();
+        let x1: u32 = 0x8000_00FF;
+        let x2: u32 = 0x00FF_00FF;
+        let x3: u32 = 0x0000_80FF;
+        assert_eq!(a1.color, x1);
+        assert_eq!(a2.color, x2);
+        assert_eq!(a3.color, x3);
+
+        let unpacked: Srgb<u8> = Packed::<Rgba>::from(0x80FF_80FF).into();
+        assert_relative_eq!(
+            Srgb::new(0.5, 1.0, 0.5),
+            unpacked.into_format(),
+            epsilon = 0.01
+        );
+
+        let b1: Packed<Rgba> = Srgba::new(0.5, 0.0, 0.0, 0.0).into_format().into();
+        let b2: Packed<Rgba> = Srgba::new(0.0, 1.0, 0.0, 0.0).into_format().into();
+        let b3: Packed<Rgba> = Srgba::new(0.0, 0.0, 0.5, 0.0).into_format().into();
+        let b4: Packed<Rgba> = Srgba::new(0.0, 0.0, 0.0, 1.0).into_format().into();
+        let y1: u32 = 0x8000_0000;
+        let y2: u32 = 0x00FF_0000;
+        let y3: u32 = 0x0000_8000;
+        let y4: u32 = 0x0000_00FF;
+        assert_eq!(b1.color, y1);
+        assert_eq!(b2.color, y2);
+        assert_eq!(b3.color, y3);
+        assert_eq!(b4.color, y4);
+
+        let unpacked: Srgba<u8> = Packed::<Rgba>::from(0x80FF_80FF).into();
+        assert_relative_eq!(
+            Srgba::new(0.5, 1.0, 0.5, 1.0),
+            unpacked.into_format(),
+            epsilon = 0.01
+        );
+    }
+
+    #[test]
+    fn argb() {
+        let a1: Packed<Argb> = Srgb::new(0.5, 0.0, 0.0).into_format().into();
+        let a2: Packed<Argb> = Srgb::new(0.0, 1.0, 0.0).into_format().into();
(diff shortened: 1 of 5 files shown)


okaneco:
There are a few different ways to do this and I wasn't sure what was best. I've seen all different orderings in libraries I've used, so I figured it's best to include them all.

I wasn't sure if I should implement any further derives on the structs like serialization or `#[repr(C)]` since it only has one field. Another way of attacking this was with a trait but I chose a struct with only one field so that the types couldn't be mixed together unintentionally.

Ogeon:
This is an interesting way of tackling it. I would probably just have made a simple conversion method like `from_hex(ordering: ChannelOrdering, value: u32) -> Rgb<S, u8>` (where `ChannelOrdering` is an enum) and nothing more, but I see the value of making it an actual type if it's part of a larger workflow.

Thinking a bit about this, I think making to/from_hex be parametric on the value type would be to over engineer it for no good reason if there are separate packed types that can have their own converters. The bare `u32` is the odd one, since it doesn't have a defined channel ordering.

Would something like, this make sense? See it as a sketch, but the idea is to twist the concept a bit to turn the channel ordering into a type parameter (allowing operations on any packed RGB) and unify that with the `{from,into}_hex` methods. The downside is, of course, that it's complex in a different way and it forces the user to be aware of the channel ordering.

```rust
/// Splits and combines RGB(A) types with some channel ordering.
pub trait RgbChannels {
    fn split_rgb<S>(rgb: Rgb<S, u8>) -> (u8, u8, u8);
    fn combine_rgb<S>(channels: (u8, u8, u8)) -> Rgb<S, u8>;
}


struct Abgr;
impl SplitChannels for Abgr {
    //...
}

//...

// Adding Pixel here makes it easy to convert to and from slices of u32.
#[derive(..., Pixel)]
#[repr(transparent)]
struct PackedRgb<C: RgbChannels> {
    color: u32,
    channel_order: PhantomData<C>, // This is not a nice aspect of this, but implementing `From<u32>` should make it less painful to construct.
}

impl<S, C: RgbChannels> Into<Rgb<S, u8>> for PackedRgb<C> {
    fn into(self) -> Rgb<S, u8> {
        //...
        C::combine_rgb((c1, c2, c3))
    }
}

// rgb.rs:
impl<S> Rgb<S, u8> {
    fn from_hex<C: RgbChannels>(color: u32) -> Self {
        PackedRgb::<C>::from(color).into()
    }

    fn into_hex<C: RgbChannels>(self) -> Packed<C> {
        PackedRgb::from(self).color
    }
}
```

Whichever way it's done, I think it's best to start with a conservative approach and only allow it for RGB(A)8, to begin with. That's the most straightforward one and anyone who want to convert to any other component format can easily do it anyway.

okaneco:
I liked those ideas a lot and thought they would clean up some parts nicely. 👍 

It was very helpful and chopped down about 100 lines of repetitive code. Luckily I had the test data all there waiting for the implementation to finish which gave me confidence in refactoring, and the tests barely needed changing.

During my first go, I did try some type of channel ordering marker data but that and other complexity got in the way of a rough initial implementation. The bounds trip me up with the error messages not being that helpful sometimes and any time I write what I think should work I run into features that are currently unstable, but I think I'm starting to understand a bit more.

---

In no particular order, there are a few implementation details which came up in the refactor that I'm not sure about.

### `RgbaPacked`
The four impls of `RgbChannels` are on `RgbaPacked`, `Argb`, `Bgra`, or `Abgr`. It's a horrible name but I didn't want to collide with the type `Rgba` (I was confused coming across `Srgb` the type and struct when first using the library). I initially named it `RgbaDefault` since the rest of the crate is RGBA order and set it as the default for the trait `RgbChannels`. `RgbaPacked` feels verbose and confusing since the containing struct is called `PackedRgb`.
```rust
pub struct PackedRgb<C: RgbChannels = Argb> {
    /// The sRGB color packed into a `u32`.
    pub color: u32,

    /// The channel ordering for red, green, blue, and alpha components in the
    /// packed integer; can be `Argb`, `Abgr`, `Bgra`, or `RgbaPacked`.
    #[palette_unsafe_zero_sized]
    pub channel_order: PhantomData<C>,
}
```

### Defaulting to `Argb` and writing Hex values
This continues from the previous point. An observation made during the first implementation was that I normally think in RGBA like the library, but when I write hex RGB values I'm really thinking ARGB (0x00rrggbb). The rest of the crate is RGBA, why should this be ARGB?

Because the least surprising thing to me for `Rgb::from_hex(0x112233)` would be getting rgb(0x11, 0x22, 0x33) and *not* rgb(0x00, 0x11, 0x22). For this reason I decided to make ARGB the default channel ordering for `PackedRgb<C>`.

#### **Rgb packing, 0x00 or 0xFF for dummy alpha?**

When creating a `PackedRgb` from Rgb type, it needs an alpha value for one of the bytes. I can see arguments for either 0 or 255. I chose 0 for now because with `Argb`, `Rgb::into_hex` would return 0x112233 while `RgbaPacked` would be 0x11223300. The first result is what I would expect however it might not be obvious to users that `Argb` is the default since there's a chance they haven't had to think about this. That's why I do think it's good that channel order is maintained as a parameter for users to keep in mind.

#### Type annotations
I had some slight frustrations with how it still needs the <C> type annotation despite a default being specified, is there something more I could do so that's not needed? I don't really understand why the `Argb` is needed here, and gets slightly long to type with needing to specify the `u8` as well (I know that's being worked on). I can see users expecting `Rgba` to go there and being surprised with results, I've added some more comments mentioning ARGB to the functions in `rgb.rs`.
```rust
let unpacked = Srgb::<u8>::from_hex::<Argb>(0xFF8000);
```

palette/src/rgb/packed.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Split RGBA components into a `(u8, u8, u8, u8)` tuple.
+    fn split_rgb<S: RgbStandard>(rgb: Rgba<S, u8>) -> (u8, u8, u8, u8);
+    /// Create an RGBA color from a `(u8, u8, u8, u8)` tuple.
+    fn combine_rgb<S: RgbStandard>(channels: (u8, u8, u8, u8)) -> Rgba<S, u8>;
+}
+
+/// RGBA color packed in RGBA order.
+///
+/// See [PackedRgb](struct.PackedRgb.html) for more details.
+pub struct RgbaPacked;
```
Ogeon: The fact that this is in a separate module makes it possible to tell them apart, so I think the `Packed` part can be dropped. It's still annoying that color spaces are also named after their channels. It could perhaps be further emphasizes with a module for these types (like `channels::Rgba`), or by going a bit unorthodox and using all capitals. Using all caps may be confusing, though.

palette/src/rgb/packed.rs
```diff
@@ -0,0 +1,364 @@
...
+/// assert_eq!(0x80, float.round() as u32);
+/// ```
+///
+/// When an `Srgb` type is packed, the alpha value will be `0x00` in the
+/// corresponding `u32`. Converting from a packed color type back to an `Srgb`
+/// type will disregard the alpha value.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Pixel)]
+#[palette_internal]
+#[repr(C)]
+pub struct PackedRgb<C: RgbChannels = Argb> {
```
Ogeon: The `Rgb` suffix may be redundant, given by the context and the type parameter. I'm definitely open for other names, both here and elsewhere.

palette/src/rgb/rgb.rs
```diff
@@ -118,6 +118,22 @@ impl<S: RgbStandard, T: Component> Rgb<S, T> {
...
+    /// See [PackedRgb](struct.PackedRgb.html) for more details.
+    pub fn from_hex<C: RgbChannels>(color: u32) -> Self {
+        PackedRgb::<C>::from(color).into()
+    }
+
+    /// Convert to a [PackedRgb](struct.PackedRgb.html) with color packed into
+    /// a `u32`. Defaults to ARGB ordering (0x00RRGGBB).
+    pub fn into_hex<C: RgbChannels>(self) -> PackedRgb<C> {
+        PackedRgb::from(self)
+    }
```
Ogeon: These are asymmetric now, when we take `u32` as input, but give `PackedRgb` as output. I think it's redundant to give out `PackedRgb` in this case, since it can easily be replaced by `PackedRgb::from(my_rgb)`. Thinking a bit more about it, it may be better to call them `{from|into}_u32` to be really explicit about it, and also perhaps implement `From<u32>` with the ARGB default.

palette/src/rgb/packed.rs
```diff
@@ -0,0 +1,364 @@
...
+        let src: Alpha<Rgb<S, u8>, u8> =
+            Rgba::from_components((color.red, color.green, color.blue, 0u8));
+        let bytes = C::split_rgb(src);
+        PackedRgb {
+            color: 0x0000_0000
+                | ((bytes.0 as u32) << 24)
+                | ((bytes.1 as u32) << 16)
+                | ((bytes.2 as u32) << 8)
+                | (bytes.3 as u32),
+            channel_order: PhantomData,
```
Ogeon: This could let the `From<Rgba>` implementation do the bit shifting. :slightly_smiling_face: It could just end with `Self::from(src)`. The same goes for `Into`.

palette/src/rgb/packed.rs
okaneco: I can't think of another name, I think `Packed` works best for now. No other names stick out as bad to me right now.

palette/src/rgb/rgb.rs
okaneco: I agree they shouldn't be asymmetric and I originally wrote a blurb in my last comment as a concern but got rid of it. 

I went back and forth on u32 or the struct return. I think returning the struct makes the user have to reason about the channel ordering more which can be a good thing but `into_u32` should return a u32 if we're changing the method name.

okaneco:
> As for the default for alpha, I wonder if it would be any bad consequences if we pick 0xFF. I'm thinking it may even have the opposite effect of being helpful when it's interpreted on the receiving end.

I changed to 0xFF again because that's probably more in line and expected during use to have full alpha for an RGB color. If you want 0x00 for the alpha use an `Rgba` with the same hex number instead of `Rgb` so we have the best of both worlds.

> I believe the issue may be minimized if we implement and maybe somehow promote From<u32> as a shortcut. Then it would just be Srgb::<u8>::from(...) at worst.

What would this look like roughly? If I try to implement for `Packed<C>` and again for `Packed<Argb>` I get an error for conflicting implementations of a trait.

okaneco:
I was able to simplify the `From<Rgb>` but not `Into<Rgb>`. This builds:

```rust
impl<S, C> Into<Rgb<S, u8>> for Packed<C>
where
    S: RgbStandard,
    C: RgbChannels,
{
    fn into(self) -> Rgb<S, u8> {
        Rgb::from_u32::<C>(self.color)
    }
}
```

but fails `cargo test`. It seems to be caused by spawned thread stacks (in this case, testing) being smaller than main thread stacks https://stackoverflow.com/a/42960702.
```rust
thread 'rgb::packed::test::argb' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/Users/foo/rust/palette_local/target/debug/deps/palette-658e34a922312806` (signal: 6, SIGABRT: process abort signal)
```

---

This is the current implementation. I tried to piggyback on Rgba when I could it didn't seem possible to use anything more.
```rust

fn into(self) -> Rgb<S, u8> {
    let bytes: [u8; 4] = self.color.to_be_bytes();
    let color: Alpha<Rgb<S, u8>, u8> = C::combine_rgb((bytes[0], bytes[1], bytes[2], bytes[3]));
    Rgb::from_components((color.red, color.green, color.blue))
}
```
Is there an easier way to convert from `Rgba` to `Rgb` in this case? 

```rust
error[E0277]: the trait bound `u8: FromF64` is not satisfied
   --> palette/src/rgb/packed.rs:158:19
    |
158 |         Rgb::from(color)
    |                   ^^^^^ the trait `FromF64` is not implemented for `u8`
    |
    = note: required because of the requirements on the impl of `component::FloatComponent` for `u8`
    = note: required because of the requirements on the impl of `std::convert::From<alpha::Alpha<rgb::rgb::Rgb<S, u8>, u8>>` for `rgb::rgb::Rgb<_, u8>`
    = note: required by `std::convert::From::from`
```

palette/src/rgb/packed.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl<S, C> Into<Rgb<S, u8>> for Packed<C>
+where
+    S: RgbStandard,
+    C: RgbChannels,
+{
+    fn into(self) -> Rgb<S, u8> {
+        let bytes: [u8; 4] = self.color.to_be_bytes();
+        let color: Alpha<Rgb<S, u8>, u8> = C::combine_rgb((bytes[0], bytes[1], bytes[2], bytes[3]));
+        Rgb::from_components((color.red, color.green, color.blue))
```
Ogeon: When this is changed to be a `From` implementation, you can change it to `Rgba::from(packed).color`. `Alpha<Rgb>` wraps an `Rgb` as `color` and exposes its content via `Deref`, so that's it!

palette/src/rgb/packed.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl<S, C> From<Rgb<S, u8>> for Packed<C>
+where
+    S: RgbStandard,
+    C: RgbChannels,
+{
+    fn from(color: Rgb<S, u8>) -> Self {
+        // Alpha value of 0xFF declared here as specified in doc for `Packed`
+        let src: Alpha<Rgb<S, u8>, u8> =
+            Rgba::from_components((color.red, color.green, color.blue, 0xFF));
```
Ogeon: Pretty sure `Rgba::from(color)` should also work, unless I have managed to confuse myself. It defaults to fully opaque. Otherwise it's `Alpha {color, alpha: 0xFF}`. :slightly_smiling_face:

palette/src/rgb/packed.rs
```diff
@@ -0,0 +1,305 @@
...
+    C: RgbChannels,
+{
+    fn from(color: Rgba<S, u8>) -> Self {
+        let bytes = C::split_rgb(color);
+        Packed {
+            color: 0x0000_0000
+                | ((bytes.0 as u32) << 24)
+                | ((bytes.1 as u32) << 16)
+                | ((bytes.2 as u32) << 8)
+                | (bytes.3 as u32),
```
Ogeon: I saw you switched to `to_be_bytes` above. I assume this can be replaced with `from_be_bytes`, then. :slightly_smiling_face:

Ogeon:
> What would this look like roughly?

We are probably in name confusion land. I had this in mind:

```rust
impl<S: RgbStandard> From<u32> for Rgb<S, u8> {
    fn from(color: u32) -> Self {
        Self::from_u32::<Argb>(color)
    }
}

impl<S: RgbStandard> From<u32> for Alpha<Rgb<S, u8>, u8> {
    fn from(color: u32) -> Self {
        // Notice the ordering here, going for the seemingly popular one.
        Self::from_u32::<Rgba>(color)
    }
}
```

That way one can get quick and dirty default conversions. I suppose `From<Rgb<...>> for u32` is also possible now.

palette/src/rgb/packed.rs
okaneco: Good catch 👍

okaneco:
Oh I see what you mean now. That works nicely and gives us a somewhat sane "universal" behavior. I assume that should live in `packed.rs` and not `rgb.rs`.

I changed the other `Into` for Rgba to a `From`. I implemented `From<Rgb<...>> for u32` since I could. The other cleanups removed a lot of clutter. Thanks for explaining the Alpha<Rgb>, I didn't fully grasp it until then.

Ogeon:
I would probably have placed it in rgb.rs, since it doesn't mention anything from packed.rs publicly, but both are valid option! (No need to change)

Yes, the `Alpha` type is a bit magical. `Deref` is not really meant to be used that way, since it's not really a reference, but I figured being a wrapper would be close enough and that the benefits would outweigh the costs. Like with `Packed` and `Rgb` and all other types, it reduces them from combinatorial monsters to something more manageable.

Anyway, I do really like the way it's implemented now! Very nice! Thanks! The only improvements I can think of now is just a tiny bit of documentation polish:

 * Mentioning/showing `u32::from(Rgb)` and `Rgb::from(u32)` in the documentation for `from_u32` and `into_u32`, just to make people more aware of them.
 * Maybe demoing converting a `&[u32]` to `&[Packed]`, via the `Pixel` trait. It has a few examples of its own that can be used as inspiration. It's also a free unit test.

Essentially a couple of "did you know..." examples, that I think may make someone's life a bit easier. Other than those details, I'm ready to merge when you are!

palette/src/rgb/packed.rs
```diff
@@ -49,6 +49,22 @@ use crate::Pixel;
 /// When an `Rgb` type is packed, the alpha value will be `0xFF` in the
 /// corresponding `u32`. Converting from a packed color type back to an `Rgb`
 /// type will disregard the alpha value.
+///
+/// `Packed` implements [Pixel](encoding/pixel/trait.Pixel.html) and can be
+/// contructed from a slice of `&[u32]`.
```
Ogeon: I just noticed a typo:

```suggestion
/// constructed from a slice of `&[u32]`.
```

Ogeon:
Nice! I think that's all for this one! Feel free to squash it.

okaneco:
Those examples good ideas, and I'm happy with how this turned out. 👍  

Massive Github outage so wasn't able to do post comments or push. I'll squash this now.

I've not used the `Pixel` trait before so I wasn't sure what the example should show.

I was trying to think of ways to improve discovery or surface these features. I noticed there's no `Rgb` or `Srgb` structs or types visible on the doc page and wasn't sure why that is, is that due to re-exports? There are plenty of doctest examples that show how to use them but you have to know to click the `rgb` module to see the structs/types.
https://docs.rs/palette/0.5.0/palette/#modules

Ogeon:
I think the fact that it refers to the Pixel trait, combined with the small and straight forward example, shows what it may be good for.

You are right in that it's not super easy to find examples from the documentation root. Part of why I haven't put the `Rgb` type there is that I don't want to encourage people to use it over the `Srgb` and `LinSrgb` aliases. It was something I noticed in an earlier version. People tend to reach for the familiar name, but the `Rgb` type is not that ergonomic on its own. It's often nicer to make an alias for it, especially because one will often need two versions (linear and non-linear) of it.

The examples and documentation is in need of a bit of an overhaul, I think, but that's much larger than this PR. Hence why I'm not requiring anything more fancy than this. I would like to document various common patterns too, a bit like making a cookbook, for people to look up how to do things like read and write to images and/or pixel buffers, how to interoperate with other graphics crates, how to set up a basic linear pipeline, and all sorts of other things one may use this for.

Ogeon:
But the point of the Pixel trait is glue code for converting arrays and slices to Palette types. If you have some data that can be represented as a slice or array, you can turn it into a Palette color without copying anything. It's pretty neat to be able to take a massive `&mut [u8]` from somewhere, turn it into `&mut [Srgb<u8>]` with almost no cost, and get all the type system benefits.

okaneco:
That makes sense, I can imagine some applications with a slice of u32.

I think that wraps this up. I accidentally re-inserted a gap between the documentation and the `Packed` struct when I added the `Pixel` example; with the gap, there's no context for the documentation to appear when I hover over in my IDE. I also removed a type annotation in the assert that was leftover from copying the value.

Ogeon:
Alright! Let's get it merged.

bors r+

PR #171, opened 2020-03-13
Remove remaining extern proc_macro, remove `strict` feature, and update documentation

### LINKED ISSUE #129: Show how to convert from "raw" color types in the documentation and the readme
As mentioned in #128, there is no clear (or really easy to find) documentation for when one starts out with `[u8; 3]` or any other "raw" format (using `from_raw`). This should be added to show how interoperability with other crates and systems works.

okaneco:
Remove `extern crate proc_macro` in `palette_derive`
Remove `strict` feature which denies warnings and missing docs
Add `RUSTFLAGS="-D warnings"` flag for CI to replace `strict`
Add documentation to show raw color type usage
Update CONTRIBUTING.md with some grammar fixes and remove unnecessary style sections

Closes #129

### DIFF
--- a/palette/src/encoding/pixel/mod.rs
+++ b/palette/src/encoding/pixel/mod.rs
@@ -116,7 +116,7 @@ pub unsafe trait Pixel<T>: Sized {
         unsafe { P::from_raw_parts_mut(self as *mut Self as *mut T, Self::CHANNELS) }
     }
 
-    /// Convert from raw color components.
+    /// Convert into raw color components.
     #[inline]
     fn into_raw<P: RawPixelSized<T>>(self) -> P {
         assert_eq!(P::CHANNELS, Self::CHANNELS);
--- a/palette/src/lib.rs
+++ b/palette/src/lib.rs
@@ -133,12 +133,35 @@
 //! When the desired processing is done, it's time to encode the colors back
 //! into some image format. The same rules applies as for the decoding, but the
 //! process reversed.
+//!
+//! # Working with Raw Data
+//!
+//! Oftentimes, pixel data is stored in a raw buffer such as a `[u8; 3]`. The
+//! [`Pixel`](encoding/pixel/trait.Pixel.html) trait allows for easy interoperation between
+//! Palette colors and other crates or systems. `from_raw` can be used to
+//! convert into a Palette color, `into_format` converts from  `Srgb<u8>` to
+//! `Srgb<f32>`, and finally `into_raw` to convert from a Palette color back to
+//! a `[u8;3]`.
+//!
+//! ```rust
+//! use approx::assert_relative_eq;
+//! use palette::{Srgb, Pixel};
+//!
+//! let buffer = [255, 0, 255];
+//! let raw = Srgb::from_raw(&buffer);
+//! assert_eq!(raw, &Srgb::<u8>::new(255u8, 0, 255));
+//!
+//! let raw_float: Srgb<f32> = raw.into_format();
+//! assert_relative_eq!(raw_float, Srgb::new(1.0, 0.0, 1.0));
+//!
+//! let raw: [u8; 3] = Srgb::into_raw(raw_float.into_format());
+//! assert_eq!(raw, buffer);
+//! ```
 
 // Keep the standard library when running tests, too
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 #![doc(html_root_url = "https://docs.rs/palette/0.5.0/palette/")]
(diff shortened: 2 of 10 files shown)


okaneco:
When I saw the announcement for Rust 1.42 today, the first thing I thought of were those extern crates that couldn't be removed in the 2018 migration. Now the `proc_macro` can be, I thought there were more but it was just one. https://blog.rust-lang.org/2020/03/12/Rust-1.42.html#use-proc_macro:tokenstream;-now-works
>In this release, if you are using Cargo, you no longer need this line when working with the 2018 edition; you can use use like any other crate. Given that most projects will already have a line similar to use proc_macro::TokenStream;, this change will mean that you can delete the extern crate proc_macro; line and your code will still work. This change is small, but brings procedural macros closer to regular code.

Above the `extern crate` I saw the strict feature for denying warnings. If the RUSTFLAGS are instead added to the Travis environment flags then the feature doesn't seem necessary. The `strict` feature seems like more overhead and indirection to figure out what it's doing in the crate. I recently ran into an issue with `Error::description` being a soft-deprecated method but it wasn't apparent until I ran the CI for it on nightly/beta. Running the test locally on stable wouldn't have caught that so it doesn't seem like a `strict` feature is that helpful compared to the alternative for CI checks. It's easy to miss the `strict` feature in CONTRIBUTING.md and the build will fail anyway when the PR is attempted. After finishing the code for this, I did come across discussion you were involved in here and I don't think much has changed https://users.rust-lang.org/t/how-to-fail-a-build-with-warnings/2687. Developing locally I do use `strict` for `palette` or already have the env flags aliased to my `cargo check` and `cargo test`.

This lead to looking at the documentation in the `CONTRIBUTING.md` and giving that a quick edit or update based on what I knew changed. Since I was touching documentation and had recently worked with `from_raw` I decided to approach the doc issue. 

The PR is small and I know it's not totally related so I can break this into other PRs/rollback the CI parts.

palette/src/encoding/pixel/mod.rs
```diff
@@ -116,7 +116,7 @@ pub unsafe trait Pixel<T>: Sized {
         unsafe { P::from_raw_parts_mut(self as *mut Self as *mut T, Self::CHANNELS) }
     }
 
-    /// Convert from raw color components.
+    /// Convert into raw color components.
```
okaneco: I think this was intended to be `into`?

palette/src/lib.rs
```diff
@@ -133,12 +133,35 @@
...
+//! let raw: [u8; 3] = Srgb::into_raw(raw_float.into_format());
+//! assert_eq!(raw, buffer);
+//! ```
 
 // Keep the standard library when running tests, too
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 #![doc(html_root_url = "https://docs.rs/palette/0.5.0/palette/")]
-#![cfg_attr(feature = "strict", deny(missing_docs))]
-#![cfg_attr(feature = "strict", deny(warnings))]
+#![deny(missing_docs)]
```
Ogeon: I'm hesitant to deny this one, rather than having it as a warning. It would be bad if someone's future build breaks because of missing documentation, even though it's very unlikely. It will still become denied in CI.

okaneco:
Yes, upon reaching the "Code Style" section I realized a lot of it was from a different era and didn't need to stay but left it for you to decide. I kept the part about docs and text length because `rustfmt` can't help us there.

We can stay in the root of the directory but the `no_std_test` needs to be excluded from the builds otherwise it'll try to build on stable/beta.

okaneco:
Nevermind, looks like we'll have to keep the old way. I forgot about this quirk with workspaces.
```
testing with --no-default-features --features std
error: --features is not allowed in the root of a virtual workspace
note: while this was previously accepted, it didn't actually do anything
The command "bash ./scripts/test_" exited with 101.
```

Ogeon:
Ah, right, in that case I don't think there's any gain from building from the root.

okaneco:
Unless there's anything else, I think this is good to go.

Ogeon:
I don't know if there was a misunderstanding, but there doesn't seem to be any warning attribute for missing documentation. It's allowed by default: https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#missing-docs

okaneco:
Ah, I did misunderstand that. I've added it back to `lib.rs`.

Ogeon:
The rest looks good!

Ogeon:
Nice, thanks a lot! And I forgot to acknowledge the proof reading earlier, but it's very much appreciated. It's always nice with these kind of pull requests.

bors r+

PR #176, opened 2020-03-28
Rewrite the conversion traits to work more like From and Into

### LINKED ISSUE #41: Converting from a wider gamut to a smaller one should return an option

### LINKED ISSUE #111: Replace the hidden module for derives with a hidden function
This allows derives to be used in functions (see workaround from https://github.com/rust-lang/rust/pull/51952). The problem is that `extern crate` seems to be hard or impossible to use in a function scope, so it will most likely be a breaking change.

Ogeon:
This replaces the old `FromColor` and `IntoColor` traits, as well as the implementations of `From` and `Into` between color types, with new traits that are more flexible. The new `FromColor` and `IntoColor` traits will also clamp the output to prevent out-of-bounds results.

It's a _very_ breaking change. To migrate from the old system, any failing uses `from` and `into` should be replaced with `from_color` and `into_color`.  Any custom color that used to implement `IntoColor` should switch to implementing `FromColorUnclamped`.

A new trait, called `Transparency`, has also been added to make it possible to implement `FromColorUnclamped` for `Alpha` without running into conflicting implementations. It makes it possible to convert from any color type that implements `Transparency`, and not just `Alpha<A, T>` to `Alpha<B, T>`, for example.

The semantics of the new traits are a bit different, so it may affect type inference. It may also be necessary to convert using more than one step. This is mostly for the RGB family and Luma, when changing RGB space or standard.

The syntax for the helper attributes were also changed to the more common `#[palette(...)]` pattern. It helped when implementing them and looks a bit nicer, IMO.

Closes #41, fixes #111.


Ogeon:
It's finally here, after all this time. I don't know how many times I rewrote the derives and how many versions of the traits I went through, but I'm quite happy with the current result. I was also helped a lot by more recent language features, such as not having to implement `Into*` for the "foreign" types.

I'm leaving it up for a while, in case anyone want to help reviewing. I may very well have forgotten something along the way.

palette/src/alpha/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// This is typically `Self`.
+    type Color;
+
+    /// The color type with transparency applied.
+    ///
+    /// This is typically `Alpha<Self::Color, A>`.
+    type WithAlpha: Transparency<A, Color = Self::Color, WithAlpha = Self::WithAlpha>;
+
+    /// Transforms the color into a transparent color with the provided
+    /// alpha value. If `Self` already has a transparency, that one is
```
okaneco:
```suggestion
    /// alpha value. If `Self` already has a transparency, that value is
```
I think the instances of "that one" in this doc section would be better as "that value" or "that field". I try to generally avoid pronouns and vague words like "it" and "one" in tech writing when possible.

palette/src/alpha/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+
+mod alpha;
+
+/// A trait for color types that can have or be given transparency (alpha channel).
+///
+/// It's an interface for adding, removing and setting the transparency of a
+/// color type. The type itself doesn't need to store the transparency
+/// value, as long as it can be transformed into or wrapped in a type that
+/// does have a representation of transparency. This would typically be done
+/// by wrapping it in an [`Alpha`](struct.Alpha.html) instance.
```
okaneco:
```rust
/// `Transparency` is an interface for adding, removing and setting the alpha
/// channel of a color type. The type itself doesn't need to store the
/// transparency value as it can be transformed into or wrapped in a type that
/// has a representation of transparency. This would typically be done by
/// wrapping it in an [`Alpha`](struct.Alpha.html) instance.
```

palette/src/alpha/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// assert_eq!(transparent.alpha, 0.3);
+    ///
+    /// // This changes the transparency to 0.8
+    /// let transparent = transparent.with_alpha(0.8f32);
+    /// assert_eq!(transparent.alpha, 0.8);
+    /// ```
+    fn with_alpha(self, alpha: A) -> Self::WithAlpha;
+
+    /// Removes the transparency from the color. If `Self::Color` has
+    /// an internal transparency field, that one should be set to
```
okaneco:
```suggestion
    /// an internal transparency field, that field will be set to
```
similar change in L119

palette/src/alpha/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    ///
+    /// let transparent = Srgba::new(255u8, 0, 255, 10);
+    ///
+    /// // This unwraps the color information from the `Alpha` wrapper
+    /// let color = transparent.without_alpha();
+    /// assert_eq!(transparent.color, color);
+    /// ```
+    fn without_alpha(self) -> Self::Color;
+
+    /// Removes the color into separate color and transparency values.
```
okaneco:
```suggestion
    /// Splits the color into separate color and transparency values.
```

palette/src/convert.rs
```diff
@@ -1,488 +1,277 @@
...
+//! may be any other type, including type parameters.
+//!
+//! * `component = "some::component::Type"`: Sets the color
+//! component type that should be used when deriving. The default is `f32`, but
+//! it may be any other type, including type parameters.
+//!
+//! * `rgb_space = "some::rgb_space::Type"`: Sets the RGB space
+//! type that should be used when deriving. The default is to either use `Srgb`
+//! or a best effort to convert between spaces, so sometimes it has to be set
+//! to a specific type. This does also accept type parameters.
```
okaneco:
```suggestion
//! to a specific type. This also accepts type parameters.
```

palette/src/convert.rs
```diff
@@ -508,127 +297,184 @@ impl<T> OutOfBounds<T> {
...
+/// but important, differences in their semantics:
+///
+/// * `FromColor` and `IntoColor` are allowed to be lossy, meaning converting `A -> B -> A`
+/// may result in a different value than the original. This applies to `A -> A` as well.
+/// * `From<Self>` and `Into<Self>` are blanket implemented, while `FromColor<Self>` and
+/// `IntoColor<Self>` have to be manually implemented. This allows additional flexibility,
+/// such as allowing implementing `FromColor<Rgb<S2, T>> for Rgb<S1, T>`.
+/// * Implementing `FromColorUnclamped` and `Limited` is enough to get all the other conversion
+/// traits, while `From` and `Into` would not be possible to blanket implement in the same way.
+/// This does also reduce the work that need to be done by macros.
```
okaneco:
```suggestion
/// This also reduces the work that needs to be done by macros.
```

palette/src/convert.rs
```diff
@@ -508,127 +297,184 @@ impl<T> OutOfBounds<T> {
...
+///
+/// See the [`convert`](index.html) module for how to implement `FromColorUnclamped` for
+/// custom colors.
+pub trait FromColor<T>: Sized {
+    ///Convert from T with values clamped to the color defined bounds.
```
okaneco:
```suggestion
    /// Convert from T with values clamped to the color defined bounds.
```

palette/src/convert.rs
```diff
@@ -508,127 +297,184 @@ impl<T> OutOfBounds<T> {
...
+/// A trait for unchecked conversion of one color from another.
+///
+/// See [`FromColor`](trait.FromColor.html) for a lossy version of this trait.
+/// See [`TryFromColor`](trait.TryFromColor.html) for a trait that gives an error when the result
+/// is out of bounds.
+///
+/// See the [`convert`](index.html) module for how to implement `FromColorUnclamped` for
+/// custom colors.
+pub trait FromColorUnclamped<T>: Sized {
+    ///Convert from T. The resulting color might be invalid in its color space.
```
okaneco:
```suggestion
    /// Convert from T. The resulting color might be invalid in its color space.
```

okaneco:
This looks great 🎉 

I don't have much to add other than small doc suggestions.

Since the file's being modified, the README's description section up top can probably be shortened to something like `A Rust library that makes linear color calculations and color space conversion easy and accessible for anyone.` since there's not a general color type anymore?

Ogeon:
Thank you for taking a look at this! Those are good suggestions and feedback. I value any improvements to my writing, especially since English isn't my first language. I'll implement the suggestions and a couple of other improvements. I somehow forgot that "lossless" is a word.

Ogeon:
I applied the suggested documentation changes (and a few more) and rebased to get the latest changes.

Ogeon:
I decided to rename the `Transparency` to `WithAlpha` to keep the terminology more consistent. I was on the fence at first, but everything in and around it uses "alpha", so in the end I thought it would be weirder not to. This covers the last thing I wanted to change before squash+merge.

Ogeon:
bors  r+

PR #179, opened 2020-04-17
Add benchmarking with criterion to track changes

### LINKED ISSUE #95: Add benchmarks to have a baseline and be able to track improvements
There is currently nothing that shows if there are any improvements or regressions between changes. If criterion is used, we may be able to set up something like https://github.com/matchai/criterion-compare-action (or if there are better alternatives) to visualize how each pull request affects the numbers.

Suggestions for what's worth measuring are welcome!

okaneco:
Add coverage for Rgb color family and Cie family conversions using the colormine test data
Add benchmarks for matrix operations
Add benchmarks for testing conversion of Srgb\<u8\> and LinSrgb\<f32\>
Small doc fixups

Benchmarks can be run with `cargo bench`. You can also run select benchmarks by typing in a regex pattern that matches the benchmark id name.

Closes #95

### DIFF
--- a/palette/benches/cie.rs
+++ b/palette/benches/cie.rs
@@ -1,0 +1,88 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion, Throughput};
+use palette::convert::FromColorUnclamped;
+use palette::{Lab, Lch, Xyz, Yxy};
+
+#[path = "../tests/convert/data_color_mine.rs"]
+#[allow(dead_code)]
+mod data_color_mine;
+use data_color_mine::{load_data, ColorMine};
+
+/* Benches the following conversions:
+    - xyz to lab
+    - lch to lab
+    - lab to lch
+    - rgb to xyz
+    - yxy to xyz
+    - lab to xyz
+    - xyz to yxy
+*/
+
+fn cie_conversion(c: &mut Criterion) {
+    let mut group = c.benchmark_group("Cie family");
+    let colormine: Vec<ColorMine> = load_data();
+    let lab: Vec<Lab> = colormine
+        .iter()
+        .map(|x| Lab::from_color_unclamped(x.xyz))
+        .collect();
+    let lch: Vec<Lch> = colormine
+        .iter()
+        .map(|x| Lch::from_color_unclamped(x.xyz))
+        .collect();
+
+    group.throughput(Throughput::Elements(colormine.len() as u64));
+
+    group.bench_with_input("xyz to lab", &colormine, |b, colormine| {
+        b.iter(|| {
+            for c in colormine {
+                black_box(Lab::from_color_unclamped(c.xyz));
+            }
+        })
+    });
+    group.bench_with_input("lch to lab", &lch, |b, lch| {
+        b.iter(|| {
+            for c in lch {
+                black_box(Lab::from_color_unclamped(*c));
+            }
+        })
+    });
+    group.bench_with_input("lab to lch", &lab, |b, lab| {
+        b.iter(|| {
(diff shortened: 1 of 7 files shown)


okaneco:
I wasn't sure if you wanted to keep that issue open or make a new general tracking feature for performance.

I was able to leverage the colormine data but it's a bit ugly. There's no easy way to access inside the tests folder from the benches if it's not exported in the lib. I exported the matrix module so I could use the functions in the bench, but I hid them from appearing in the docs.

palette/benches/cie.rs
```diff
@@ -0,0 +1,92 @@
...
+        .collect();
+
+    group.throughput(Throughput::Elements(colormine.len() as u64));
+
+    group.bench_with_input("xyz to lab", &colormine, |b, colormine| {
+        b.iter(|| {
+            colormine
+                .iter()
+                .map(|c| Lab::from_color_unclamped(c.xyz))
+                .collect::<Vec<_>>()
```
Ogeon, line 39: I believe the `Vec` creation should be either moved out or changed to use a loop and `black_box`, so we don't measure the time it takes to allocate and write to the `Vec` too. Maybe something like this for this one:

```rust
for color in &colormine {
  black_box(Lab::from_color_unclamped(c.xyz));
}
```

Ogeon:
Performance improvements are a never ending task as long as the library evolves, so let's not have a mega thread for it. I'm fine with closing the issue. It's better to discuss individual bottlenecks and improvements separately, as they are discovered. Getting the measuring part in place was the essential first step for everything else.

Really cool! I'll have to look into getting some continuous stats our of it. How long time does a full run take on your machine?

okaneco:
There doesn't seem to be a measurable performance difference with collecting the vec or not but I've changed it anyway because I agree. When I started, I was getting strange panics before I started collecting them. I assume they were being optimized out or something strange was happening because criterion would overflow.

It's slow. When you run cargo bench after any changes, you have to do a release build which takes about 17 seconds for me on my desktop. Then, there's a warmup period for each benchmark of 3 seconds and then at least 5 seconds to run the bench. There are 26 tests, 8 times 26 is 208 seconds for the full suite of benches that are inescapable plus overhead for the statistics.

It's bad enough when you're iterating on changes that you have to wait for the release build, so you end up running only the benches on code you've touched with something like `cargo bench -- "matrix"` if you want to iterate quickly.

Initial impressions are that most conversions are pretty stable, although spurious "regressions" do show up. The tests that seem most temperamental are the Hsv/Hsl tests. I assume because it's a lot of branching and dividing going on so there can be more latency occurring.

All that said, that's less time than it takes for a travis build to finish (if it finishes...) so it's not too bad.

Ogeon:
Alright, yes our time budget is pretty generous in that sense. And I suppose it has to take some time to get a somewhat stable result. So I guess we can expect around 10 minutes for it to run twice and compare, if it would be enabled for pull requests. Not the worst.

A cheap way to cut the time could be to run with only every second value, or fewer, if they are "granular" enough. And if that's the main reason for the long runtimes. It's something that can be experimented with later. I think it's good to get this in as it is now, to be able to start using it.

okaneco:
I can squash this then.

I wasn't sure if it was granular enough or it'd make sense to split them out more. Ideally there'd be some command to only rerun the benches you want through the action on smaller PRs.

Ogeon:
Yes, I think it's good to go. We can iterate upon it when it's merged. I'm sure there are things that can be improved and cases it's missing, but they will have to come over time. I think the most important part this is adding is a foundation for measuring.

bors r+

PR #181, opened 2020-04-19
Benchmark on pull request

### LINKED ISSUE #180: Continuous benching action
As discussed in #95, the available github actions seem to be
- https://github.com/rhysd/github-action-benchmark
- https://github.com/matchai/criterion-compare-action

continuous-benchmark only supports the unstable `cargo bench` at this time, not criterion. criterion-compare-action posts comments on the commit, not inside the PR. Since it's the only one that works for our needs, criterion-compare-action is probably worth experimenting with.

Caveat, Github Actions may be pointless for benching if they don't have consistent machines they use to execute.

Ogeon:
This adds an Actions workflow that runs benchmarks on pull requests.

Closes #180

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,15 @@
+on: [pull_request]
+name: Benchmark pull requests
+jobs:
+  runBenchmark:
+    name: Run benchmark
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Fetch master
+        run: git fetch --prune origin master
+      - name: Run benchmark
+        uses: matchai/criterion-compare-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CARGO_TARGET_DIR: ${{ format('{0}/target', github.workspace) }}


Ogeon:
Looks like it's _running_ now, but `master` doesn't have the necessary changes. I'll put them in a separate branch and get them in before merging this...

okaneco:
I've been peeking at this and it's frustrating but looks like you're almost there. I'm wondering if you can avoid some of the changes by `cd`ing into each branch's palette crate directory? I run the benchmarks from within palette and not the crate root.

Ogeon:
Possibly, if I add `cd` before the benchmarks run, but the inability to also run `cargo test` and `cargo build` from the root has been bothering me for a while. It's a mild inconvenience, but still something I fall for all the time. I wouldn't even have wanted to included the `#[no_std]` test crate in the workspace if it didn't seem to be necessary, so I don't worry too much about it being weird.

Ogeon:
Besides, it feels like it could end up being a pitfall in case any assumptions would change, or if the project structure changes to have benchmarks outside the `palette` directory. I don't want it to end up like the Travis file.

okaneco:
I'm not complaining, this should make things easier :smile: 

Workspaces have their oddities, it was especially confusing when I was starting out and getting unintended behavior.

When I was searching for the errors in the action, I came across other issues where they've deprecated or changed previous actions integrations.  It's good to make it more robust against assumptions of the future.

Ogeon:
It could have been an easy solution in this case. :+1: It looks like actions are still a bit of a moving target. I wonder where that `git checkout master` problem came from, for example.

Now, let's see if we can get some numbers out of this.

okaneco:
That master issue was perplexing. I saw some things about adding a `ref: master` under the job name in the script but I can't say for sure. Debugging CI things is very frustrating when it takes its time to fail. Now it looks like the CARGO_TARGET_DIR has to be set to `/github/workspace/palette/target/criterion/`.

Ogeon:
Hmm, why are they in there? I'm getting the same result on my machine, but that's not what https://doc.rust-lang.org/cargo/commands/cargo-bench.html says it should do by default. :thinking: Anyway, I think doing the opposite and setting it to what `critcmp` expects is what I want.

Ogeon:
Seem to be somewhat stable, considering it's running in a shared cloud environment. It's just that last one that went really off.

okaneco:
It's noisy but better than I expected. Probably enough to show overall trends.

Ogeon:
Yes, I think it's good enough to highlight large changes. I looked at how the they are using it for boa and they seem to have about the same noise level. Fluctuations within +/- 10 percentage points.

I'll clean this up a bit and try a couple of changes before merging it, but I think it looks good.

Ogeon:
That precision, though...

okaneco:
Not only that, the swings are massive with no rhyme or reason. Noise is probably around +/- 20%.

Ogeon:
It may have been a bad run, considering the PR column is quite consistently worse. The changes I'm making should have no effect on the result. We'll have at least two more samples in this PR. :slightly_smiling_face:

okaneco:
Ok, just a bad run it seems. Those runs will probably be filtered out over the course of commits/reviews to a PR. The best sanity check we can get for our local runs without dedicated hardware.

Ogeon:
Yes, looks like that was the case. Once again, it's better than nothing and it's more as a canary or sanity check, as you said, than exact measurements. :slightly_smiling_face: And dedicated hardware is not exactly in the budget at the moment. :smile:

It looks like the cache wasn't able to extract the files, so I'll just drop that one for now. Just to get this in place and be able to use it today, rather than later.

Ogeon:
Alright, that's it for this time. Note to self: rerun benchmarks if they look weird.

bors r+

PR #188, opened 2020-05-02
Allow HSV, HSL and HWB to represent nonlinear RGB

### LINKED ISSUE #160: Understanding when to use Srgba vs LinSrgba
Hello!

I am having trouble understanding a particular situation is it pertains to `Srgba` vs `LinSrgba`. I've read through https://github.com/Ogeon/palette/issues/133 and believe that I understand the general distinction between linear and perceptual RGB, but the following micro-example leaves me confused. I apologize if this is the wrong place to surface this.

This came about because I noticed that images I generated using code very similar to the [hue shift example](https://github.com/Ogeon/palette/blob/master/palette/examples/hue.rs) did not match reference images created using image editing software such as GIMP and Photoshop.

------

I have a source SRGB image, and I experiment by shifting its hue by 180 degrees in the HCL and LCH colorspaces, much like this [example](https://github.com/Ogeon/palette/blob/master/palette/examples/hue.rs). However, comparing the results to reference images I find that when dealing in HCL I need to treat the source image as though it were `Srgba` (expected), but when dealing in LCH I apparently need to treat the source image as though it were `LinSrgb` (unexpected). I am struggling with understanding the why of the difference.

In the source below, I'm using `Hsla` and `Srgba` to match the example. I swap these out for `Lcha` and/or `LinSrgba` as needed to generate the images below:
```
extern crate image;
extern crate palette;

use crate::palette::Saturate;
use palette::{Hue, Hsla, Lcha, Pixel, LinSrgba, Srgba};

fn main() {
    let mut image = image::open("colors.png").expect("Could not open image").to_rgba();

    for pixel in image.pixels_mut() {
        let color = Srgba::from_raw(&pixel.0).into_format();
        let adjusted = Hsla::from(color).shift_hue(180.0);
        pixel.0 = Srgba::from_linear(adjusted.into()).into_format().into_raw()
    }

    image.save("out.png");
}
```

These are the reference images: the original, and hue shifted using GIMP (Original / HSL / LCH):
Original colors-180-hsl-reference   colors-180-lch-reference

These images are HSL shifted. The left is using `LinSrgba`, the middle is the GIMP reference image, and the right one using `Srgba`. In this case, the left image appears to match the reference:
colors-180-hsl-linear colors-180-hsl-reference colors-180-hsl-srgb

These images are LCH shifted. The left is using `LinSrgba`, the middle is the GIMP reference image, and the right one using `Srgba`. In this case, the right image appears to match the reference:
colors-180-lch-linear colors-180-lch-reference colors-180-lch-srgb

This surprises me, because it seems to me that the source image is either linear or it is not, and I don't understand why the intermediary color space I work through would matter.

### LINKED ISSUE #187: Strange conversion results
Hi consider this code:
```rust
let t1 : palette::Hsla = color::Hsla::new(0.0,0.0,0.03,1.0);
println!("{:?}", palette::Srgba::from(t1));
```

The result we got is:
```rust
Alpha { color: Rgb { red: 0.18974826, green: 0.18974826, blue: 0.18974826, standard: PhantomData }, alpha: 1.0 }
```

Which is of course wrong. It doesn't look like there is a space error, as `Hsla` is defined with `Srgb` space. Moreover, palette is type safe, so I'm just cannot get what is wrong here. The correct result of converting `hsla(0,0,0.03,1.0)` is around `rgba(0.03, 0.03, 0.03)`. What I'm doing wrong here?

Ogeon:
This removes the old restriction that meant HSV, HSL and HWB can only represent linear RGB. They can now be converted to and from RGB (and each other) as long as the RGB standard is the same. This restriction is there to still allow type inference and reduce implementation complexity. So converting `Hsl<Srgb>` (not linear) to `Rgb<Srgb>` is still a single step process, while converting to `Rgb<Linear<Srgb>>` is a two step process. There are still shortcuts in place to only change the RGB standard, meaning it's possible to write `Hsl::<Linear<Srgb>>::from_color(Hsl::<Srgb>::new(...))`, similar to how it worked before.

This will mean that existing code may have a different output, adding more breaking changes to the pile.

Fixes #160, fixes #187

### DIFF
--- a/palette/src/hsl.rs
+++ b/palette/src/hsl.rs
@@ -27,7 +27,7 @@ pub type Hsla<S = Srgb, T = f32> = Alpha<Hsl<S, T>, T>;
 /// HSL color space.
 ///
 /// The HSL color space can be seen as a cylindrical version of
-/// [RGB](rgb/struct.LinRgb.html), where the `hue` is the angle around the color
+/// [RGB](rgb/struct.Rgb.html), where the `hue` is the angle around the color
 /// cylinder, the `saturation` is the distance from the center, and the
 /// `lightness` is the height from the bottom. Its composition makes it
 /// especially good for operations like changing green to red, making a color
--- a/palette/src/hsl.rs
+++ b/palette/src/hsl.rs
@@ -10,9 +10,9 @@ use rand::distributions::{Distribution, Standard};
 #[cfg(feature = "random")]
 use rand::Rng;
 
-use crate::convert::{FromColorUnclamped, IntoColorUnclamped};
+use crate::convert::FromColorUnclamped;
 use crate::encoding::pixel::RawPixel;
-use crate::encoding::{Linear, Srgb};
+use crate::encoding::Srgb;
 use crate::float::Float;
 use crate::rgb::{Rgb, RgbSpace, RgbStandard};
 use crate::{
@@ -24,7 +24,7 @@ use crate::{
 /// `Alpha`](struct.Alpha.html#Hsla).
 pub type Hsla<S = Srgb, T = f32> = Alpha<Hsl<S, T>, T>;
 
-/// Linear HSL color space.
+/// HSL color space.
 ///
 /// The HSL color space can be seen as a cylindrical version of
 /// [RGB](rgb/struct.LinRgb.html), where the `hue` is the angle around the color
@@ -39,16 +39,16 @@ pub type Hsla<S = Srgb, T = f32> = Alpha<Hsl<S, T>, T>;
 #[cfg_attr(feature = "serializing", derive(Serialize, Deserialize))]
 #[palette(
     palette_internal,
-    rgb_space = "S",
-    white_point = "S::WhitePoint",
+    rgb_standard = "S",
+    white_point = "<S::Space as RgbSpace>::WhitePoint",
     component = "T",
-    skip_derives(Xyz, Rgb, Hsv, Hsl)
+    skip_derives(Rgb, Hsv, Hsl)
 )]
 #[repr(C)]
 pub struct Hsl<S = Srgb, T = f32>
 where
     T: FloatComponent,
-    S: RgbSpace,
+    S: RgbStandard,
 {
     /// The hue of the color, in degrees. Decides if it's red, blue, purple,
     /// etc.
@@ -67,20 +67,20 @@ where
     /// is the sRGB standard.
     #[cfg_attr(feature = "serializing", serde(skip))]
     #[palette(unsafe_zero_sized)]
-    pub space: PhantomData<S>,
+    pub standard: PhantomData<S>,
 }
 
 impl<S, T> Copy for Hsl<S, T>
 where
     T: FloatComponent,
-    S: RgbSpace,
+    S: RgbStandard,
 {
 }
 
 impl<S, T> Clone for Hsl<S, T>
 where
     T: FloatComponent,
-    S: RgbSpace,
+    S: RgbStandard,
 {
     fn clone(&self) -> Hsl<S, T> {
         *self
@@ -97,7 +97,7 @@ where
             hue: hue.into(),
             saturation,
             lightness,
-            space: PhantomData,
+            standard: PhantomData,
         }
     }
 }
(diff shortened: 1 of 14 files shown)


okaneco:
This is a really nice change. :+1: 

Should a `TODO` note be added to the Rgb family benchmarks for implementing Hsl/Hsv/Hwb <-> Linear Hsl/Hsv/Hwb, and Linear Hsl/Hsv/Hwb <->LinSrgb? They're not completely necessary right now but it'd be good to let people know they're not covered currently.

Ogeon:
I don't know how interesting they are. Linear Hsl/Hsv/Hwb <->LinSrgb is the exact same as Hsl/Hsv/Hwb <-> Srgb, so that's already implicitly covered, and Hsl/Hsv/Hwb <-> Linear Hsl/Hsv/Hwb would just measure Srgb <-> LinSrgb plus the roundtrip that's hopefully optimized away. But I can add Hsl/Hsv/Hwb <-> Linear Hsl/Hsv/Hwb tomorrow if you think it would be interesting to keep an eye on.

okaneco:
I completely agree that they're not that interesting and are probably unnecessary. I meant it more for the sake of thoroughness. It can be hard to keep track of what the internal conversions are if you're looking in from the outside.

Ogeon:
True. It should be quick to add anyway.

Ogeon:
Wow, how do I even git?

Anyway, I added some benchmarks. Not all of them, for the reasons above.

palette/src/hsl.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
 /// `Alpha`](struct.Alpha.html#Hsla).
 pub type Hsla<S = Srgb, T = f32> = Alpha<Hsl<S, T>, T>;
 
-/// Linear HSL color space.
+/// HSL color space.
 ///
 /// The HSL color space can be seen as a cylindrical version of
 /// [RGB](rgb/struct.LinRgb.html), where the `hue` is the angle around the color
```
okaneco, line 30:
```suggestion
/// [RGB](rgb/struct.Rgb.html), where the `hue` is the angle around the color
```
These docs are currently broken in Hsl, Hsv, Hwb.

palette/src/hsv.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
 /// `Alpha`](struct.Alpha.html#Hsva).
 pub type Hsva<S = Srgb, T = f32> = Alpha<Hsv<S, T>, T>;
 
-/// Linear HSV color space.
+/// HSV color space.
 ///
 /// HSV is a cylindrical version of [RGB](rgb/struct.LinRgb.html) and it's very
```
okaneco, line 29:
```suggestion
/// HSV is a cylindrical version of [RGB](rgb/struct.Rgb.html) and it's very
```

palette/src/hwb.rs
```diff
@@ -24,10 +24,10 @@ use crate::{
 /// `Alpha`](struct.Alpha.html#Hwba).
 pub type Hwba<S = Srgb, T = f32> = Alpha<Hwb<S, T>, T>;
 
-/// Linear HWB color space.
+/// HWB color space.
 ///
 /// HWB is a cylindrical version of [RGB](rgb/struct.LinRgb.html) and it's very
```
okaneco, line 29:
```suggestion
/// HWB is a cylindrical version of [RGB](rgb/struct.Rgb.html) and it's very
```

palette/src/hsl.rs
Ogeon: Ah, but they should point to just `Rgb` after these changes, anyway, since none of them are necessarily based on linear RGB anymore. Good catch!

okaneco: Exactly :+1: 

What I also meant was that the current release's docs don't point anywhere  
https://docs.rs/palette/0.5.0/palette/rgb/struct.LinRgb.html

Ogeon: Yep, it will be two birds with one stone. Would be nice with a lint for it, but I guess we will just have to wait for intra-rustdoc links first.

Ogeon:
I think that's it for this one. Thanks for reviewing!

bors r+

PR #189, opened 2020-05-16
Correct scaling on random distribution of Lab/Lch

okaneco:
Add the proper scaling factor for random color distribution in Lab/Lch.

Lab/Lch brightness was being treated as 0-1 instead of 0-100.

### DIFF
--- a/palette/src/lab.rs
+++ b/palette/src/lab.rs
@@ -701,7 +701,7 @@ where
     // `a` and `b` both range from (-128.0, 127.0)
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Lab<Wp, T> {
         Lab {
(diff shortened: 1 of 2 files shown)


okaneco:
I finally got around to playing with the rng color feature. I was curious what the distributions looked like for each space and generated an image for each color. Something seemed off with Lab/Lch as you can see in the image. When I checked the code I realized a factor got overlooked or lost somewhere along the way in treating the range of Lab/Lch as 0.0 to 1.0 instead of 0.0 to 100.0. The other spaces seemed fine.

Left side is before the fix, right side is this PR. Top is Lab, bottom is Lch.

randfix

Ogeon:
Ah yes. Good catch!

bors r+

PR #190, opened 2020-11-19
Convert documentation to intra doc links, add default whitepoint for Lab/Lch, make code fixups

### LINKED ISSUE #177: Forgot default white point in alpha type aliases
`Lch` is a valid type, but `Lcha` is not. I have to specify the white point with `Lcha<white_point::D65>`. In the type alias definition, `type Lcha<Wp, T = f32> = Alpha<Lch<Wp, T>, T>;` should be changed to `type Lcha<Wp = D65, T = f32> = Alpha<Lch<Wp, T>, T>;` to remove that inconsistency.

I haven't checked all color types, but at least the `Laba` type suffers from this as well.

okaneco:
Add default D65 WhitePoint to Lab and Lch
Collapse if else statements in color_difference.rs
Use !is_empty instead of `len > 0` comparisons in gradient.rs
Use strip_prefix instead of manually stripping, avoids manual indexing
Remove extraneous clones, into_iters, and closures
Simplify matching on single condition into an if let
Remove format! in favor of to_string
Fix deprecated `image` functions in examples
Change some `as` casts to use direct `from` casts when applicable

Closes #177

### DIFF
--- a/palette/src/hsl.rs
+++ b/palette/src/hsl.rs
@@ -21,19 +21,19 @@ use crate::{
 };
 
 /// Linear HSL with an alpha component. See the [`Hsla` implementation in
-/// `Alpha`](struct.Alpha.html#Hsla).
+/// `Alpha`](crate::Alpha#Hsla).
 pub type Hsla<S = Srgb, T = f32> = Alpha<Hsl<S, T>, T>;
 
 /// HSL color space.
 ///
 /// The HSL color space can be seen as a cylindrical version of
-/// [RGB](rgb/struct.Rgb.html), where the `hue` is the angle around the color
+/// [RGB](crate::rgb::Rgb), where the `hue` is the angle around the color
 /// cylinder, the `saturation` is the distance from the center, and the
 /// `lightness` is the height from the bottom. Its composition makes it
 /// especially good for operations like changing green to red, making a color
 /// more gray, or making it darker.
 ///
-/// See [HSV](struct.Hsv.html) for a very similar color space, with brightness
+/// See [HSV](crate::Hsv) for a very similar color space, with brightness
 /// instead of lightness.
 #[derive(Debug, PartialEq, Pixel, FromColorUnclamped, WithAlpha)]
 #[cfg_attr(feature = "serializing", derive(Serialize, Deserialize))]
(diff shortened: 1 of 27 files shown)


okaneco:
I noticed the white point issue after submitting the PR, it was a quick fix so I force pushed. The commits are separated by the intra doc changes and fixups for easier review.

palette/src/hsl.rs
```diff
@@ -21,19 +21,19 @@ use crate::{
 };
 
 /// Linear HSL with an alpha component. See the [`Hsla` implementation in
-/// `Alpha`](struct.Alpha.html#Hsla).
+/// `Alpha`](crate::Alpha#Hsla).
```
Ogeon, line 24: Oh, do anchors work too? That would be amazing!

Ogeon (approved):
Oh, yes, this is great! I have been waiting for this feature for so long now. 😁 Time for even more cross-references in the future! And thanks a lot for the cleanup pass, too. Lots of nice changes there.

palette/src/hsl.rs
okaneco: Yes and no :sweat_smile:  
That's using the `span id=` on the `impl Alpha` block for `Hsl`, so anchors work but they don't seem to be so fine-grained to refer to what we want.
https://github.com/Ogeon/palette/blob/7b8272ee6c25c86960ff54e3766afe786078d0a0/palette/src/hsl.rs#L161

In the Alpha doc, the anchor for the Hsl implementation is `palette/struct.Alpha.html#impl-1`.

I tried searching briefly but couldn't find anything about this. It might need to be suggested as an issue/feature request for rustdoc. It'd be great to have the anchor be the type name instead of `impl-1`, which would mean no more markdown/html spans.

Otherwise, it's an amazing feature and makes cross-linking so much easier. When you use a path that doesn't exist, you get something like a lint warning but the doc build still goes through. I used `cargo doc --no-deps --no-default-features --features std` with an optional `--open` to quickly iterate on builds.

Ogeon: Alright, so it's basically the same old issue. :smile: I had even forgotten about it. It's good that they work at all! I don't know if it can be improved that much more in rustdoc, considering you can have multiple `impl` for types and type compositions. But I wouldn't complain if it did. :slightly_smiling_face:

Anyway, let's merge.

PR #196, opened 2021-01-01
Add uniformity tests for random colors and fix mistakes

Ogeon:
Adds tests that makes sure randomly generated colors are uniform within their "main" color space. I realized a Chi-squared goodness-of-fit test would do the trick, so I put this together over the past couple of days. Also found a few mistakes. I'm considering replacing the RNG with a fake RNG to avoid requiring as many samples, but that can wait.

The only space I didn't find a good test method for is `Lch`, which covers a circle. My attempts showed a slight bias towards the center, but I can't explain it. I ended up not looking further into it for now, since the test method I used (sampling only a central square) rejects too many samples anyway.

### DIFF
--- a/palette/src/random_sampling/mod.rs
+++ b/palette/src/random_sampling/mod.rs
@@ -1,3 +1,81 @@
 mod cone;
 
 pub use self::cone::*;
+
+#[cfg(test)]
+pub(crate) mod test_utils {
+    pub(crate) const BINS: usize = 10;
+    pub(crate) const SAMPLES: usize = 20_000;
+
+    /// Perform a Chi-squared goodness-of-fit test to check if the bins are
+    /// uniformly distributed. Returns the p-value.
+    pub(crate) fn uniform_distribution_test(bins: &[usize]) -> f64 {
+        let sum = bins.iter().sum::<usize>() as f64;
+        let expected = sum / bins.len() as f64;
+        let critical_value = bins
+            .iter()
+            .map(|&bin| {
+                let difference = bin as f64 - expected;
+                difference * difference / expected
+            })
+            .sum::<f64>();
+
+        chi_square(bins.len() - 1, critical_value)
+    }
+
+    // Shamelessly taken from https://www.codeproject.com/Articles/432194/How-to-Calculate-the-Chi-Squared-P-Value
+    fn chi_square(dof: usize, critical_value: f64) -> f64 {
+        if critical_value < 0.0 || dof < 1 {
+            return 0.0;
+        }
+        let k = dof as f64 * 0.5;
+        let x = critical_value * 0.5;
+        if dof == 2 {
+            return (-x).exp();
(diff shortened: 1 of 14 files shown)


palette/src/random_sampling/mod.rs
```diff
@@ -1,3 +1,81 @@
...
+            denom *= s;
+            sum += nom / denom;
+        }
+
+        sum * sc
+    }
+
+    fn approximate_gamma(z: f64) -> f64 {
+        const RECIP_E: f64 = 0.36787944117144232159552377016147; // RECIP_E = (E^-1) = (1.0 / E)
+        const TWOPI: f64 = 6.283185307179586476925286766559; // TWOPI = 2.0 * PI
```
okaneco, line 71: I know this is from that site but we have τ now which seems to include a few more digits
https://doc.rust-lang.org/stable/core/f64/consts/constant.TAU.html

okaneco:
Good catches on the hues. Is the problem with Lch related to using a chroma of 128 instead of the "max_extended_chroma" which is about 181? 

`rand` was recently updated to 0.8 but I haven't used it yet, I'm not sure if that makes sense to upgrade here or in another PR.
https://github.com/rust-random/rand/blob/master/CHANGELOG.md#080---2020-12-18

To use thread_rng in tests, I think you might have to add `rand` as a dev dependency and enable `std`. If upgrading to 0.8 then I'm not sure if the features have to be ["std", "std_rng"] or just "std_rng" or "std".

```toml
[dev-dependencies.rand]
version = "0.7"
default-features = false
features = ["std"]
```

palette/src/random_sampling/mod.rs
Ogeon: Good point, I might as well do that when fixing the rng.

Ogeon:
I tried to limit it to only check a and b within +/-80, which should be within the radius if I got it right, and it has a clear increase towards the middle of the histogram array. I don't think the max radius is the cause, but it would have mattered if I made it check the full Lab space. I also tried to add or remove sqrt calls to make sure that wasn't the cause... 🤔

I completely forgot that the thread rng would need the standard library. 🤦‍♂️ I might as well look into using a fake rng or something with a stable seed, and avoid having randomly failing tests. It's very rare with the current parameters, but not impossible. This test is quite sensitive to any noise in the numbers, so I had to crank up the number of samples and change the confidence.

Ogeon:
Oof, I just realized I only checked if a component is in range just before adding it to the statistics. That means if a is within +/-80, but b isn't, I would still count a and then reject b! That would explain the weird increase! 🤔🤔🤔

Ogeon:
Yep, that was the issue. It works perfectly now. I also changed it to use a Mersenne Twister with a predefined seed, allowing me to reduce the number of samples a lot. This should be good enough as a simple test to detect odd distributions.

Issue #194, opened 2020-12-13
Add an "anything" white point for agnostic CIE spaces


Ogeon:
Some color spaces, such as XYZ, doesn't depend on having an associated white point. It would be nice to be able to express that without losing the white point tracking. I would suggest adding an `Any` white point and a special trait (`OptionalWhitePoint`?) for white points that can be used in these spaces. The trait would be blanked implemented for any `T: WhitePoint` and `Any` would not implement `WhitePoint`, preventing it from be used in anything white point dependent. Converting to other spaces would require some concrete white point to be specified first.

PR #198, opened 2021-01-03
Finish ITU-R YCbCr encodings

HeroicKatora:
Adds the missing RGB spaces and transfers functions of issue #121 (BT.2020, BT.2100).

There are some quantization functions here as well but since they target 10-bit and 12-bit depths, which have no native type, there is no straightforward way of including them. They can be added separately based on the floating point signals, it's not a huge priority for the digital signals. 

For BT.2100 only the HLG (Hybrid Log-Gamma) transfer method has been added, and only its for scene-signal-EOTF. The  PQ-scene-signal (recording and editing) transfer, as well as the second method, hybrid log gamma (HLG), use an extended  model of the signal pipeline to modifying the signal to be appropriate for the viewing environment. In the former case no easily reversible encoding is given, while the second depends on many parameters. This is not supported by the `TransferFn` trait which does not permit such state. There are replacement values for when the viewing environment is unknown but I wanted to avoid the temptation to guess and rather get some feedback/directions/think about this some more. These would be required for producing display luminance only, a specific use case that's not part of the color editing pipeline anyways.

Closes: #121


palette/src/yuv/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+pub trait DifferenceFn {
+    /// The weights of the luminance transform.
+    ///
+    /// The linear transform is assumed to happen after the opto-electric transfer function is
+    /// applied to each color value. This is true for all ITU-R standards. Nevertheless, A
+    /// different form of encoding exists, called YcCbcCbr or constant luminance, which calculates
+    /// the luminance value from the linear RGB values instead to optimize the accuracy of its
+    /// result.
+    ///
+    /// The luminance weights correspond closely to the `Y` components of the `yxy`
```
okaneco, line 44:
```suggestion
    /// The luminance weights correspond closely to the `Y` components of the `Yxy`
```

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,149 @@
...
+impl<S: YuvStandard, T: Float> Copy for Yuv<S, T> {}
+
+impl<S: YuvStandard, T: Float> Clone for Yuv<S, T> {
+    fn clone(&self) -> Yuv<S, T> {
+        *self
+    }
+}
+
+impl<S: YuvStandard, T: Float> Yuv<S, T> {
+    /// Create an YUV color (in YCbCr order)
```
okaneco, line 54:
```suggestion
    /// Create a YUV color (in YCbCr order).
```

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Generic YUV.
+///
+/// YUV is an alternate representation for an RGB color space with a focus on separating luminance
+/// from chroma components.
+#[derive(Debug, PartialEq, FromColorUnclamped, Pixel)]
+#[cfg_attr(feature = "serializing", derive(Serialize, Deserialize))]
+#[palette(
+    palette_internal,
+    // rgb_standard = "S::RgbSpace",
```
Ogeon, line 20: You can remove this line while fixing other parts.

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,149 @@
...
+    white_point = "<S::RgbSpace as RgbSpace>::WhitePoint",
+    component = "T",
+    skip_derives(Luma, Rgb)
+)]
+#[repr(C)]
+pub struct Yuv<S: YuvStandard, T: Float = f32> {
```
Ogeon, line 26: This type could also be fleshed out with more traits and methods.

palette/src/yuv/ycbcr.rs
```diff
@@ -0,0 +1,34 @@
...
+)]
+#[repr(C)]
+pub struct YCbCr<S: YuvStandard, Q: QuantizationFn> {
+    /// The lumnance signal where `0.0f` is no light and `1.0f` means a maximum displayable amount
+    /// of all colors.
+    pub luminance: Q::Output,
+
+    /// The difference between overall brightness and the blue signal. This is centered around
+    /// `0.0f` but its maximum absolute value depends on the standard.
+    pub blue_diff: Q::Output,
```
Ogeon, line 24: Would it make sense to just shorten these to `blue` and `red`? Otherwise I feel like "diff" is such a common abbreviation that I could let it pass. Once again, arbitrary, but I'm conflicted.

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,149 @@
...
+where
+    T: Float,
+    S: YuvStandard,
+{
+    fn default() -> Yuv<S, T> {
+        Yuv::new(T::zero(), T::zero(), T::zero())
+    }
+}
+
+impl<S, T, St> From<Luma<St, T>> for Yuv<S, T>
```
Ogeon, line 96: You can remove these `From` implementations if they don't happen to be super common operations for this type. The promoted way of converting, starting with the next release, is via `FromColor` and `IntoColor`. If they should stay, they should be implemented using those traits.

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,149 @@
...
+impl<S, St, T> FromColorUnclamped<Luma<St, T>> for Yuv<S, T>
+where
+    S: YuvStandard,
+    T: FloatComponent,
+    St: LumaStandard<WhitePoint = <S::RgbSpace as RgbSpace>::WhitePoint>,
+{
+    fn from_color_unclamped(color: Luma<St, T>) -> Self {
+        let luma = color.into_linear();
+
+        Self::from(Rgb::<Linear<S::RgbSpace>, T> {
```
Ogeon, line 129: This is not the same shortcut as the `From` implementation uses... :thinking:

Ogeon (changes requested):
Welcome back! I think the implementations look nice and clean, so I have mostly just been commenting on nit picks. The only major improvements I would like to request, outside what's in the comments, are

* Symmetric conversion, if possible, to and from `Yuv` (now it's just to). You can also add it to the derive macro, to make it available everywhere. Also keep in mind that there is implicit clamping in `FromColor` and `IntoColor`, so you may want to change how the conversion is implemented. Let me know if the new model is unclear.
* I think I mentioned it earlier too, but unit tests (and examples in the documentation) helps a lot with verifying the code both now and later. Having them as examples helps explaining things too! It doesn't have to be super complicated. Just a few cases to make sure it's not off the rails, if there isn't any good test data source out there.

The more "first class citizens" these new types can be, the better! :grin:

For the transfer functions, would it help if the `TransferFn` trait could be split into a `From`/`Into` pair? Not in this PR, but in the future. It's something I have considered as a possible improvement, both there and elsewhere. Are there any other changes that could help with the other parts?

I think it's a good call to not draw the line where it's not too device or environment specific. Especially if it already covers a significant part of the process. I imagine some of it could be done dynamically instead, in a future iteration.

palette/src/yuv/yuv.rs
HeroicKatora: Whoops, I messed this up in `From<_>` assuming it was only for linear. I'll fix it and add some tests for it.

palette/src/yuv/ycbcr.rs
HeroicKatora: `blue` alone would be confusing. If `diff` is too generic then `blue_chroma` would be an alternative.

HeroicKatora:
Applied fixes for the style portion of the first round of review. One question on quantization. The implementation is currently not publicly exposed. When you say:

> Maybe it can be moved to the issue?

Do you mean moving it to a new issue that just deals with this part and has a references to a commit containing the files? Then they could be removed from the source tree for the moment.

palette/src/yuv/ycbcr.rs
Ogeon: Nah, leave it as it is. It's better than the alternatives.

Ogeon:
Nice, I will have another look later. But what I meant with "moving to the issue" was to add it to #121. I see now that you intend to close it with this PR. I'm not sure how you intend to proceed with the parts that are not done. I would like to have the code free from loose ends, so I'm trying to find a way to clean out anything that isn't ready to be used straight away.

I would like to have any future plans and TODOs (including their reference material) in issues and/or on a branch until it's implemented. Anything that's released should be useful in its current state. If everything is documented, sufficiently tested and gets the job done, it can be released at any time. That's the philosophy I'm following.

Ogeon:
But yeah, a new issue would be fine if you want to close the current one. If the temporary code is just a few snippets, it can just be added to the issue text. Having it in a commit sounds like a good idea if it's too much.

HeroicKatora:
> I'm not sure how you intend to proceed with the parts that are not done. 

The main portion is adding rgb color spaces, primaries, transfer functions, and YUV scaling as done here. The quantization is separate from that. In particular, full range quantization is a simple `round` of the `[0; 1]` clamped values to the proper number of binary digits or lossy IEEE conversion to 16-bit half-width floats, without any magic numbers. As such, I would proceed by removing all quantization related additions and view adding `quant.rs`, which contains the narrow range quantization for analog use, as a separate extension for later.

Ogeon:
Ok, that sounds like a good plan. And I'm sorry You can still create issues if you would like to have a discussion space for it. I'm interested to know if there are ways to make it easier to implement the quantization, so don't hesitate to reach out when you get back to those parts.

Ogeon:
I hope I didn't come forward as accusing earlier. I wanted to make sure the expectations are as clear as possible, so you know the background to my review comments.

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,203 @@
...
+        Self::new(luminance, red_diff, blue_diff)
+    }
+
+    fn from_rgb_internal<Sp>(rgb: Rgb<Linear<Sp>, T>) -> Self
+    where
+        T: FloatComponent,
+        Sp: RgbSpace<WhitePoint = <S::RgbSpace as RgbSpace>::WhitePoint>,
+    {
+        let xyz = Xyz::<<S::RgbSpace as RgbSpace>::WhitePoint, T>::from_color(rgb);
+        let rgb = Rgb::<(S::RgbSpace, S::TransferFn), T>::from_color(xyz);
```
Ogeon, line 101: You should probably use `from_color_unclamped` internally, unless clamping is always required. If it is required, it would be nice to have a comment with a short explanation. Just so it's not "fixed" in the future. The reason why is that certain applications want numbers outside the typical range to not get lost. That's why clamping only happens at the end of the conversion chain for other color spaces.

palette/src/yuv/yuv.rs
```diff
@@ -0,0 +1,203 @@
...
+
+impl<S, St, T> FromColorUnclamped<Rgb<St, T>> for Yuv<S, T>
+where
+    S: YuvStandard,
+    T: FloatComponent,
+    St: RgbStandard,
+    St::Space: RgbSpace<WhitePoint = <S::RgbSpace as RgbSpace>::WhitePoint>,
+{
+    fn from_color_unclamped(color: Rgb<St, T>) -> Self {
+        let linear = color.into_linear();
```
Ogeon, line 191: See the other comment about using this method internally.

Ogeon:
I should make some kind of checklist for new color spaces. They are far from as simple as they were. I forgot to mention a couple of more things earlier, that is good to know:

* Each color space has an alias for a version with an alpha channel. In this case the name would be `Yuva`. It can basically be copied and renamed from some other place, since they all follow the same pattern. But make sure to check that the documentation link works.
* It has to implement the `Limited` trait to get all of the conversion traits. It's where the clamping is done.

I would strongly suggest taking inspiration from both Rgb and Lab for the traits and methods.

HeroicKatora:
I have another potential problem: Linear RGB does not refer to any particular absolute luminance. Thus the conversion between HDR (bt.2020/2100) and non-HDR (everything else) color spaces won't work like written. The reference would be `1.0 = 100cd/m²` for SDR while HDR uses `1.0 = 10000cd/m²` as brightness.

Ogeon:
Right, yeah, it's not something I have thought of modelling as part of the type system. It could potentially be awrapper (like `Hdr<Something>`) that transparently implements the correct traits and let SDR be the default. Or have a `Luminance` trait for exposing that value. This is a bit outside my sphere of knowledge, so any input on what would be important to model and where is greatly appreciated. Is it tightly connected to certain standards or does it make sense to work with HDR and SDR versions of the same standard? I'm assuming it's not desirable to scale the values up or down too much when converting, since that would potentially cause precision errors.

HeroicKatora:
Worse than the slight difference introduces by multiplying/dividing through 100 a couple of times would be accidental clamping to the standard range, resulting in complete loss of high-dynamic range data. One possibility that I arrived at would be to represent this using the `Whitepoint`. This would be compatible with all existing data types as well as provide a distinct type. Then the HDR variant of the D65 whitepoint would be used in the new HDR color spaces.

```rust
/// Specify a high-dynamic range whitepoint.
struct Hdr<W>(PhantomData<W>);

impl Xyz<W> {
    /// Multiplies Y by 100, unclamped.
    pub fn to_hdr(self) -> Xyz<Hdr<W>> { … }
}

impl Xyz<Hdr<W>> {
    /// Multiplies Y by 100, unclamped.
    pub fn to_srd(self) -> Xyz<W> { … }
}
```

Ogeon:
That would be an interesting way of representing it. Then it would indeed be carried over between all spaces in a mostly transparent way. I suppose an alternative would be to embed it in the component type instead, so for example `Hdr<f32>`, but that may end up being unwieldy... Even if I feel like the semantics would perhaps be better. If I understand it correctly.

But go ahead and implement it with the white point, as an experiment if nothing else.

Ogeon:
Hmm, I realized this suggestion is going to cause trouble with some implementations that rely on specific white point types or properties. #194 what was made me think of it. It would have to special case `Hdr` somehow, since it's an orthogonal concept. But I have a new suggestion. How about having a `Hdr` wrapper for the whole color, similar to `Alpha`? It can blanket implement traits and operations the same way, and maybe (if we are lucky with the rules) even have `From` implementations for going to and from SDR. How does that sound as a compromise?

Ogeon:
Hey, it's been a while and I hope all is well! I'm closing this due to inactivity and because it's not something I'm expecting to be able to continue on my own in the near future. Feel free to come back to it some day (one way or another) if you get the motivation back!

PR #205, opened 2021-03-17
Generalizing gradients and add constant gradients

### LINKED ISSUE #62: Add convenience gradients
Just like you have named colors, it would be nice to have a collection of predefined gradients.

For reference:
http://matplotlib.org/xkcd/examples/color/colormaps_reference.html
https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/_cm.py

NicolasKlenert:
This pull request contains two commits:

### 1. Generalization of  `Gradient` 

This allows creation of gradients from arrays. The inner collection type only has to be `ArrayLike`. For backward compatibility the default is still vector. This allows constant gradients and should also allow gradients to be supported for the next `#[no-std]` build. Other options to enable `#[no-std]` for gradients were briefly disscused in #156.

### 2. Adding some constant gradients

This commit adds constant gradients. To be precise, it adds all 4 new matplotlib color gradients, which are perfectly perceptually-uniform, both in regular form and also when converted to black-and-white and therefore one of the most useful gradients. These are build the same way the named colors are built. A new feature-flag `named_gradients` is added to toggle said constants. This closes #62.

### Alterantives

- The generalization of gradients can be achieved in a multiple of ways. Using a trait is just one of them and may be the wrong way to go. 
   - However I think because this pull request doesn't have any breaking changes and gradients should be supporting arrays in future versions of the crate, it doesn't seem like this update of  `Gradient` will cause more breaking API changes in the future than otherwise. Also constant gradients may be the only interaction with gradients a user needs, such that introducing them could reduce the number of users which actually relies on the generation of `Gradient` itself.
- At the moment the 4 constant gradients are using linear interpolation but in nature these gradients are Spline-Interpolations of exaclty two points (and 2 controlpoints). If `Gradient` will support Spline-Inerpolation in the future and the exact controlpoints of these gradients can be found (I only found the colormaps), the gradients could be implemented more memory efficient.

### Remark

These commits depend on const generics, which is a feature in beta but is planned to be stable on 2021-03-25 onwords (see [#79135](https://github.com/rust-lang/rust/pull/79135)).


Ogeon:
Oh, wow! This looks really good! I'll of course have to let this rest here for a bit, at least until const generics are on stable, but I think the solution looks good and it's definitely a nice extension to the way it works now.

I just want to say that if there is anything you had in mind that would have been a breaking change, don't hesitate to let me know. Its current shape and form isn't necessarily the optimal solution (probably far from it). I have to admit that I even considered throwing it out at some point and just recommend finding a separate spline crate...

palette/src/gradient/mod.rs
```diff
@@ -97,6 +129,16 @@ impl<C: Mix + Clone> Gradient<C> {
...
     }
 
+    /// Create a gradient of colors with custom spacing and domain. There must
+    /// be at least one color and they are expected to be ordered by their
+    /// position value.
+    pub fn with_domain(colors: T) -> Gradient<C, T> {
+        assert!(!colors.is_empty());
+
+        //Maybe sort the colors?
+        Gradient(colors)
```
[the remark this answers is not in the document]
Ogeon: It could be nice, but I'm not sure what's the better option. On one hand it removes panics from the code, on the other hand it pushes them over to the user's code instead. It's not a complicated validation so I'm still leaning towards not returning an `Option` and sparing people the need to always check it themselves.

palette/build/named.rs
```diff
@@ -40,7 +47,56 @@ pub fn build() {
...
+        //every line should have the same info: name type number_of_colors [\n red green blue]^number_of_colors
+        let name = parts.next().expect("couldn't get the color name");
+        let color_type = parts.next().expect("couldn't get the type of the colors");
+        //we assume that color_type is a rgb type
+        let color_type = format!("crate::rgb::{}", color_type);
+        let number_of_colors : usize = parts.next().expect("couldn't get the number of colors")
+            .parse().unwrap_or_else(|_| panic!("couldn't parse the number of colors for color {}", name));
+        writeln!(writer, "/// New matplotlib colormap by Nathaniel J. Smith, Stefan van der Walt, and (in the case of viridis) Eric Firing.").unwrap();
+        writeln!(writer, "/// This gradient is perfectly perceptually-uniform, both in regular form and also when converted to black-and-white.").unwrap();
+        //writeln!(writer, "/// The colormap is released under the CC0 license public domain dedication.").unwrap();
```
Ogeon, line 74: This should probably not be commented out. We can also link to an official source if possible.

palette/src/gradient/mod.rs
```diff
@@ -12,6 +13,61 @@ use crate::float::Float;
 use crate::Mix;
 use crate::{from_f64, FromF64};
 
+#[cfg(feature = "named_gradients")]
+pub mod named;
+
+/// Types which represent an ordered collection of known size
+pub trait ArrayLike : Index<usize>{
```
Ogeon, line 20: After looking at this again, it seems to me like we could replace it with `AsRef<[T]>` or `AsMut<[T]>`. That would remove the const generics dependency and lower the required Rust version.

Ogeon (commented):
Ok, so I have taken another look at this, just to check if I missed anything, and added a couple of comments. I'm thinking we can merge it as it is after those have been addressed and think about any breaking changes as a larger makeover effort.

I have restarted the build to see what it says, but it looks like it's going to fail on nightly due to this:

```text
error: derive helper attribute is used before it is introduced

   --> palette/src/convert.rs:535:7

    |

535 |     #[palette(

    |       ^^^^^^^

...

543 |     #[derive(FromColorUnclamped, WithAlpha)]

    |              ------------------ the attribute is introduced here

    |

    = note: `-D legacy-derive-helpers` implied by `-D warnings`

    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

    = note: for more information, see issue #79202 <https://github.com/rust-lang/rust/issues/79202>

error: derive helper attribute is used before it is introduced

   --> palette/src/convert.rs:604:7

    |

604 |     #[palette(

    |       ^^^^^^^

...

612 |     #[derive(Copy, Clone, FromColorUnclamped, WithAlpha)]

    |                           ------------------ the attribute is introduced here

    |

    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

    = note: for more information, see issue #79202 <https://github.com/rust-lang/rust/issues/79202>

error: aborting due to 2 previous errors
```

Would you mind going in and just reordering those attributes?

NicolasKlenert:
> Would you mind going in and just reordering those attributes?

I don't mind, however if I'm not mistaken, these errors don't have anything to do with this PR. This error was previously a warning and already existed in the crate. Such a seperate commit would be more idomatic.

palette/build/named.rs
NicolasKlenert: I commented it out because I tried to be in line with the named colors, which do not include any license in the documentation of each color but do include the source in the documentation of the named module. Such I also included the source in the documentation of the `gradient/named` module.

How and where to include credits and licenses if of course up to you. So my questions are these:
- What should I include in the documentation of the named module?
- What should I include in the documentation of each gradient?

Things possible to include would be:
- License
- Content of the license
- Source
- Authors / Creators

palette/src/gradient/mod.rs
NicolasKlenert: I thought about using `AsRef<[T]>`  instead of creating a custom trait. You already mentioned the advantages. The disadvantages would be to have to call `as_ref()` everytime we would want to use the collections, we can't know if `Gradient` would own it's collection and it would be less generic, because the underlying data would need to act like a slice (so generators would be difficult to implement). These are all minor, so to be honest I'm not sure why I decided against it.

I will try to replace the custom trait with `AsRef<[T]>` in the next few days and comment here again afterwards.

Ogeon:
> I don't mind, however if I'm not mistaken, these errors don't have anything to do with this PR. This error was previously a warning and already existed in the crate. Such a seperate commit would be more idomatic.

It's a surprise change from the nightly channel, so giving it its own commit sounds perfect! 🙂 I'm most grateful for the help!

palette/src/gradient/mod.rs
Ogeon: It would be great if you can give it a try! 👍 Calling `as_ref` is probably not a big deal compared to maintaining a custom trait, to be honest, so I prefer `AsRef<T>` if it works. It's usually only necessary to do it once per function, though.

You are right that generators will be excluded in this case, but a workaround would be to collect values in an array or `Vec` and pass that in. As a slice, even, to save some allocations.

palette/build/named.rs
Ogeon: Welcome to software licensing. 😬 I think you can just uncomment the line and it will be fine. We are then mentioning the authors and the license, and covering our backs even though it's public domain. Here are the guidelines for a CC license that forces attribution: https://creativecommons.org/use-remix/attribution/ Public domain doesn't require that or anything else as I understand it.

palette/src/gradient/mod.rs
NicolasKlenert: The latest commit removes the custom trait in favor of `AsRef<[T]>`. To make it work, I added `Phantomdata<Color>` to the Gradient. Other than that, all changes were straightforward.

Ogeon:
It looks great with the new changes, thank you!

Ogeon:
Alright, I'm assuming this is good to go! Thank you and thanks for adding in the attribute fix! This is a good step towards a better gradient.

bors r+

PR #210, opened 2021-04-04
Rename Limited trait to Clamp

### LINKED ISSUE #209: Rename Limited to Clamp
I think "Limited" sounds like there will be "limited" and "unlimited" color spaces, but the reality is not as clear cut. It's more that most of them have a typical range. I think it's better to rebrand the trait to an "action" trait, by naming it with a verb, and letting the details be implementation specific. So it will still make sense to have a clamp method even if a color space is only limited by convention, but still could have a theoretically infinite range.

What to do:
 * Rename the trait
 * Rename `is_valid` to `is_within_bounds` or something more appropriate
 * Update documentation and examples

okaneco:
Rename `is_valid` to `is_within_bounds`
Update the `assert_ranges` macro to use `clamped` instead of `limited`
Update relevant documentation and examples

Closes #209

### DIFF
--- a/palette/src/alpha/alpha.rs
+++ b/palette/src/alpha/alpha.rs
@@ -14,7 +14,7 @@ use crate::convert::{FromColorUnclamped, IntoColorUnclamped};
 use crate::encoding::pixel::RawPixel;
 use crate::float::Float;
 use crate::{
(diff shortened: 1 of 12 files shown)


Ogeon:
Oh, excellent! I wasn't expecting such a quick response, but it's a most welcome one! And thanks for going the extra mile with the `assert_ranges` macro.  This looks good, so let's merge!

bors r+

PR #216, opened 2021-04-11
Add doc alias, doc cleanups, remove trait from Packed struct

okaneco:
Add "hsb" alias for Hsv
Add "linear" for LinLuma/a and LinSrgb/a
Add "wcag" for RelativeContrast trait
Add "xyY" for Yxy
Add "gray/grey" for Luma
Make rgb::FromHexError public and document it
Clean up document example in RelativeContrast
Remove RgbChannels trait from Packed struct

### DIFF
--- a/palette/src/rgb/packed.rs
+++ b/palette/src/rgb/packed.rs
@@ -12,10 +12,9 @@ use crate::Pixel;
 /// express each value of the Red, Green, Blue, and Alpha components in the
 /// RGBA color.
 ///
-/// Note that conversion from float to integer component types in `palette` maps
-/// the floating point value to the integer's value range and then rounds the
-/// result before casting to the integer type. An example of the difference in
-/// conversion methods is shown below.
+/// Note that conversion from float to integer component types in `palette`
+/// rounds to nearest even: an `Rgb` component of `0.5` will convert to
+/// `0x80`/`128`, not `0x7F`/`127`.
 ///
 /// ```
 /// use approx::assert_relative_eq;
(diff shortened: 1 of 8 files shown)


okaneco:
I added a few doc aliases that seemed appropriate and did some small fixups in docs. The aliases mak searching a bit more ergonomic, the search is surprisingly fuzzy but also doesn't surface as much as I'd probably want based on the documentation text.

I know people can tend to copy/paste code straight from docs so I wanted to change the relative contrast example to show `From<u32>` exists on `Srgb` instead of having to convert from a string -- this was written before that impl existed. In the process, I realized that the error wasn't available outside of the crate so I think it should be made `pub` since that's what is returned in `from_str`.

palette/src/rgb/packed.rs
```diff
@@ -25,9 +24,9 @@ use crate::Pixel;
 /// let packed: Packed = Srgb::new(0.5, 0.0, 0.5).into_format().into();
 /// assert_eq!(0xFF80_0080, packed.color);
 ///
-/// let unpacked: Srgba<u8> = Packed::<Rgba>::from(0x80FF_FF80).into();
+/// let unpacked: Srgba<u8> = Packed::<Rgba>::from(0xFFFF_FF80).into();
 /// assert_relative_eq!(
-///     Srgba::new(0.5, 1.0, 1.0, 0.5),
+///     Srgba::new(1.0, 1.0, 1.0, 0.5),
```
Ogeon, line 29: I just want to say that I appreciate that you spotted this. Nice to remove the ambiguity!

Ogeon:
This is a good set of aliases and some nice cleanup! Thank you!

bors r+

PR #219, opened 2021-05-07
Feature/hsluv

masonium:
Only 2.5 years late :-P

This commit adds support for the HSLuv color format. The implementation is based on conversion from `Luv`, which is the most direct conversion in the reference implementation.

I also include the test suite from the reference implementation.


palette/src/hsluv.rs
```diff
@@ -0,0 +1,689 @@
...
+use crate::{encoding::pixel::RawPixel, white_point::D65};
+
+/// `HSLuv` with alpha an alpha component. See the [`Hsluva`
+/// implementation in `Alpha`](crate::Alpha#Hsluv).
+pub type Hsluva<Wp, T> = Alpha<Hsluv<Wp, T>, T>;
+
+/// The HSLuv color space.
+///
+/// [HSLuv](https://www.hsluv.org/) is designed as a "human-friendly
+/// alternative to HSL". It is an extension of the CIELuv / CIELch
```
Artoria2e5, line 28: Should probably be CIE Lch(uv).

palette/src/hsluv.rs
```diff
@@ -0,0 +1,689 @@
...
+}
+
+///<span id="Hsluva"></span>[`Hsluva`](crate::Hsluva) implementations.
+impl<T, A> Alpha<Hsluv<D65, T>, A>
+where
+    T: FloatComponent,
+    A: Component,
+{
+    /// CIE L\*C\*h° and transparency with white point D65.
+    pub fn new<H: Into<LabHue<T>>>(hue: H, saturation: T, l: T, alpha: A) -> Self {
```
Artoria2e5, line 193: HSLuv is supposed to be based on Luv! I think what you are making here is HSLab instead.

palette/src/lch.rs
```diff
@@ -136,6 +136,18 @@ where
         T::zero()
     }
 
+    /// Return the maximum `chroma` value that is representable for
+    /// the given lightness and hue values.
+    pub fn max_chroma_for_lh(&self) -> T {
```
Artoria2e5, line 141: Again, the bounds are for l and h defined in CIELch(uv), which is different from the CIELch(ab) of this file.

Artoria2e5 (changes requested):
Anyway, the code is nice, but it's a bit wrong. I think you will need to:
* Add `Luv<Wp, T>` from `Xyz` (basically take the `Lab` code and change the matrix and non-linearity)
* Add `Lchuv<Wp, T>` and `LuvHue<T>` based on the above (take `Lch` and... you know)
* Replace the references in `LuvBounds` and `Hsluv<Wp, T>`  to `LabHue<T>` with `LuvHue<T>`
* Remove the bounds stuff in `Lch<Wp, T>` and put it in `Lchuv<Wp, T>`

Ogeon (commented):
Thanks for coming back and picking this up! Better late than never, as they say! 😄 It looks good overall, so most of my improvement requests are details. However, @Artoria2e5 got it right regarding which LCH this has to be based on. It's easy to miss and I made the same mistake at first. This opens a can of worms, since we need two additional color spaces (seem to be the simplest solution), but they should have a lot in common with `Lab` and `Lch`. Apart from that and what I have commented on directly in the code, here are a couple of more things I think can be improved:

 * Random sampling is a bit tricky here, but I think it can use the same method as for HSL, only re-scaled to the correct ranges. I mean, considering it's supposed to be a better HSL.
 * I think the boundary methods could be free functions with `pub(crate)` in the `hsluv` module, unless they happen to be useful outside our internal implementations.

palette/src/hsluv.rs
Ogeon: While correct, it seems like your comment got misplaced.

masonium:
Thank you for the feedback! There's definitely a bit of learning as I go, and I sense something wasn't quite right. (Your test suggestion is a good, @Ogeon, as the conversion to/from `XYZ` fails.) I'll give it a second pass with strategy outlined by @Artoria2e5 (`Luv` -> `Lchuv` and `LuvHue` -> `Hsluv`).

masonium:
Actually, since this is going to be even larger than scope. I'm going to close this pull request and resubmit with multiple pull requests that break down the steps above. 

1. Implement `Luv` and `LuvHue`.
2. Implement `Lchuv`
3. Implement 'Hsluv` properly based on the `Luv`/`Lchuv` spaces.

Artoria2e5:
> @Ogeon they should have a lot in common with Lab and Lch

I think there is something to do about this, maybe in `palette_derive`... The hue-angle-blending code in h\*\* and lch can largely be shared too.

PR #221, opened 2021-05-09
Feature/cieluv

masonium:
Implements the CIELUV color space, translating directly to and from `Xyz`.

First step towards `Hsluv` implementation.


palette/src/lch.rs
```diff
@@ -8,7 +8,7 @@ use rand::distributions::{Distribution, Standard};
 #[cfg(feature = "random")]
 use rand::Rng;
 
-use crate::color_difference::ColorDifference;
+use crate::{color_difference::ColorDifference};
```
Ogeon, line 11: You can revert this change while fixing other parts.

palette/src/luv.rs
```diff
@@ -0,0 +1,857 @@
...
+    Wp: WhitePoint,
+{
+    /// L\* is the lightness of the color. 0.0 gives absolute black and 100
+    /// give the brightest white.
+    pub l: T,
+
+    /// u* goes from -100 to 100.
+    pub u: T,
+
+    /// v* goes from -100 to 100.
```
Ogeon, line 54: It's probably good to escape the `*` for good measure, to avoid accidental emphasis formatting.

palette/src/luv.rs
```diff
@@ -0,0 +1,857 @@
...
+    }
+
+    fn clamp(&self) -> Luv<Wp, T> {
+	let mut c = *self;
+	c.clamp_self();
+	c
+    }
+
+    fn clamp_self(&mut self) {
+	self.l = clamp(self.l, T::zero(), from_f64(100.0));
```
Ogeon, line 306: You could use `{min,max}_l()` here too.

palette/tests/hsluv_dataset/hsluv_dataset.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use palette::convert::IntoColorUnclamped;
+use palette::white_point::D65;
+use palette::{Lch, Luv, Xyz};
+use std::collections::HashMap;
+
+#[derive(Clone, Debug)]
+struct HsluvExample {
+    name: String,
+    lch: Lch<D65, f64>,
```
Ogeon, line 16: Is this the correct LCH? 😬

palette/src/luv.rs
```diff
@@ -0,0 +1,857 @@
...
+	let prime_denom_recip = prime_denom.recip();
+	let prime_ref_denom_recip = (w.x + from_f64(15.0) * w.y + from_f64(3.0) * w.z).recip();
+
+	let u_prime: T = from_f64(4.0) * color.x * prime_denom_recip;
+	let u_ref_prime = from_f64(4.0) * w.x * prime_ref_denom_recip;
+	dbg!(u_ref_prime.to_f64().unwrap());
+
+	let v_prime: T = from_f64(9.0) * color.y * prime_denom_recip;
+	let v_ref_prime = from_f64(9.0) * w.y * prime_ref_denom_recip;
+	dbg!(v_ref_prime.to_f64().unwrap());
```
Ogeon, line 245: These `dbg` lines should be removed.

Ogeon (commented):
Nice! First step! I have only nit picks, really, except the `dbg` lines that got into the commit. They are also why the tests are failing. I also noticed the indentation was a bit mixed, so please run this through `rustfmt` as well. 🙂 Well done!

palette/tests/hsluv_dataset/hsluv_dataset.rs
masonium: I believe this is the appropriate LCH, but I'll remove it until I actually implement LCH in the next round so it's actually tested.

masonium:
There are a few other files that `cargo fmt` modifies when running it over the whole repo (`palette/build/named.rs`, `palette/examples/color_scheme.rs`, `palette/src/named.rs`). I didn't include those with my formatting commit because they are technically unrelated, but I can just add them if you like.

palette/tests/hsluv_dataset/hsluv_dataset.rs
masonium: Silly me. That's *definitely* not the right `Lch`. I started working on `Lchuv` in a new branch on top of this one, and mentally I've already changed this. But of course it's not right in this branch yet 🤦🏾

Ogeon: Parallel universes can be confusing like that. 😁

Ogeon:
No need to include the additional files. I think this is ready for a merge. Thanks!

bors r+

Issue #107, opened 2018-06-09
Add missing conversion tests for Lch and Lab


Ogeon:
There are no tests for conversion from `Lch` and `Lab` to other spaces in the test suite in `palette/tests/convert`. They were not added, due to problems with scaling and hue, but the scaling should be correct now, so I hope it should be possible to at least add some tests.

CGUTA:
Hello @Ogeon

I would like to give this a go if that is ok with you?

I see that the data is already in colormine.csv and that the patterns will be similar to the tests for the other colors as far as i can tell.

Ogeon:
Sure go ahead! Let me know if you run into any problems.

CGUTA:
The conversion from xyz to lab goes ok for colormine data if the epsilon is increased to 0.1
The conversion for yxy needs a bigger epsilon between 5.0 (fail) and 10.0 (pass).

With a CIExyY input:

|Y | x | y|
| - | - | - |
|92.88 | 0.3 | 0.32|

And using the colormine csv as gold standard for the CIE Lab these are the results using palette and the colormine converter http://colormine.org/convert/lab-to-yxy (converts both ways):

|color_data|L|a|b|RMSE|
|-|-|-|-|-|
colormine_csv|97,18|-1.34|-4.27|0.0
colormine_web|97.01|-2.75|-4.40|0.61|
palette|97.17|2.22|-5.72|0.95|

To give an idea of Epsilon vs RMSE the from XYZ conversion generated a Lab result that needs an epsilon of 0.1 to pass. This has a RMSE (root mean squared error) of 0.004.

From this I conclude that is a problematic conversion also for the colormine converters so it should not be too uncareful to increase the epsilon to make the test pass.

PR #223, opened 2021-05-10
Feature/lchuv

masonium:
Round two! This branch implements the CIELCHuv color space.

I also cleaned up the `Uniform*Hue` implementation in `hues.rs` to use a macro to save on some repetitive typing.


palette/tests/hsluv_dataset/hsluv_dataset.rs
```diff
@@ -72,3 +80,22 @@ pub fn run_luv_to_xyz_tests() {
...
+        let to_luv: Luv<D65, f64> = v.lchuv.into_color_unclamped();
+        assert_relative_eq!(to_luv, v.luv, epsilon = 0.1);
+    }
+}
+
+#[test]
+pub fn run_luv_to_lchuv_tests() {
+    for (_, v) in TEST_DATA.iter() {
+        let mut to_lchuv: Lchuv<D65, f64> = v.luv.into_color_unclamped();
+        if to_lchuv.chroma < 1e-8 {
```
masonium, line 96: This is a worth a comment.
In the implementation that generates the examples, `chroma` values below a certain threshold are a treated as 0.0, so the hue is set as 0 rather than using whatever comes out of `atan` (and probably isn't all that meaningful). My implementation matches the existing `Lab` -> `Luv` implementation and only does this if `u*` and `v*` are *strictly* 0. I can change my implementation to use a threshold on chroma instead, or we can leave this tweak in place in the test.

palette/src/hues.rs
```diff
@@ -20,7 +20,7 @@ macro_rules! make_hues {
         /// number (like `f32`). This makes many calculations easier, but may
         /// also have some surprising effects if it's expected to act as a
         /// linear number.
-        #[derive(Clone, Copy, Debug, Default)]
+            #[derive(Clone, Copy, Debug, Default)]
```
masonium, line 23: `cargo fmt` does weird indenting here.

masonium:
The benchmark is failing for some pretty inscrutable reason, even though the CI build passed.

palette/tests/hsluv_dataset/hsluv_dataset.rs
Ogeon: There's going to be a point where the hue is almost just noise from the floating point operations. This was mentioned just yesterday in #107. I'm ok with a cut-off point like this.

palette/src/hues.rs
Ogeon: Indeed, what is this? Maybe some bug. You can try to shuffle the attributes around, but if this is what it want to do, then so be it. 🤷‍♂️

Ogeon: I saw some other odd thing further down, so maybe it's not a big fan of macros at the moment.

masonium: Playing around with it more, it seems to just be not opinionated inside the macro. If I manually correct the indentation, `rustfmt` doesn't "fix" it.

palette/tests/hsluv_dataset/hsluv_dataset.rs
masonium: To be clear, do you want the cut-off point in the code itself for hue calculation, or just the tests?

palette/src/hues.rs
Ogeon: I would expect it to not want to do too much with macro code, considering it's much more free form. 🤔 Just do your best to make it look nice.

Ogeon:
You can ignore the benchmarks they are not working for PRs from forks at the moment.

palette/tests/hsluv_dataset/hsluv_dataset.rs
Ogeon: Ah, yes, sorry. Let's have it only in the test for now.

Ogeon:
Hah, simple but effective. And the implementation looks great! Thank you!

bors r+

PR #224, opened 2021-05-11
add hue forgiving tests for CIEL*a*b* and CIELch color spaces

### LINKED ISSUE #107: Add missing conversion tests for Lch and Lab

CGUTA:
Tests for lab and lch were added.

Results of tests were not precise enough for epsilon of 0.05.
A more lenient epsilon was set at 7.0.

The biggest imprecision originated from hues calculated for colors with low chroma in Lch.
Hues are now ignored for testing if chroma is smaller than 20 only for lch.

Some tests require color parameters not present in the original colormine.csv so they were added to prevent multiplication corrections when instatiating the colors used in tests.

Closes #107


CGUTA:
an epsilon of 0.1 and `convert::color_mine_from_xyz` passes
but `convert::color_mine_from_yxy` is not happy until an epsilon of 10.0.

I will try to investigate more about the colormine data and yxy bcs im not familiar with that colorspace.

CGUTA:
Fixed the epsilon to make them pass.
You can find the resoning for this at #107

Ogeon:
Hi, sorry I didn't get the time to give this a proper look yesterday. But it's good to see that you found a solution! You are probably right about the epsilon. I wonder if `f64` would have been better in this case, since we want to check that it comes as close as possible. I know there are situations where `f32` simply isn't good enough.

CGUTA:
Hello, no problem!
I added now for lch. I also got some unpassed tests and thought it may be because palette expects the h on the range of (-180,180) normalized and the input was not.

But still after that sanity check I saw some texts that dont pass because there seem to be some problems with the hue and the conversion comes out too far.

Unfortunately i dont know what is causing it so I dont know how could it be solved. Any ideas?

```
thread 'convert::color_mine_from_lab' panicked at 'assert_relative_eq!(src.lch, tgt.lch, epsilon = 0.1)

    left  = Lch { l: 69.24, chroma: 0.01, hue: LabHue(-90.0), white_point: PhantomData }
    right = Lch { l: 69.24, chroma: 0.01, hue: LabHue(-63.19), white_point: PhantomData }

', palette/tests/convert/data_color_mine.rs:181:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- convert::color_mine_from_linear_rgb stdout ----
thread 'convert::color_mine_from_linear_rgb' panicked at 'assert_relative_eq!(src.lch, tgt.lch, epsilon = 0.1)

    left  = Lch { l: 97.17865, chroma: 4.471093, hue: LabHue(-107.55552), white_point: PhantomData }
    right = Lch { l: 97.18, chroma: 4.48, hue: LabHue(-107.45), white_point: PhantomData }
```

CGUTA:
I checked with another color conversion library and it agrees with palette for the first example (69.24). I wonder if the colormine data is 100% trustable.

Ogeon:
It's an old test (from 2016), so things may have changed, but I found the source and colormine seem to use doubles (so f64 in our case) in their implementation. It may matter, as mentioned above, so we should probably switch to it for these tests anyway. I also found their definition of being "close enough": https://github.com/hvalidi/ColorMine/blob/1df40a84376430c77f49f580379b3327c1c4bde8/ColorMine.Test/ColorSpaces/ColorSpaceTest.cs#L83 Basically, the difference should be within 1% of the expected value. 🤔 But I haven't managed to find their data source.

CGUTA:
I will try to instatiate everything as `f64` and see if it improves.

Thanks for the link, I can see that they use a 1% numerical difference regrardless of the component.

For my understanding. How is an epsilon calculated in palette between colors that may have different number of compoents? I had trouble finding the code but I probably I am missing it.

Carlos

Ogeon:
Most of the epsilon stuff happens in the [approx](https://crates.io/crates/approx) crate. We just take the default epsilon from the component type (f32 or f64 in this case). But the f64 didn't seem to improve much. I think we could make an exception for the hue when chroma is near zero (we have already done it for the *uv types in parallel with this PR), since it's going to be all over the place. Otherwise I guess it's either a matter of double checking the data or raising the epsilon a bit.

CGUTA:
solves most of problems but yxy conversion still needs bigger epsilon to pass for `lch`, hue is not the problem anymore.
all pass with a epsilon of 10 if that is ok then it would be complete.

palette/tests/convert/data_color_mine.rs
```diff
@@ -165,9 +178,19 @@ fn check_equal_cie(src: &ColorMine, tgt: &ColorMine) {
...
+
     // hue values are not passing for from_yxy conversion. Check github #48 for
     // more information assert_relative_eq!(src.lch.hue, tgt.lch.hue, epsilon =
     // 0.05);
+    //low chroma colors have hues that are hard to calculate precisely
+    
+    if src.lch.chroma > 20.0 {
+        println!("{}", src.lch.chroma);
+        assert_relative_eq!(src.lch, tgt.lch, epsilon = 10.0);
+    }
```
Ogeon, line 192: We could set hue to 0 (or the expected value) for low chroma cases to sort of "ignore" it. That way the remaining components will still be checked.

Ogeon:
Ok, sounds good to sort of just consider this as a ballpark check. And maybe the errors are amplified with the larger values as well. Anyway, I added a small comment, but I think the rest looks really good and I'm prepared to merge after it as been addressed!

CGUTA:
Done, let me know if you need me to do something extra for the merge.

Ogeon:
Nice, thank you! I forgot to check for it before, so thanks for reminding me, but it would be nice if you can edit the PR description (a.k.a. the initial comment) to describe the change, and also have a "closes #XYZ" line. The title and description are used as the merge commit message, so that's why I'm a bit picky about it. 🙂 You could also squash the commits while you are at it, there have been some changes back and forth. It just cleans it up a bit.

Ogeon:
Nice, thank you!

bors r+

PR #225, opened 2021-05-14
`Hsluv` implementation

### LINKED ISSUE #112: HSLuv support
[HSLuv](http://www.hsluv.org/) is a more intuitive version of CIELAB Lch, designed for programmatic color generation. I would be happy to attempt an implementation, if you're interested in adding it to `palette`.

masonium:
Round three!

This finally implements the HSLuv color space as `Hsluv`. A few notes.

- I didn't implement Uniform generation for `Hsluv` because I wasn't convinced that the naive version would be anything close to correct. I think you could do something like rejection sample in `Lchuv` (based on if, after converting that to Xyz, it were in-gamut) and then convert the result to `Hsluv`, but that's a very different approach than anything I see currently implemented. 
- I also added macros for implementing `core::ops::Add` and related traits, since they're pretty repetitive. I also added tests to confirm that all of the arithmetic operations are implemented (though there is no correctness checking).
- I kept `l` as the field name, rather than `lightness`, since it is the same as the `Lchuv` `l`. 

Closes #112


palette/src/arithmetic.rs
```diff
@@ -0,0 +1,117 @@
+//! Macros to implement arithmetic traits on Color spaces.
+
+/// Implement `Add` and `AddAssign` traits for a color space.
+///
+/// Both scalars and color arithmetic are implemented.
+#[macro_export]
```
Ogeon, line 6: I always forget the details, but I'm pretty sure `#[macro_export]` is unnecessary for internal macros, _but_ the order the modules are declared in the `lib.rs` file matters. Macros and macro modules have to come first. I would suggest putting this module inside the `macros` module that's already in here.

Ogeon (commented):
Nice! Good call with the macros. I think they could be used for the other spaces as well. As for the random sampling, I'm pretty sure you can simply use the same method as "regular" HSL uses, which is to sample a [bi-cone](https://upload.wikimedia.org/wikipedia/commons/b/b3/HSL_color_solid_dblcone_chroma_gray.png), where the saturation becomes slimmer towards min and max lightness. There's already code for it in the HSL module. It's going to be its own space in a way, since it's warped from Luv's perspective, but I think that's fine. It's meant to be a more balanced alternative to the naive sampling (which anyone can do on their own), but not necessarily perfect in every way.

palette/src/arithmetic.rs
masonium: Ah, I bet the ordering was tripping me up on macros. I'll take your suggestion and just move them to `macros.rs`, though. I'll also do a bi-cone style implementation for `Hsluv`.

Ogeon:
Nice, thank you! It looks to me like this is ready to be merged. I just need to make the `no_std_test` work again. Looks like it requires `__aeabi_unwind_cpp_pr0` now, all of a sudden. Could be that some dependency changed.

Ogeon:
It should hopefully be fixed now in the master branch 🤞, so you need to merge it in here or rebase your branch. Whatever you are more comfortable with. 🙂

masonium:
I re-based the branch on top of master.

Ogeon:
Nice and the tests passed this time, so it seem to be all good! Thank you for all of this! I'll just take the liberty to add a "closing" line to the description, so that goes into the machinery.

PR #228, opened 2021-05-22
Update criterion-compare-action

### LINKED ISSUE #208: Update benchmarks on pull requests
At the moment the benchmarks which are done automatically at every PR are always failing. This should not be the case if the `criterion-compare-action` is updated (there is a [new repo](https://github.com/boa-dev/criterion-compare-action), because matchai stopped working on it).

Ogeon:
This should hopefully make it work for forks as well.

Closes #208

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,11 +5,14 @@ jobs:
     name: Run benchmark
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
-      - name: Fetch master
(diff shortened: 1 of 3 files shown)


Ogeon:
Ah, it looks like it doesn't handle throughput anymore.

Ogeon:
I decided to just disable the throughput for now. I'll trade it for functioning benchmark reports.

Ogeon:
Right, there's still throughput on master. But it should work. I'll take my chances.

bors r+

PR #236, opened 2021-07-11
Pre-release polishing

Ogeon:
This is a catch-all PR for anything I find while preparing for the release. Nothing that should affect already released functionality.

Changes:

* `Oklab` and `Oklch` no longer implements `ColorDifference`, since I can't find any sign of CIEDE2000 being correct there.
* Separated the first sentence for named gradients as its own paragraph. This makes nicer looking docs.
* Added an example of the difference between "naive" randomization and our implementation.


Ogeon:
I have also started writing the release blog post. It's a work in progress over at https://github.com/Ogeon/Ogeon.github.io/pull/6, in case anyone want to proof read or remind me of something important.

PR #238, opened 2021-07-15
Relax trait bounds for all color types

Ogeon:
I removed all trait bound on color type definitions and made them more relaxed in some places. This should also enable `const fn` in some places! I guess the unfortunate (or fortunate?) effect of this is that it highlighted some shortcomings in the current set if traits. Particularly that the `Component` + `FloatComponent` combo doesn't exactly reflect how the type parameters are used and comes with a lot of assumptions. In other words, there's more trait fixing than I thought.

## Breaking Change

Some forced trait bounds were changed in ways that will be breaking. Particularly on `WhitePoint`.

### DIFF
--- a/palette/examples/random.rs
+++ b/palette/examples/random.rs
@@ -5,7 +5,7 @@ fn main() {
 
 #[cfg(feature = "random")]
 fn main() {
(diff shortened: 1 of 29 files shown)


Ogeon:
I just realized I forgot `PreAlpha`, as usual. Will merge after this!

PR #239, opened 2021-07-15
Make color constructors const and remove or replace all with_wp

### LINKED ISSUE #134: Mark `new` functions as `const`
It allows:
`const COLOR: Srgba = Srgba::new(1.0, 1.0, 1.0, 1.0);`

Ogeon:
I went through the color constructors and made them `const fn` (or made a `const fn` version of them). The named colors are using them now!

I did also change the naming so that `new` is always producing a fully generic type and names like `new_srgb` can be used for specialized constructors. It's easier to discover via searching or auto complete if the name contains "new". I removed `with_wp` while doing this, except in `Hsv`, `Hsl` and `Hwb` where I figured it would be good to still have an sRGB specialization. They are more likely to be "entry points" than most other types.

## Closed Issues

* Closes #134

## Breaking Change

The renamed and removed constructors will cause some breakage. Type inference will also be affected and may require additional type hints.

### DIFF
--- a/palette/README.md
+++ b/palette/README.md
@@ -146,8 +146,8 @@ where
     color.mix(&new_color, amount)
 }
 
-let new_hsl = transform_color(&Hsl::new(0.00, 0.70, 0.20), 0.8);
-let new_hsv = transform_color(&Hsv::new(0.00, 0.82, 0.34), 0.8);
+let new_hsl = transform_color(&Hsl::new_srgb(0.00, 0.70, 0.20), 0.8);
+let new_hsv = transform_color(&Hsv::new_srgb(0.00, 0.82, 0.34), 0.8);
 ```
 
 This image shows the transition from the color to `new_color` in HSL and HSV:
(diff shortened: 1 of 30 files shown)


Ogeon:
I couldn't convert any function that moves out of any kind of value in any way, like `from_components((red, green, blue): (T, T, T)) -> Self`, That's not yet stabilized.

mockersf:
@Ogeon would it be possible to have a released version with this PR?

Ogeon:
I will do it sooner than later. I will start looking through it today to make sure I didn't forget anything important. My day job has just been quite draining recently so I haven't had much energy (and time) for it during the past few weeks.

mockersf:
Thank you for your reply.

Do take care!

Ogeon:
I have started writing up a quick blog post to summarize the changes, so it should be done sometime this week if nothing comes in the way. 🙂

> Do take care!

Thanks! I know it hasn't been fun for those who are waiting for the next release but this is how I keep it at a good level. I will try to separate breaking and non-breaking changes in the future so I can make smaller releases. 🤔

PR #246, opened 2021-07-18
Make most operator traits take their input by value and change `TransferFn` to `TransferFn<T>`

Ogeon:
This makes all of the traits have the same convention, modeled after the `core::ops` traits, and leaves any cloning to the caller. I considered adding an `Output`type too, but that can go into a separate pass.

## Breaking Change

Almost every trait that takes parameters have changed here, so that may be breaking for non-Copy types. The `clamp_self` method is also gone.


Ogeon:
I'm going to go through the traits again and do some more focused improvements. These were just some fixes and low hanging fruit they had in common.

bors r+

PR #251, opened 2021-08-01
Split Saturate into Saturate and Desaturate

Ogeon:
The `Saturate` trait is split to give us a separate `Desaturate` trait, and `*Assign` variants are added. Just like how `Shade` was changed. I did also notice a small bug in how `DarkenAssign` was blanket implemented.

## Breaking Change

`desaturate` and `desaturate_fixed` have been moved into a `Desaturate` trait. `Saturate` does also not require/imply `Sized`, and `Saturate::Scalar` does no longer require/imply `Float`.


Ogeon:
Oh I almost forgot...

bors r+

PR #269, opened 2022-01-15
Rework component traits to be more granular and remove num_traits

Ogeon:
This is quite a lot, but the main points are:

 * `num_traits` is replaced with more granular and permissive traits in `palette::num`.
 * Hues are no longer float only, leading to...
 * ...HSV, HSL, HWB are no longer float only, although with lots of limitations.
 * A `Stimulus` trait is introduced and `{From, Into}Component` is now `{From, Into}Stimulus`

In addition to that, I have moved some macros around and fixed things on the way.

I'm also adding a minimal VSCode config with some useful settings for rust-analyzer, so we all can have more uniform settings.

## Breaking Change

The new traits are not implemented for third party types and generic code will most likely break.


Ogeon:
This was a tough decision and a bit of a mess. I'm considering putting `num_traits` back as an optional dependency with default implementations of the traits, but that may have bad side effects with conflicting implementations that need to be investigated. The main rationale for all of this is to lay the foundation for allowing more third party types (not only floats) and eventually SIMD. The latter will require much more work, but this is a first step.

Ogeon:
I moved the benchmarks to run on Windows, hoping those servers will be a bit less noisy. I have noticed that the CI tasks run faster there, so hoping for the best in the long run. As for this PR, it seem to stay at about the same performance as before.

PR #270, opened 2022-02-14
Correcting documentation link.

dlfelps:
Fixes incorrect link in the documentation.


Ogeon:
Ah, good catch! Thank you!

bors r+

PR #273, opened 2022-03-06
Split and rework the Blend trait and bump MSRV to 1.55.0

### LINKED ISSUE #178: Lch color blending
Color blending in the Lch color space would be awesome to have.

Some resources and implementations I found on the way:

- https://web.archive.org/web/20150214062932/http://www.stuartdenman.com/improved-color-blending
- http://labs.adamluptak.com/javascript-color-blending/
- https://ninedegreesbelow.com/photography/gimp-lch-blend-modes.html
- http://davidjohnstone.net/pages/lch-lab-colour-gradient-picker

I didn't find much sample code, but most implementations seem to do simple linear interpolation with special wrap-around handling of the hue channel.

### LINKED ISSUE #243: Improve the blend traits and minimize their dependencies
## Description

The traits in the `blend` module could be split up and/or reduced in complexity. Some ideas so far:

* Replace `ComponentWise` with `Pixel`.
* Make `Blend` and `BlendFunction` require as little as possible.
* `Blend` shouldn't have `from_premultiplied` and `into_premultiplied` methods.
* Do we even need `BlendFunction` or would a closure be enough?

I count this as "old code" so any past API design will have to prove its usefulness or be changed. :axe: :broom: :paintbrush: 

## Motivation

It makes them easier to implement, use and reason about.

Ogeon:
The `Blend` trait has been split into `Compose`, `Blend` and `BlendWith`, to allow more color spaces to be composed. The trait implementations were also changed from a default implementation to blanket implementations. Having blanket implementations allows users to only require the trait to be implemented without having to specify the trait's requirements. For example, it's only required to specify `where C: Blend` instead of listing all the mathematical requirements.

The `PreAlpha` type has also been changed a lot to get rid of the alpha type parameter and to leave premultiplication to the color types. A lot of it was developed in tandem with changing how the blending was implemented to make those implementations make more sense.

## Closed Issues

* Closes #243.

## Breaking Change

The `Blend` trait, the `PreAlpha` type and a lot of associated API has been changed. Most users should not notice too much, except for needing a different trait or getting it from the `blend` module. The `ComponentWise` trait is also gone, since all colors can be cast to arrays for iteration

MSRV is changed to 1.55.0 to be able to use `[T; N]::map`.


Ogeon:
This is what I have been up to for the past few weeks. At least to and from. There is still room for improvement but it should at least be more useful and less confusing than before. And the examples helps with verifying the implementation. Note, when looking at them, that they use linear colors. Many applications will not do that, so they may differ a bit. I'm considering allowing any RGB encoding for that reason.

PR #278, opened 2022-04-02
Implement SIMD support and add `wide` integration

Ogeon:
This adds initial support for SIMD types in most places. An exception is the `Luv` related types, where the conversion logic need extra attention. Some of the conversions aren't necessarily optimal but the focus was on making it work at all.

Integration with the [wide](https://crates.io/crates/wide) crate has been added behind a feature flag, as a first example. More SIMD crates can be added in the future.

## Breaking Change

Some functions that used to return `bool` is now returning a mask type. This mask type is still `bool` for regular floats and ints, so this change will mostly affect generic code. `GetHue` was also changed to no longer return `Option<T>` for SIMD friendliness.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -13,7 +13,8 @@ jobs:
           toolchain: stable
           override: true
           profile: minimal
-      - uses: boa-dev/criterion-compare-action@master
+      - uses: boa-dev/criterion-compare-action@v3.2.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           branchName: ${{ github.base_ref }}
+          features: "wide"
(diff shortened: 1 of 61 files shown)


Ogeon:
It's a bummer that the RGB to HSL and RGB to HSV conversion is so much slower. I'll try with the old one behind type ID checks (i.e. Great Value Specialization) for now and see if it works better. I should see if I can add benchmarks for the SIMD versions before merging this.

Ogeon:
Looks like the performance gain varies from nothing to several times faster, depending on the work. Converting sRGB to linear RGB is even a bit slower on my machine (possibly due to the `powf` implementation), converting RGB to HSV or HSL is slightly faster if I use `f32x8` but almost equal with `f32x4`, and converting between XYZ and RGB scales pretty good with the amount of lanes.

I don't think I will go through and optimize everything now. Just making sure there's any improvement at all.

Ogeon:
The benchmark fails because the `wide` feature isn't on `master`. But the logs show similar results. And it's pretty cool that it keeps on being feasible to run these benchmarks here!

PR #279, opened 2022-04-03
Split the TransferFn trait and add lookup tables for sRGB

### LINKED ISSUE #126: sRGB conversion lookup tables
Using lookup tables for at least converting u8 sRGB to and from linear RGB would skip a bunch of calculations, at the cost of space for the table(s). u8 -> f32 is just a simple lookup, while f32 -> u8 is trickier, but should be possible. Maybe through interpolation or if f32 -> linear u8 -> sRGB u8 has a reasonable precision loss. It needs some research.

### LINKED ISSUE #245: Improve the (non-)linear transfer function trait
## Description

We should try to split the `TransferFn` trait into at least `FromLinear` and `IntoLinear`. That should cover cases where it's not so trivial, or even impossible, to reverse the process.

I have also considered baking in number type conversion to allow possibly better quantization quality and conversion speed, but I have to read up some more on that. That could perhaps be covered separately with a more precise `IntoComponent` option.

## Motivation

The current setup assumes both that the process can be reversed and that the output type should be the same as the input type. That may not always be the case.

Ogeon:
This changes how the transfer functions work a bit:

* `TransferFn` has been split into `FromLinear` and `IntoLinear`.
* The two traits take two type parameters; one for the linear encoding and one for the non-linear encoding.

The split is useful for cases when a conversion is irreversible, as mentioned in #245. The two type parameters makes it possible to convert between number types while encoding and decoding. This makes it possible to skip `.into_format()` in some cases (currently when converting between `u8` and either `f32` or `f64`). In addition to that, I made those special cases faster by using [`fast-srgb8`](https://crates.io/crates/fast-srgb8), and a similar lookup table for `f64`.

## Closed Issues

* Closes #126.
* Closes #245.

## Breaking Change

* `TransferFn` has been split into `FromLinear` and `IntoLinear`.
* `{Rgb, Luma}::{from, into}_linear` and `{Rgb, Luma}::{from, into}_encoding` are now allowed to change the output type. This changes the API and type inference. The solution for the type inference is usually to remove `into_format` from the call chain.
* `{Rgb, Luma}::{from, into}_encoding` will only convert from linear encoding, not "cross convert" between encodings.
* The number type parameter and `'static`lifetime were removed from `RgbStandard`, `LumaStandard` and `RgbSpace`.


Ogeon:
Oh, right, the throughput still breaks it. :tired_face:

Ogeon:
Some of those benchmark results don't make sense. I'll give them another spin and see if they are less noisy... But I may also reconsider the float -> byte conversion. The improvement wasn't so large anyway.

Ogeon:
It's interesting how it's consistently severely slower. I'll have to double check and see if I measured it correctly later.

Ogeon:
I _had_ measured it in some weird way. :thinking: I have simply replaced my hack solution with [`fast-srgb8`](https://crates.io/crates/fast-srgb8). It uses the same table as I had added for `f32` and has a very fast algorithm for the reverse case.

PR #290, opened 2022-07-16
Update phf to 0.11.0

### LINKED ISSUE #289: Windows Defender identifies `proc-macro-hack`'s transitive dependency artifacts as a threat
## How To Reproduce

1. Use a Windows 11 installation with up-to-date threat definitions for Windows Defender (as of 2022-07-15).
2. Clone this project using Git.
3. Run `cargo build`.
4. Observe that Windows Defender will (eventually) flag files in the user's `Temp` folder as severe threats, like the following screenshot:

    image

    This can be expedited by navigating to `%USERPROFILE%\AppData\Local` in Windows Explorer and initiating a `Quick Scan`:

    image

## Expected Outcome

Windows shouldn't be flagging this library's build artifacts as malicious...unless that's your intent! 😜

## Actual Outcome

Windows flags a threat as described above.

## Additional Details

* This should be fixable after https://github.com/rust-phf/rust-phf/issues/255 is resolved and published. I have a [PR](https://github.com/rust-phf/rust-phf/pull/256) open for it already.
* Cargo.toml entry: Reproducing on my machine with commit 88f40283898b3992e9dc5604afdc16edd6a6d88f with no modifications.
* Rust version(s): `rustc 1.60.0 (7737e0b5c 2022-04-04)`
* Target platform: Windows 11, version 10.0.22000.

Ogeon:
As mentioned in #289 a transitive dependency, via `phf`, was flagged by Windows Defender. The dependency could be removed in `phf`, and this change updates the version of `phf` in `palette` to one that doesn't depend on the removed dependency. `phf` is not part of the public API and can thus be patched as a non-breaking change.

I'm treating this as a security patch and making an exception from the general MSRV rule. Also clarifying in print that this exception may have to happen. It may still support older versions when the `named` feature is disabled, unless they are eagerly parsing the `phf` crate.

## Closed Issues

* Fixes #289, by changing to a newer `phf` version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        toolchain: [1.48.0, stable, beta, nightly]
+        toolchain: [1.60.0, stable, beta, nightly]
     runs-on: ${{ matrix.os }}
     env:
       RUSTFLAGS: -D warnings
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ on:
   pull_request:
     branches:
       - master
+      - "[0-9]+.[0-9]+"
   push:
     branches:
       - staging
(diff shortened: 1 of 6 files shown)


Ogeon:
I'm not going to bother updating the benchmark test for this.

PR #297, opened 2022-10-16
Expose doc comment for Rgb::from_str

brandondong:
I was looking through the crate's documentation for a way to parse an RGB hex code string and stumbled on [Rgb::from_str](https://docs.rs/palette/0.6.1/palette/rgb/struct.Rgb.html#impl-FromStr-for-Rgb%3CS%2C%20u8%3E). However, there were no examples/description provided and so I had no idea if the format I was working with would be accepted. Clicking the source tab revealed a helpful comment.

This change just exposes that as a doc comment.


Ogeon:
Oh, that's a really good point and a great example of when doc comments on trait implementations become useful! Thanks a lot!

bors r+

PR #301, opened 2023-01-09
Remove the Gradient type and its module

### LINKED ISSUE #152: B-spline interpolation
A lot of color schemes that are used in data visualization (https://github.com/d3/d3-scale-chromatic/) are based on b-spline interpolations. Palette already has linear interpolation (`Mix`), is adding b-spline interpolation something that you would consider?

I have done first experiments with the `bspline` crate (which is only passively maintained) and it works well for `LinSrgb`. Here is an example of the _Spectral_ diverging color scheme:

```rust
pub type Color = palette::LinSrgba;

pub fn interpolate_spectral(t: f32) -> Color {
    let colors: Vec<palette::LinSrgb> = vec![
        palette::LinSrgb::<u8>::new(0x9e, 0x01, 0x42).into_format(),
        palette::LinSrgb::<u8>::new(0xd5, 0x3e, 0x4f).into_format(),
        palette::LinSrgb::<u8>::new(0xf4, 0x6d, 0x43).into_format(),
        palette::LinSrgb::<u8>::new(0xfd, 0xae, 0x61).into_format(),
        palette::LinSrgb::<u8>::new(0xfe, 0xe0, 0x8b).into_format(),
        palette::LinSrgb::<u8>::new(0xff, 0xff, 0xbf).into_format(),
        palette::LinSrgb::<u8>::new(0xe6, 0xf5, 0x98).into_format(),
        palette::LinSrgb::<u8>::new(0xab, 0xdd, 0xa4).into_format(),
        palette::LinSrgb::<u8>::new(0x66, 0xc2, 0xa5).into_format(),
        palette::LinSrgb::<u8>::new(0x32, 0x88, 0xbd).into_format(),
        palette::LinSrgb::<u8>::new(0x5e, 0x4f, 0xa2).into_format(),
    ];

    // Similar to the implementation in d3, the first and the last element are tripled.
    // More information: https://github.com/d3/d3-shape/blob/master/README.md#curveBasis
    let knots = vec![
        0f32, 0f32, 0f32,
        0f32,
        1f32 / 8f32,
        2f32 / 8f32,
        3f32 / 8f32,
        4f32 / 8f32,
        5f32 / 8f32,
        6f32 / 8f32,
        7f32 / 8f32,
        1f32,
        1f32, 1f32, 1f32

    ];
    let spline = bspline::BSpline::new(3, colors, knots);
    let point = spline.point(t);
    point.into()
}
```

Thanks!

### LINKED ISSUE #156: Redesign and improve the Gradient type and friends

Ogeon:
As discussed in #218, the `Gradient` type has not gotten the attention it deserves and it's better to remove it in favor of other alternatives. This will also make the library more focused and easier to maintain. The predefined, named gradient constants are also removed.

## Closed Issues

* Closes #152, since the `Gradient` is now gone.
* Closes #156, since the `Gradient` is now gone.

## Breaking Change

Both `Gradient` and `gradient::named` are gone with no replacement, in favor of better alternatives in other crates.


Ogeon:
I'm thinking of making it possible to multiply hues as if they are just regular numbers but will probably do that separately.

medzernik:
Are there any alternatives in other libraries that could replace the lost functionality? I would appreciate a tip.

Ogeon:
I would recommend [`enterpolation`](https://crates.io/crates/enterpolation) if you want something similar to the old gradient module. Other interpolation crates should work as well, but this one is the closest to a spiritual successor.

PR #320, opened 2023-04-23
Implement struct-of-arrays (SoA) utilities

### LINKED ISSUE #305: Pixel to Struct of Arrays (and back again!)
## Description

As discussed in https://github.com/Ogeon/palette/discussions/303, it would be useful to have an interface for converting an array/`Vec` of colors to and from Struct of Arrays format. The first step should be:

- [x] Construction of `Rgb<S, Vec<T>>` from some `I: IntoIterator<Item = Rgb<S, T>>`, and the same for the other color types.
- [x] Make it possible to get a color at some index. For example `fn get_mut(&mut self, index: usize) -> Option<Rgb<S, &mut T>>`.
- [x] Make it possible to iterate over the colors. For example `fn iter_mut(&mut self) -> impl Iterator<Item = Rgb<S, &mut T>`.
- [x] Make it simple to read from and assign to colors with reference components, such as `Rgb<S, &mut T>`.
- [x] Add counterparts to  `Vec::clear` and `Vec::extend`.
- [x] Make it possible to push and pop individual colors.

In addition to that, we may want to add chunks and windows iterators.

The reference components will likely be necessary, due to the split into separate arrays, unless there's a smarter solution. They don't have to enable much more than reading and writing, though. It's also what would make it hard to implement the `Index` and `IndexMut` traits, if possible at all.

## Motivation

The discussions in https://github.com/Ogeon/palette/discussions/303 covers it pretty well. One of the main reasons for a built-in interface is that it gives a way to correctly preserve the color type information across conversions. It also lowers the bar for working with SoA data.

Ogeon:
This adds a set of iterators and utilities for converting between arrays of structs and structs of arrays. There's definitely more that can be done, but it's a start. This pull request includes:

* Creating  and extending`ColorType<Collection<T>>` from any `IntoIterator<Item=ColorType<T>>`.
* Iterating over `ColorType<Collection<T>>`.
* Some `Vec` functions are available for `ColorType<Vec<T>>`, with more that can be added.
* Slicing and indexing with `get` and `get_mut`.
* Working with `ColorType<&T>` and `ColorType<&mut T>`, using `copied`, `cloned` and `set`.

See the discussion in #303 for some background.

## Closed Issues

* Closes #305.

### DIFF
--- a/palette/examples/struct_of_arrays.rs
+++ b/palette/examples/struct_of_arrays.rs
@@ -1,0 +1,50 @@
+use palette::{cast, color_difference::EuclideanDistance, IntoColor, Oklab, Srgb};
+
+fn main() {
+    let image = image::open("example-data/input/fruits.png")
+        .expect("could not open 'example-data/input/fruits.png'")
+        .to_rgb8();
+
+    let image = cast::from_component_slice::<Srgb<u8>>(image.as_raw());
+
+    // Convert and collect the colors in a struct-of-arrays (SoA) format, where
+    // each component is a Vec of all the pixels' component values.
+    let oklab_image: Oklab<Vec<f32>> = image
+        .iter()
+        .map(|rgb| rgb.into_linear::<f32>().into_color())
+        .collect();
+
+    // Find the min, max and average of each component. We are doing it by
+    // iterating over each component Vec separately, starting with l...
+    let (min_l, max_l, average_l) = get_min_max_average(oklab_image.l.iter().copied());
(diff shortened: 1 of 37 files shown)


Ogeon:
There are still a few pieces left, but it's time to stop for today.

Ogeon:
I believe this fulfills the initial requirements and some more. :slightly_smiling_face: I struggled a bit with coming up with a good example, so that may be improved later.

PR #328, opened 2023-05-06
Add the HyAB color difference metric

### LINKED ISSUE #318: Implement HyAB color difference metric
_Copied from https://github.com/Ogeon/palette/issues/288#issuecomment-1519012587_
> The HyAB (hybrid model of `|delta L*| + Euclidean(a*, b*)`) color distance metric is presented in Equation 3 on page 4. It seems like it could be a good addition to `Lab` because it separates the effect of luma and chroma/hue in color difference calculation while remaining computationally simple. Unfortunately, the `Lch` Manhattan distance and hybrid models require calculating the expensive factors in the DE2000 calculation and perform slightly worse overall than HyAB, so they probably wouldn't be used as much and seem like less of a priority.
> http://markfairchild.org/PDFs/PAP40.pdf

Ogeon:
Adds the HyAB color distance metric for `Lab`, `Oklab` and `Luv`. It's primarily only meant for CIELAB, but I figured it may translate well to the other two as well, since they have similar semantics.

I have also added a summary of the available difference metrics in the module description. It's not exactly precise or scientific, but may still be helpful.

## Closed Issues

* Closes #318.

### DIFF
--- a/palette/src/color_difference.rs
+++ b/palette/src/color_difference.rs
@@ -1,4 +1,27 @@
 //! Algorithms for calculating the difference between colors.
+//!
+//! ## Selecting an algorithm
+//!
+//! Different distance/difference algorithms and formulae are good for different
+//! situations. Some are faster but less accurate and some may only be suitable
+//! for certain color spaces. This table may help navigating the options a bit
+//! by summarizing the difference between the traits in this module.
+//!
+//! **Disclaimer:** _This is not an actual benchmark! It's always best to test and
+//! evaluate the differences in an actual application, when possible._
+//!
+//! Property explanations:
+//! - **Complexity:** Low complexity options are generally faster than high
+//!   complexity options.
+//! - **Accuracy:** How the numerical difference compares to the perceived
+//!   difference. May differ with the color space.
+//!
+//! | Trait | Complexity | Accuracy | Notes |
+//! |-------|------------|----------|-------|
+//! | [`Ciede2000`] | High | High for small differences, lower for large differences | The de-facto standard, but requires complex calculations to compensate for increased errors in certain areas of the CIE L\*a\*b\* (CIELAB) space.
+//! | [`EuclideanDistance`] | Low | Medium to high for perceptually uniform spaces, otherwise low | Can be good enough for perceptually uniform spaces or as a "quick and dirty" check.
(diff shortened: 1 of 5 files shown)


Ogeon:
@okaneco, you may want to have a look at this, so I don't make too many things up in the documentation. I would appreciate an extra pair of eyes, if you want to/have time. :slightly_smiling_face:

palette/src/color_difference.rs
```diff
@@ -1,4 +1,27 @@
...
+//! - **Complexity:** Low complexity options are generally faster than high
+//!   complexity options.
+//! - **Accuracy:** How the numerical difference compares to the perceived
+//!   difference. May differ with the color space.
+//!
+//! | Trait | Complexity | Accuracy | Notes |
+//! |-------|------------|----------|-------|
+//! | [`Ciede2000`] | High | High for small differences, lower for large differences | The de-facto standard, but requires complex calculations to compensate for increased errors in certain areas of the CIE L\*a\*b\* (CIELAB) space.
+//! | [`EuclideanDistance`] | Low | Medium to high for perceptually uniform spaces, otherwise low | Can be good enough for perceptually uniform spaces or as a "quick and dirty" check.
+//! | [`HyAb`] | Low | High for large differences, compared to CIEDE2000, for CIE L\*a\*b\* (CIELAB) | Similar to Euclidean distance, but separates lightness and chroma more. Limited to Cartesian spaces with a lightness axis and a chroma plane.
```
okaneco, line 23: I think `HyAb` is presented as being a "good enough" option even for smaller differences according to the conclusion. It's not as accurate as `Ciede2000` for small differences but the performance benefits outweigh the loss of accuracy for general purpose uses.

> It is reasonable to conclude that the HyAB formula has the best correlation with visual observations for very large color differences. It has the best performance for the first category and third category of the first dataset. It
also holds up well for smaller color differences and thus is a good candidate formula for image processing and
computer vision applications.
> ...
> So, the HyAB and HyCH could be used for small to large color difference calculation. However, the HyAB has a slightly better performance and the advantage of simpler computation.

In summary, something like
> High accuracy for medium to large differences. Less accurate than `Ciede2000` for small differences, but still performs well and is much less computationally expensive.

Everything else looks good 🙂

Ogeon: Ah, yeah, I missed that part. I'll update it! I think I'll also highlight that the WCAG 2.1 contrast only checks lightness. Thanks for having a look and I hope you didn't mind the ping!

okaneco: No worries, glad to help.

Good idea on the highlight in the relative contrast notes. I would expand the explanation of computer graphics slightly,
> Meant for checking contrast in computer graphics (such as between text and background colors)

Ogeon: Yeah, that's good to clarify. :+1:

PR #338, opened 2023-07-05
Add traits for casting collections of colors to and from other data types

Ogeon:
These traits are alternatives to the free functions in `cast`. The idea is to have less "type information" in the naming and only focus on the high level transform. In essence what `From` and `Into` also does. For example, with these changes:

```rust
let raw = &[0xFF7F0080u32, 0xFF60BBCC];

// Before
use palette::{rgb::PackedArgb, cast};
let colors = cast::from_uint_slice::<PackedArgb>(raw);

// After
use palette::{rgb::PackedArgb, cast::UintsInto};
let colors: &[PackedArgb] = raw.uints_into();
``` 

Casting from owned to slice is limited to a fixed set of owning types, since it's hard to cover everything that resembles a slice. Arrays and slices don't implement `Deref<Target=[T]>`, for example, while `AsRef<[T]>` doesn't work for the `Self` type...

In addition to that, I also upgraded some actions in the workflows and added Miri checks to CI. It's good to keep an eye on those transmutes and pointer casts and even better to do it automatically. I have only been running it manually so far.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,18 +1,17 @@
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - '**.rs'
 name: Benchmark pull requests
 jobs:
   runBenchmark:
     name: Run benchmark
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2.3.4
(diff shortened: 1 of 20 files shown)


Ogeon:
I replaced the generic implementations with specific, but more predictable implementations. Also added doc examples for each trait.

PR #342, opened 2023-07-16
Implement CAM16

### LINKED ISSUE #199: CAM16 implementation
It seems that the state-of-the-art currently in colour spaces is the CAM16 colour space. @nschloe wrote [a paper](https://arxiv.org/pdf/1802.06067.pdf) concerning improvements to the colour space, which they then [implemented in Python](https://github.com/nschloe/colorio/blob/master/colorio/_cam16.py) as part of a larger colour space library. I know very little about colour spaces or mathematics, so I don’t know how much I could do apart from mechanically porting the Python implementation to Rust. Is this something that would be helpful?

Ogeon:
Implements support for CIE CAM 16 and surrounding parts.  This is a work in progress that is put up for early review, in case anyone has feedback or want to play Find The Typo. Based on https://observablehq.com/@jrus/cam16.

- [x] XYZ to CAM16
- [x] CAM16 to XYZ
- [x] XYZ to UCS
- [x] UCS to XYZ
- [x] DeltaE
- [x] Add missing traits
  - [x] Full
  - [x] Partial
  - [x] UcsJab
  - [x] UcsJmh
- [x] Add more documentation
  - [x] Full
  - [x] Partial
  - [x] UcsJab
  - [x] UcsJmh
- [x] Add missing tests
  - [x] Full
  - [x] Partial
  - [x] UcsJab
  - [x] UcsJmh
- [x] Make `Surround` represent percent (20% instead of 2.0) and rename the `Custom` variant to `Percent`.
- [x] Change `discounting` to an enum to make room for a custom discounting degree later.
- [x] Figure out what the typical scale for values are. The reference page assumes XYZ is 0-100, for example.
- [x] Make sure the default parameters are reasonable.

## Closed Issues

* Closes #199.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,8 @@ jobs:
         run: cargo clippy -v -p palette --no-default-features --features "std find-crate"
       - name: Default check
         run: cargo clippy -v -p palette
+      - name: Check all features
+        run: cargo clippy -v -p palette --all-features
       - name: Test all features
         run: cargo test -v -p palette --all-features
       - name: Test each feature
@@ -80,7 +82,7 @@ jobs:
         with:
           targets: thumbv6m-none-eabi
       - name: "Build with #[no_std]"
-        run: cargo build -v -p no_std_test --features nightly --target thumbv6m-none-eabi
+        run: cargo build -v -p no_std_test --all-features --target thumbv6m-none-eabi
   miri:
     name: Miri tests
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ Have a shiny new feature in store? Remember to explain it thoroughly and provide
 
 ## Testing
 
-Every pull request is automatically tested with continuous integration to deny warnings and any missing documentation. It's a good idea to run your local tests with `RUSTFLAGS="-D warnings" cargo test` and also to run `cargo check` and `cargo build` with the compiler flag prepended. This will make sure that there are no warnings or missing documentation, all for the benefit of the user.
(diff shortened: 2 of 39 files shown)


Ogeon:
The precision seem strangely low on my computer (3-4 digits are the same before it diverges). I can't find what causes it, but it's probably in `xyz_to_cam16`, since the output of `prepare_parameters` was identical to the reference.

Ogeon:
The accuracy difference may be some trigonometry function that's different. Converting back to RGB yields near perfect results, so the round trip seem to be working as it should.

Ogeon:
Getting UCS to work well is pretty challenging. The main issue is that it's based on brightness, while the "main" metric and default for `PartialCam16` is lightness. The current version makes it possible to set the attributes statically, so UCS can require the input to have brightness and chroma. The downside with this is that there are essentially 7 versions of `PartialCam16` (more if static and dynamic attributes are mixed), which adds a bunch of extra for a user to wrap their head around.

The good news is that this solution works relatively well with the derive macro for `FromColorUnclamped`. It just required a small extension (I also made errors a bit nicer). Part of the "trick" is to bypass `Cam16`, so `PartialCam16` becomes the "parent" type for UCS. This breaks up the triangle relationship between `Xyz`, `Cam16` and `PartialCam16`, and makes it so the conversions are as direct as possible.

Ogeon:
Splitting the colors into groups, and forcing XYZ <-> CAM16 conversion to be explicit, avoids the issue with the default parameters. Now it's up to the user to make sure they are set as expected.

Ogeon:
I think this latest change solved the last major challenge with this addition. It's nice that splitting the spaghetti of conversations into groups actually seems to work so well, since I really didn't like the idea of having default parameters. It never felt right, but this is better.

Also, in case anyone wonders what's taking so long, it's both because of the tricky problems like these (I also made some other changes earlier, such as the macros, to lay a better foundation) and because I haven't been feeling like programming as much in my free time lately. This comes and goes over time, so nothing to worry about. It's just how I manage my time and energy. 🙂

Ogeon:
For anyone who's watching this, feel free to start testing it. I would appreciate any input on the API and of course possible bugs.

Ogeon:
I ended up changing the approach for partial CAM16. The generic struct complicated trait implementations to the degree that it didn't really save any effort, compared to a cookie cutter macro solution. That and the awkward naming and newtype structs didn't sit right with me. This new approach is easier to treat as any other type, at the price of more repetition in some places. Still not too bad.

Ogeon:
The `cam16` module needs some documentation, but I think it will finally be good to go after that. This is enough for today, though.

Ogeon:
That's the largest squash I have ever done. I think it's ready for merging now.

PR #344, opened 2023-07-26
Fix Oklab from Oklch hue conversion

tomcur:
The Oklch -> Oklab conversion currently inverts the hue by swapping the a and b components. This PR corrects the calculation and adds an Oklch -> Oklab -> Oklch roundtrip test.

### DIFF
--- a/palette/src/oklab.rs
+++ b/palette/src/oklab.rs
@@ -357,13 +357,13 @@ where
     T: RealAngle + Zero + MinMax + Trigonometry + Mul<Output = T> + Clone,
 {
     fn from_color_unclamped(color: Oklch<T>) -> Self {
-        let (sin_hue, cos_hue) = color.hue.into_cartesian();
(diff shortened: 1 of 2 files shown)


Ogeon:
Hi, thank you for spotting this and for taking the time to fix it (with added tests too!). It may have been a naming confusion when replacing `sin_cos`. Either way, I'll let the tests run through, but it looks good to me. :+1:

Ogeon:
Alright, it's all green, so let's merge this.

bors r+

tomcur:
Thank you for the quick merge!

wash2:
Is there planned to be a release with this fix?

Ogeon:
I'll try to get it out within the next few days. I was initially going to add CAM16 (#342) in the same release, but it turned out to be a bit trickier than expected. I'll put it to the side for a bit and add a couple of missing casting traits, and then make the release.

PR #347, opened 2023-08-18
Implements Rgba::from_str()

rafaelbeckel:
This change implements the from_str() trait for the Rgba type, adding support for parsing a RGBA hex string with the format `#rrggbbaa`.

No related issue, no breaking changes.

### DIFF
--- a/palette/src/rgb/rgb.rs
+++ b/palette/src/rgb/rgb.rs
@@ -936,6 +936,8 @@ pub enum FromHexError {
     ParseIntError(ParseIntError),
     /// The hex value was not in a valid 3 or 6 character format.
     HexFormatError(&'static str),
+    /// The hex value was not in a valid 8 character format.
+    RgbaHexFormatError(&'static str),
 }
 
 impl From<ParseIntError> for FromHexError {
@@ -946,9 +948,13 @@ impl From<ParseIntError> for FromHexError {
 
 impl From<&'static str> for FromHexError {
     fn from(err: &'static str) -> FromHexError {
-        FromHexError::HexFormatError(err)
+        match err {
+            err if err.contains("rgba") => FromHexError::RgbaHexFormatError(err),
+            _ => FromHexError::HexFormatError(err),
+        }
     }
 }
+
 impl core::fmt::Display for FromHexError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match &*self {
@@ -958,6 +964,9 @@ impl core::fmt::Display for FromHexError {
                 "{}, please use format '#fff', 'fff', '#ffffff' or 'ffffff'.",
                 s
             ),
+            FromHexError::RgbaHexFormatError(s) => {
+                write!(f, "{}, please use format '#ffffffff' or 'ffffffff'.", s)
+            }
         }
     }
 }
@@ -967,6 +976,7 @@ impl std::error::Error for FromHexError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match &*self {
             FromHexError::HexFormatError(_s) => None,
+            FromHexError::RgbaHexFormatError(_s) => None,
             FromHexError::ParseIntError(e) => Some(e),
         }
     }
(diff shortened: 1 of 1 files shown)


palette/src/rgb/rgb.rs
```diff
@@ -946,9 +948,13 @@ impl From<ParseIntError> for FromHexError {
 
 impl From<&'static str> for FromHexError {
     fn from(err: &'static str) -> FromHexError {
-        FromHexError::HexFormatError(err)
+        match err {
+            err if err.contains("rgba") => FromHexError::RgbaHexFormatError(err),
```
Ogeon, line 952: It would be nice if we can avoid this and construct the error variant directly. I'm not so fond of relying on message content and would have suggested removing this `From` implementation if it wasn't a breaking change.

palette/src/rgb/rgb.rs
```diff
@@ -999,6 +1009,27 @@ impl<S> FromStr for Rgb<S, u8> {
...
+        match hex_code.len() {
+            8 => {
+                let red = u8::from_str_radix(&hex_code[..2], 16)?;
+                let green = u8::from_str_radix(&hex_code[2..4], 16)?;
+                let blue = u8::from_str_radix(&hex_code[4..6], 16)?;
+                let alpha = u8::from_str_radix(&hex_code[6..8], 16)?;
+                let col: Rgba<S, u8> = Rgba::new(red, green, blue, alpha);
+                Ok(col)
+            }
+            _ => Err("invalid rgba hex code format".into()),
```
Ogeon, line 1028: We know this should become an rgba hex error, so we could make one explicitly instead of using `into`.

Ogeon (commented):
Thanks for taking the time to add this! I had a comment (2 about the same thing) regarding the error, but the rest looks good.

I'll take a proper look on Sunday or Monday, when I have proper computer access again, but I'm sure it will be ready to be merged as soon as the error comment(s) can be addressed. Feel free to remind me if I seem to have forgotten. 😅

Ogeon:
Another thing to consider is to have an entirely different error type for rgba. I'm not sure how helpful it would be.

Ogeon:
The tests should be fixed with #349, so you will have to rebase (or merge) this PR to get those changes here as well.

Ogeon:
Anyway, I'm back from my adventures (some nice mountain hiking) and gave this another look. Apart from the error handling, I would suggest adding support for the 4 character shorthand format, too. It may not be as common, but I think it makes sense to at least be consistent with `Rgb<S, u8>` in that way. But that would be all!

palette/src/rgb/rgb.rs
rafaelbeckel: Thanks for the feedback! I agree, I'll adjust it and push again.

rafaelbeckel:
> Another thing to consider is to have an entirely different error type for rgba. I'm not sure how helpful it would be.

I tried that, but it turned out to be too verbose, and the definitions became too far away from the usage. I decided to stick with a variant instead. It makes more sense semantically, too.

palette/src/rgb/rgb.rs
```diff
@@ -944,11 +946,6 @@ impl From<ParseIntError> for FromHexError {
     }
 }
 
-impl From<&'static str> for FromHexError {
-    fn from(err: &'static str) -> FromHexError {
-        FromHexError::HexFormatError(err)
-    }
-}
```
Ogeon, line 951: Sorry if I wasn't clear before. We need to keep this for now, so it's backwards compatible, but it doesn't have to change to produce any RGBA errors. I can remove it later, for 0.8.0.

Ogeon (commented):
Thanks for checking the separate error idea. I agree that it sounds better to have only one type.

palette/src/rgb/rgb.rs
Ogeon: It's technically still breaking, depending on how you see it, since it changes the output for anything that contains "rgba" somewhere. In practice, I doubt anyone is relying on it at all.

I think the simplest option is to leave it unchanged and stop using it, or at least not use it for new errors. Then it still behaves exactly as before. I can make an issue to clean it up further.

Ogeon: You can add a documentation comment to the impl block that explains what it outputs if you want. 🙂

Ogeon: Great, thank you! 👍

Ogeon:
The code looks good and the tests are all green. 🙂 Thanks a lot!

bord r+

PR #351, opened 2023-08-28
Improve some documentation for beginners

Ogeon:
An attempt to add some documentation that may be useful for those who are new to Palette. It can always be improved more, but this is at least something. It adds a "where do I start?" section in the crate root, explains different ways of creating RGB and RGBA values (that generally promotes `Srgb` and `Srgba`), and adds some conversion docs in `convert`.

### DIFF
--- a/palette/src/alpha/alpha.rs
+++ b/palette/src/alpha/alpha.rs
@@ -39,7 +39,7 @@ use crate::{
 ///
 /// # Creating Transparent Values
 ///
-/// The color types in Palette has a transparent type alias, such as
+/// The color types in Palette have transparent type aliases, such as
 /// [`Srgba`](crate::Srgba) for [`Srgb`][crate::Srgb] or [`Hsla`](crate::Hsla)
 /// for [`Hsl`](crate::Hsl). These aliases implement `new` and other useful
 /// methods. Here's the same example as for [`Rgb`](crate::rgb::Rgb), but with
@@ -54,7 +54,7 @@ use crate::{
 /// // `new` is also `const`:
 /// const RGBA_U8: Srgba<u8> = Srgba::new(171, 193, 35, 128);
 ///
-/// // Conversion methods from the color type is usually available for transparent
+/// // Conversion methods from the color type are usually available for transparent
 /// // values too. For example `into_format` for changing the number format:
 /// let rgb_u8_from_f32 = Srgba::new(0.3f32, 0.8, 0.1, 0.5).into_format::<u8, u8>();
 ///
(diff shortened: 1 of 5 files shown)


palette/src/convert.rs
```diff
@@ -1,18 +1,90 @@
...
+//! methods. They are possible to work around by splitting the conversion into
+//! multiple steps.
+//!
+//! # In-place Conversion
+//!
+//! It's possible for some color spaces to be converted in-place, meaning the
+//! source color will use the memory space of the destination color. The
+//! requirement for this is that the source and destination color types have the
+//! same memory layout. That is, the same component types and the same number of
+//! components. This is verifies by the [`ArrayCast`](crate::cast::ArrayCast)
```
okaneco, line 47:
```suggestion
//! components. This is verified by the [`ArrayCast`](crate::cast::ArrayCast)
```

palette/src/convert.rs
```diff
@@ -1,18 +1,90 @@
...
+//! }
+//! ```
+//!
 //! # Deriving
 //!
-//! `FromColorUnclamped` can be derived in a mostly automatic way.
-//! The default minimum requirement is to implement `FromColorUnclamped<Xyz>`, but it can
-//! also be customized to make use of generics and have other manual implementations.
+//! `FromColorUnclamped` can be derived in a mostly automatic way. The other
+//! traits are blanked implemented based on it. The default minimum requirement
```
okaneco, line 75:
```suggestion
//! traits are blanket implemented based on it. The default minimum requirement
```

palette/src/rgb.rs
```diff
@@ -72,21 +72,64 @@ pub use self::rgb::{FromHexError, Iter, Rgb, Rgba};
 pub mod channels;
 mod rgb;
 
-/// Non-linear sRGB.
+/// Non-linear sRGB, the most common RGB input/output format.
+///
+/// If you are looking for "just RGB", this is probably it. This type alias
+/// helps locking the more generic [`Rgb`] type to the sRGB format.
```
okaneco, line 78:
```suggestion
/// helps by locking the more generic [`Rgb`] type to the sRGB format.
```

palette/src/alpha/alpha.rs
```diff
@@ -29,15 +29,111 @@ use crate::{
...
+/// ```
+/// use palette::Srgba;
+///
+/// let rgba_u8 = Srgba::new(171u8, 193, 35, 128);
+/// let rgab_f32 = Srgba::new(0.3f32, 0.8, 0.1, 0.5);
+///
+/// // `new` is also `const`:
+/// const RGBA_U8: Srgba<u8> = Srgba::new(171, 193, 35, 128);
+///
+/// // Conversion methods from the color type is usually available for transparent
```
okaneco, line 57:
```suggestion
/// // Conversion methods from the color type are usually available for transparent
```

Ogeon:
Thanks for looking through it @okaneco! :pray:

Ogeon:
I wasn't entirely happy with the exceptions/restrictions list, so I rewrote it a bit. Also fixed a couple of other mistakes around it.

PR #354, opened 2023-09-01
Actually interface with codspeed

Ogeon:
Forgot this part in #353.


Ogeon:
Ouch, it needs a newer rust version... :grimacing:

Ogeon:
Hah, putting them in their own crate worked. :laughing: It's also interesting how short they run compared to boa's criterion action. It's nice to get results fast, but I wonder about the stability. I guess time will tell.

PR #355, opened 2023-09-09
Add traits for delta E and the improvemed formulas from Huang et al

Ogeon:
Adds the `DeltaE`, `ImprovedDeltaE` and `ImprovedCiede200` traits. The first one represents the "default" or canonical delta E formula for a color space as it was originally specified. If there is no specification for delta E, it's not implemented. It's also not necessarily the best. The other two implement the enhancements described in *Power functions improving the performance of color-difference formulas* by Huang et al. They are provided as separate traits so the user can choose which one is the most appropriate for their use case.

This is the result of me trying to sort out where different formulas come from and prepare for adding a delta E for CAM16-UCS in #342. The recommended formula for CAM16 turns out to be the "improved" one.

### DIFF
--- a/benchmarks/benches/cie.rs
+++ b/benchmarks/benches/cie.rs
@@ -1,5 +1,8 @@
 use codspeed_criterion_compat::{black_box, criterion_group, criterion_main, Criterion};
-use palette::convert::FromColorUnclamped;
+use palette::{
+    color_difference::{Ciede2000, DeltaE},
+    convert::FromColorUnclamped,
+};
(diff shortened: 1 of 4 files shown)


Ogeon:
Ok, this is a pickle. The `memchr` dependency is only used in tests, so the dependency shouldn't matter for MSRV. Yet, I can't run the full tests without it. I'll see if I can re-shuffle the tests for now.

Ogeon:
Interesting, those benchmarks run about 10 to 20 times faster on my machine. Either way, it's the relative difference that's interesting.

PR #356, opened 2023-09-09
Move integration tests to a separate workspace crate to avoid MSRV is…

Ogeon:
Due to [an MSRV bump in `memchr`](https://github.com/BurntSushi/memchr/commit/e1f72c67de3ec9d5188923b95f3b8a570ef69536), I decided to move the integration tests to their own workspace crate.


Ogeon:
Well, the coverage report was helpful and notified me of the missing tests for `wide` :sweat_smile:

Ogeon:
Looks like the remaining reduction in coverage are false positives.

bors r+

PR #361, opened 2023-10-15
Remove some more boilerplate with more macros

Ogeon:
The only observable effect is that equality checks for `Hwb` works the same way as other radial types, instead of ignoring the hue for grayscale colors. The rest should be the same.


Ogeon:
Huh, MSRV 1.60.0 is really starting to be old, according to the dependencies.

PR #374, opened 2024-01-24
Fix typo about max range of u8

MultisampledNight:
- https://docs.rs/palette/0.7.3/palette/stimulus/trait.Stimulus.html lists

> Unsigned integer values (`u8`, `u16`, `u32`, etc.) have a range from `0` to their largest representable value. For example `0u8` to `155u8` [...]

However, the largest value of `u8` is not `155`, it's `255`. Chances are it's just been a typo. 

No code was changed, only the documentation.


Ogeon:
Yes, that's very much a typo. Thanks for spotting it and submitting a fix!

Ogeon:
Ignore those two 😅

PR #391, opened 2024-04-26
chore(bench): update CodSpeed/action to v2

adriencaccia:
Upgrading to the v2 of https://github.com/CodSpeedHQ/action will bring a better base run selection algorithm, better logging, and continued support.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Build the benchmark target(s)
         run: cargo codspeed build -p benchmarks --features all_features
       - name: Run the benchmarks
-        uses: CodSpeedHQ/action@v1
+        uses: CodSpeedHQ/action@v2
         with:
           run: cargo codspeed run
           working-directory: integration_tests # mimics how tests work, so the colors CSV can be found


Ogeon:
Hey, thanks for this! I hadn't noticed there was a v2. I'll let it run through everything to see that it's all good.

Ogeon:
I will take care of that later. Thanks for this PR and all the work you put into this!

Ogeon:
It looks like it's having trouble finding the benchmarks now. Is this a familiar issue? https://github.com/Ogeon/palette/actions/runs/8859407386/job/24329330697

adriencaccia:
> It looks like it's having trouble finding the benchmarks now. Is this a familiar issue? [Ogeon/palette/actions/runs/8859407386/job/24329330697](https://github.com/Ogeon/palette/actions/runs/8859407386/job/24329330697)

This is because in the latest version of `cargo-codspeed`, benchmarks are now run in their package root directory.
So here in the case of `palette`, they are not run in the `integration_tests` anymore.

adriencaccia:
@Ogeon if the path in https://github.com/Ogeon/palette/blob/ffcab42be71d9250581a7ef09edcf3ca354509d5/integration_tests/tests/convert/data_color_mine.rs#L329 is relative to the folder actually running the benchmarks (ie: `benchmarks` in our case), `cargo codspeed run` works again
Maybe the solution would be to change a bit the `load_data()` function to ensure that it works, no matter what the `current_dir()` is?

adriencaccia:
> I was using the working directory parameter so it could still find the CSV files, due to the relative paths, but has its purpose changed? Running the same commands locally seemed to work like before, so I'm curious to know how that differs here. What I tried was to build the benchmarks from the workspace root and then run them from `integration_tests`. It found everything as usual. 🤔

Have you udpated `cargo-codspeed` to the latest version, `2.6.0`?
It should not work with it. And since the CI takes the latest version, it is failing as well

Ogeon:
I thought I did, but I'll have to check again. That would explain things. 🤔 It's about time I sort these files out anyway. I'll see if I give it another try on Wednesday, and get back to you about how it went. It's almost midnight where I live and tomorrow is a bit too busy.

adriencaccia:
I found a way to make it work again in #398 😇

Ogeon:
Well, I can't complain when the whole thing already was a hack. 😁 Thanks a lot! I hope you don't mind that I wait with the merge until tomorrow, when I have the rest of my brain up and running again. It looks like everything should be all good, though!

PR #398, opened 2024-04-29
chore(bench): make benchmarks work again

adriencaccia:
Make benchmarks and CodSpeed work again.

Implemented after this discussion: https://github.com/Ogeon/palette/pull/391#issuecomment-2080461609

This is not a long term fix, but it has the merit of making everything work again.


Ogeon:
Thank you!

PR #402, opened 2024-07-29
Fix `clippy::doc_lazy_continuation` lints

waywardmonkeys:
Rust 1.80 brought a new clippy lint that detected when itemized list entries didn't indent the lines after the first. Indent those lines to silence that warning.


Ogeon:
Thanks for taking the time to correct them! I see that a couple of other issues have popped up as well... :grimacing:

Ogeon:
The documentation step passed, so I will bypass the rest and fix them separately. I need to think about that `syn` MSRV issue a bit.

PR #411, opened 2024-08-11
Add Adobe RGB

Kirk-Fox:
This adds the Adobe RGB (1998) color space and standard (which uses a gamma 2.2 transfer function).

* Completes `a98-rgb` task of #408 by implementing Adobe RGB (1998).

### DIFF
--- a/palette/src/encoding/adobe.rs
+++ b/palette/src/encoding/adobe.rs
@@ -1,13 +1,14 @@
 //! The Adobe RGB (1998) standard.
 
 use crate::{
-    encoding::gamma::GammaFn,
     luma::LumaStandard,
-    num::Real,
+    num::{Powf, Real},
     rgb::{Primaries, RgbSpace, RgbStandard},
     white_point::{Any, D65},
     Mat3, Yxy,
 };
+
+use super::{FromLinear, IntoLinear};
 
 /// The Adobe RGB (1998) (a.k.a. opRGB) color space and standard.
 ///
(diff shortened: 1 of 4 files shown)


palette/src/encoding/adobe.rs
```diff
@@ -0,0 +1,109 @@
...
+             2.0413690, -0.5649464, -0.3446944,
+            -0.9692660,  1.8760108,  0.0415560,
+             0.0134474, -0.1183897,  1.0154096,
+        ])
+    }
+}
+
+impl RgbStandard for AdobeRgb {
+    type Space = AdobeRgb;
+    type TransferFn = GammaFn;
```
Ogeon, line 77: The `GammaFn` isn't going to be correct here. I noticed last weekend that there's a bug in it that makes it do the opposite transform. I would recommend implementing the traits on `AdobeRgb` and use `T::from_f64(256.0) / T::from_f64(563.0)` and its inverse. A small unit test to make sure it's not also backwards would be nice too.

Ogeon (commented):
Ah, round two! Thank you, you know now what I'm usually looking for, so there's not much to discuss this time. :D There's only the issue with `GammaFn`. I had somehow manged to get it backwards, probably when I refactored the transfer function traits.

Kirk-Fox:
Wow, yeah, that GammaFn bug is hard to spot. I added the custom implementation for `AdobeRgb`.

Ogeon:
Great, thank you! I appreciate the help with these!

By the way, I started looking into P3 before I got distracted by the backwards gamma function. Feel free to continue there if you feel like checking off more from the list, but it's enough to implement DCI-P3 and Display P3 for now. They are straight forward, as opposed to the HDR and D60/ACES cans of worms.

> Wow, yeah, that GammaFn bug is hard to spot.

Hard to spot and easy to get wrong. I'll probably deprecate parts of it and replace it with something more explicit. Or replace it completely. There's little to gain with it, compared to just calling the `powf` function, until we can have something like `f32` in `const` generics.

PR #412, opened 2024-08-12
Add DCI-P3, DCI-P3+, and Display P3

### LINKED ISSUE #367: Display P3 (and DCI-P3) support
## Description

I'm looking to implement Display P3 support (and potentially also DCI-P3) - as far as I can see, this should mostly encompass copying the sRGB code and changing the primaries to those of DCI-P3 (as the EOTF is the same as in sRGB). I'm happy to do it, if someone agrees to review and merge it.

Kirk-Fox:
This pull request implements the DCI-P3, DCI-P3+, and Display P3 color spaces and standards.

* Closes #367, by implementing the DCI-P3 and Display P3.
* Completes `display-p3` task of #408, by implementing Display P3.

### DIFF
--- a/palette/src/encoding/p3.rs
+++ b/palette/src/encoding/p3.rs
@@ -1,0 +1,277 @@
+//! The P3 color space(s) and standards.
+
+use core::marker::PhantomData;
+
+use crate::{
+    encoding::{FromLinear, IntoLinear, Srgb},
+    luma::LumaStandard,
+    num::{Powf, Real},
+    rgb::{Primaries, RgbSpace, RgbStandard},
+    white_point::{Any, WhitePoint, D65},
+    Mat3, Xyz, Yxy,
+};
+
+/// The white point of DCI-P3 (Theatrical) is based on a projector with a xenon bulb
+/// with a color temperature of ~6300K
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct XenonBulb;
+
+impl<T: Real> WhitePoint<T> for XenonBulb {
+    fn get_xyz() -> Xyz<Any, T> {
+        Xyz::new(
+            T::from_f64(0.314 / 0.351),
+            T::from_f64(1.0),
+            T::from_f64(0.335 / 0.351),
+        )
+    }
+}
+
+/// The theatrical DCI-P3 standard.
+///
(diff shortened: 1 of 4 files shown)


palette/src/encoding/p3.rs
```diff
@@ -0,0 +1,277 @@
...
+
+impl<T: Real> WhitePoint<T> for XenonBulb {
+    fn get_xyz() -> Xyz<Any, T> {
+        Xyz::new(
+            T::from_f64(0.314 / 0.351),
+            T::from_f64(1.0),
+            T::from_f64(0.335 / 0.351),
+        )
+    }
+}
```
Ogeon, line 27: It may be better to bake this into the `DciP3` type, unless it's a generally established/standardized white point. Or give it a DCI or P3 related name. I'm generally careful with suggesting that something is, in this case, the canonical approximation of a xenon bulb.

palette/src/lib.rs
```diff
@@ -313,8 +313,9 @@ pub use oklab::{Oklab, Oklaba};
 pub use oklch::{Oklch, Oklcha};
 #[doc(inline)]
 pub use rgb::{
-    AdobeRgb, AdobeRgba, GammaSrgb, GammaSrgba, LinAdobeRgb, LinAdobeRgba, LinRec2020, LinSrgb,
-    LinSrgba, Rec2020, Rec709, Srgb, Srgba,
+    AdobeRgb, AdobeRgba, DciP3, DciP3Plus, DisplayP3, GammaSrgb, GammaSrgba, LinAdobeRgb,
+    LinAdobeRgba, LinDciP3, LinDciP3Plus, LinDisplayP3, LinRec2020, LinSrgb, LinSrgba, Rec2020,
+    Rec709, Srgb, Srgba,
```
Ogeon, line 318: This is getting crowded. I can scale it back later if it messes with the other PR, but there's no need to export everything to the top module. I should have thought of it earlier.

Ogeon (commented):
Thank you! I wanted to respond earlier, but I didn't find a good moment to sit down and have a proper look during the week.

The only real uncertainty from my part is the white point. I try to name constants and algorithms in a way that makes the origin clear. Sure, it's name spaced in this case, but it can easily escape out of context.

Ogeon:
Alright, cool, thanks a lot! Once again, I appreciate all the help

Issue #352, opened 2023-08-29
Implement LMS color space and CVD simulation


Flourish38:
I was considering making my own crate for this, but figured it would probably make sense in the main palette crate. Just wanted to ask before I started work on it.

I have a project that needs color vision deficiency (CVD) simulation, and I think it would be a good fit for Palette. Along the way, this would involve implementing [the LMS color space](https://en.wikipedia.org/wiki/LMS_color_space), which I think would also be a good fit for Palette, since CVD simulation is not its only practical use (could even relate to #64? Spectrum to LMS makes arguably more sense). 

The method of simulating CVD that I am planning on implementing is Brettel & Mollon 1997, as per the [Review of Open Source Color Blindness Simulations](https://daltonlens.org/opensource-cvd-simulation/#So-which-one-should-we-use?) by DaltonLens, which I highly recommend reading as it seems to be the only comprehensive overview of the topic. I would like input if someone thinks that another CVD simulation method would be a better fit, or if implementing multiple methods would be better, which I would also be happy to do.

Thank you for the input!

Flourish38:
Yeah, I could see why you might want it in a separate crate. I still think it would be good to have it as part of palette, for a similar reason that the color difference metrics are in palette: They are a very common use case for certain color spaces. I don't have a super strong opinion one way or the other though.

Ogeon:
Let's start with the LMS and give the rest some thought. I need to understand how much work it would be to maintain it and keep it up to date with the development in the area. I don't have the personal bandwidth to keep up with the research on my own, even though it's interesting, so I have to prioritize a bit. The best scenario is if we can find some parts that are generic, established/standardized or very commonly used.

For LMS, I think you can look at XYZ. It's a plain tri-stimulus space without a lot of extra (as opposed to RGB). But you can also take a look at how the RGB standards work, it may help. I'm also trying to write some sort of guide or check list for how to add a new color type, but don't wait for that before starting. You probably know this better than I do, but my impression is that it would make sense to have a representation of the conversion matrix as the meta variable (XYZ has the reference white point for convenience). Or are they interchangeable once converted?

Ogeon:
To respond to https://github.com/Ogeon/palette/issues/376#issuecomment-1949936566

> As far as I understand, the problem with converting from LMS to CIE 1931 XYZ is that the cone response functions (or derived colour matching functions) differ. So, you would actually need to know the precise spectral make-up of the colour in question.

[...]

PR #414, opened 2024-08-13
Implement simulators for dichromacy and anomalous trichromacy

### LINKED ISSUE #401: CVD simulation
Split from #352 (see its description for more). Figure out how much of color vision deficiency simulation is useful and practical to implement as a start.

Kirk-Fox:
This pull request implements the Brettel 1997 and Vienot 1999 approaches to simulating color vision deficiency as described [here](https://daltonlens.org/understanding-cvd-simulation/). For anomalous trichromacy, I still am trying to understand the Machado 2009 method, but was unable to implement it thus far. It is still possible to include it at a later date and institute it as the new default for protanomaly and deuteranomaly, due to its handling of anomalous trichromacy being much more principled than using simple linear interpolation.

I am not sure of the best way to implement tests for these new features and any help in that respect would be appreciated.

* Closes #401, by implementing simulators for dichromacy and anomalous trichromacy.


Ogeon:
Thank you for looking into this and I appreciate the effort, but I'm still not sure it belongs in this library. It's one of those adjacent fields that I hesitate to include if they have a risk of growing too much. I'm not sure I can maintain it on my own, so it would have to be fine for it to stagnate without a "champion". It wouldn't be fair of me to merge it if I can't keep it going. That's why I phrased the issue as more of a research task, though I admit it's sparsely phrased.

I would be open to something relatively small, that's "write once, use forever". Say, something like the `chromatic_adaptation`, `color_difference` or `color_theory` modules, that very rarely need to grow and barely require maintenance. It could be a set of functions that generate `convert::Matrix3` matrices, for example. Otherwise, my standpoint from #352 hasn't changed much.

What's your point of view, so far?

Ogeon:
Yeah, it's probably something that needs to be experimented with a bit. There seem to be multiple parts to it that are a bit loosely defined in Palette (especially different CIE XYZ versions), so it may be better to start identifying them first and build up to better support. `Lms` and `Matrix3` are a start, at least.

I can't promise it will be integrated, but don't let me stop you from continuing for your own project. A clear use case always helps! If you find any problems along the way, we can tackle them separately and hopefully make things easier. That's more of a bottom-up approach that may be easier than trying to get it in as one unit.

Does that sound good?

Kirk-Fox:
Yeah, although there are other things higher on my list than more into this, so it might be a bit before I continue on it.

Ogeon:
You should of course do what you feel like. I have no expectations. I'll close this one for now.

PR #416, opened 2024-08-19
Add `FromLinear` and `IntoLinear` lookup table creation to build script

Kirk-Fox:
This PR adds the creation of float &harr; uint conversion lookup tables for `FromLinear` and `IntoLinear` to the crate's build script. As a result, this crate no longer has a dependency on [`fast-srgb8`](https://crates.io/crates/fast-srgb8) (I have confirmed that the lookup table used by [`fast-srgb8`](https://crates.io/crates/fast-srgb8) is identical to the one generated by the build script).

Along with this, I have added the features: `float_lut` (for `u8` to float conversions), `float_lut16` (for `u8`to float and `u16` to float conversions), `fast_uint_lut` (for fast `f32` to `u8` conversions), and `fast_uint_lut16` (for fast `f32` to `u8` and `f32` to `u16` conversions). Of these, I have added `float_lut` and `fast_uint_lut16` as the default features. `float_lut` must be default to not cause a breaking change for `Srgb` and `RecOetf`, whose lookup tables were replaced by the build script. I included `fast_uint_lut16` as a default feature since its largest generated table contains only 1152 `u64`s, although I would understand wanting to replace it with `fast_uint_lut` as the default since that only generates tables of less than 200 `u32`s.

Building the crate seems to take considerably longer now due to the new code. I added some statements to the main.rs file in the build script that might prevent the crate being built unnecessarily often (although I doubt it since every time I run a test it seems to rerun the build script even though I haven't changed anything). If there are improvements to be made, please let me know and I'll implement them.

Also, does removing a dependency qualify as a breaking change? If so, I can add back [`fast-srgb8`](https://crates.io/crates/fast-srgb8) so that it can be removed at the next major update.


Kirk-Fox:
I'm not sure I understand what's causing the tests to fail at this point.

Ogeon:
Interesting. The longer build times are worrying, but there may be optimization opportunities. I'm still having that cold I mentioned, so I'll have a proper look when I'm feeling better. A few quick things for now:

1. I suppose it doesn't make sense to depend on `fast-srgb8` anymore and that's not breaking, but its replacement has to be available unconditionally. It wasn't optional before.
2. The interpolation code doesn't need to be completely generated if it's largely the same every time.
3. The tests fail because `IntoLinear` is included unconditionally, but unused when some features are disabled.
4. For the cargo features in general, I would probably slice it the other way and have one (or two) per gamma function. That reduces the code size impact for each feature.

Kirk-Fox:
Alright, thanks for the direction. I'll fix the issue with `IntoLinear` and `FromLinear` being included unconditionally when I get the chance. How should I handle the interpolation code instead? I think it is useful to have this general purpose algorithm somewhere in the codebase, so how might I reconcile that with not fully running the interpolation code each time?

Kirk-Fox:
I changed around the features as you mentioned and I think I was also able to fix the issue with the build script running more often than it needs to. It should now, hopefully, only rerun if any of the files in the build folder are changed or if any features are changed. I also made it so that the code for `Srgb` isn't conditional so that it doesn't constitute a breaking change by removing `fast-srgb8`.

Currently, the conditional compilation for the 16-bit lookup table methods/structs are controlled by repeated uses of `#[cfg(feature = "prophoto_lut")]`. Ideally, I would have written it as `#[cfg(any(feature = "prophoto_lut"))]`, so that it's clear this should be compiled if any 16-bit lookup table feature is enabled (if more are added later), but this leads to a warning for redundancy. Is there a more idiomatic way of doing that?

Additionally, even though the build script runs less often now, it is still quite slow and seems to drastically increase the time the PR tests take, so, as mentioned before, do you have a suggestion on how I can reconcile including the float to uint lookup table generation code somewhere in the codebase while not fully rerunning it each time?

Kirk-Fox:
I may be starting to realize a way to optimize the build script, but it's going to take some further research. I'll hopefully be pushing a new commit within the week that will at least somewhat improve the build time.

palette/build/lut.rs
```diff
@@ -0,0 +1,592 @@
...
+            // Handle integer floats printing without decimal
+            let float_string = if linear <= 0.0 {
+                "\t0.0,\n".to_owned()
+            } else if linear >= 1.0 {
+                "\t1.0,\n".to_owned()
+            } else {
+                format!("\t{linear},\n")
+            };
+            table.extend_from_slice(float_string.as_bytes());
+        }
```
okaneco, line 260: I'm not sure of the overall impact, but it seems like a single string could be reused here and other places instead of some of the `format!` calls inside loops which return a new `String` every time. You can use `std::fmt::Write` to write formatted data to a buffer, but it has to be imported in a slightly funny way to avoid messing with the `std::io::Write` `write!`/`writeln!` macros. Also have to remember to clear it before reusing it.

https://doc.rust-lang.org/stable/std/macro.write.html#examples

---

Looking at the removed tables, it seems only the first and last values are `0.0` and`1.0`, respectively. If so, you could rewrite this loop to make fewer allocations and avoid all the clamping checks. Untested

```rust
use std::fmt::Write as _;
use std::io::Write as _;

let mut float_string = String::new();

// Handle integer floats printing without decimal
table.extend_from_slice(b"\t0.0,\n");

for i in 1..table size - 1 {
    let encoded = (i as f64) / ((table_size - 1) as f64);
    let linear = into_linear(encoded);
    
    write!(&mut float_string, "\t{linear}\n").unwrap();
    table.extend_from_slice(float_string.as_bytes());
    
    float_string.clear();
}

// Handle integer floats printing without decimal
table.extend_from_slice(b"\t1.0,\n");
```



Some possible

Kirk-Fox:
I went ahead and optimized the table building to use integration instead of summation since these are all exponential or linear functions. I also removed the tests for errors and just put in the values for the number of mantissa bits used in the index since they do not seem to change (3 for `u8`s and 7 for `u16`s).

Kirk-Fox:
Would it also be a good idea to generate the `u8` to `f32` tables separately? I'm not sure how much of a performance cost it is to convert from `f64` to `f32`

palette/build/lut.rs
```diff
@@ -0,0 +1,651 @@
...
+                    \n\t\tlet scale = entry & 0xffff;\
+                    \n\n\t\tlet t = (input_bits >> {man_shift}) & 0xff;\
+                    \n\t\tlet res = (bias + scale * t) >> 16;\
+                    \n\t\t#[cfg(test)]\
+                    \n\t\t{{\
+                        \n\t\t\tdebug_assert!(res < 256, \"{{}}\", res);\
+                    \n\t\t}}\
+                    \n\t\tres as u8\
+                \n\t}}\
+            \n}}"
```
Ogeon, line 452: This is the part I think we don't need to generate every time. It looks to me like it could be a library function that takes a table reference and input value.

palette/build/lut.rs
```diff
@@ -0,0 +1,651 @@
...
+            "\
+                    \t\tlet res = (bias + scale * t) >> 32;\
+                    \n\t\t#[cfg(test)]\
+                    \n\t\t{{\
+                        \n\t\t\tdebug_assert!(res < 65536, \"{{}}\", res);\
+                    \n\t\t}}\
+                    \n\t\tres as u16\
+                \n\t}}\
+            \n}}"
+        )
```
Ogeon, line 648: Could this also be unified as one function that we don't generate each time?

Ogeon (commented):
Alright, I'm feeling much better now and the workweek is over. Thanks for your patience in the meantime. I have also had some time to think about this and the slower build times from the code generation makes me think we should move away from `build.rs` for this. Considering none of this really depends on anything other than cargo features, I think it would be possible to generate the code once and making sure the larger tables are opt-in.

I'll try some things with a separate codegen crate...

It's also going to take me a moment to digest what everything does here. Some of the generated code becomes a bit hard to follow. Bear with me.

> Would it also be a good idea to generate the u8 to f32 tables separately? I'm not sure how much of a performance cost it is to convert from f64 to f32

You can try and see if it affects the benchmarks. Something seems to have made it a bit slower.

Kirk-Fox:
I'll start working on consolidating those into singular library functions and moving things to the codegen crate. I'll also try to document the code a bit more so it's clearer what's going on

Ogeon:
Thank you and sorry for the extra work. I do think this direction will be better. Now when the cost of generating the code is paid for in advance, we could change the cargo feature setup to something simpler (sorry again). I would suggest something like this:

* `u8` to/from float - always included.
* `u16` to/from float - behind an opt-in `gamma_lut_u16` feature.

Simple as that. This avoids inflating the test set too much and relies on the compiler's ability to disregard constants it doesn't use. The feature is more there so it's possible to keep the binary size slim. A `u16` to `f32` table is 256 kilobytes, while a `u8` to `f32` table is "only" one kilobyte, so it matters more if a `u16` table is included by accident. I hope that makes sense...

As for documentation, anything that explains the reasoning (especially where it diverges from the original) will help future us or other people who haven't been part of the process. Like with other things, I don't want this to turn into a black box later. I would also like to have references to the original implementation(s). That's key for tracing them back to the original reasoning and having something to compare to if there refactoring is required or if there are any issues. I know it's perhaps not the most fun thing to do, but it's even less fun to not have them later. So thanks in advance for taking some time to explain it. :pray: 

A side note with the new codegen crate; let me know if you run into any rough edges. One thing I have noticed with `quote` is that it's usually best to keep each invocation relatively small. Rust-analyzer and/or clippy seemed to struggle a bit with very large blocks. It's still generally better than strings IMO.

palette/src/encoding/srgb.rs
```diff
@@ -25,10 +21,10 @@ mod lookup_tables;
 /// There are some specialized cases where it has been optimized:
 ///
 /// * When converting from `u8` to `f32` or `f64`, while converting to linear
-///   space. This uses lookup tables with precomputed values. `f32` will use the
-///   table provided by [fast_srgb8::srgb8_to_f32].
+///   space. This uses lookup tables with precomputed values provided that the
+///   `float_lut` feature (enabled by default) is being used.
 /// * When converting from `f32` or `f64` to `u8`, while converting from linear
-///   space. This uses [fast_srgb8::f32_to_srgb8].
+///   space if the `fast_uint_lut` feature (enabled by default) is being used.
```
Ogeon, line 27: This mentions earlier feature names.

Ogeon (commented):
Nice, thank you! This is already easier to follow. The split with generated tables and handwritten functions seems to work well too. I will go into more details later.

palette/src/encoding/lut.rs
```diff
@@ -0,0 +1,73 @@
+mod codegen;
+
+const MAX_FLOAT_BITS: u32 = 0x3f7fffff; // 1.0 - f32::EPSILON
+
+// SAFETY: Only use this macro if `input` is clamped between `min_float` and `max_float`.
+macro_rules! linear_float_to_encoded_uint {
```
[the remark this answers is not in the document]
Kirk-Fox: I'm not sure how to define a macro as unsafe. Just adding `unsafe` before `macro_rules!` gets removed when I save the file (likely due to rustfmt)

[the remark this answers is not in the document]
Ogeon: Ah, no, I mean make the name say `unsafe_...`. It makes it look more dangerous.

codegen/src/lut.rs
```diff
@@ -0,0 +1,470 @@
...
+                        #table_ident[encoded as usize]
+                    }
+                }
+
+                impl IntoLinear<f32, u8> for #fn_type {
+                    #[inline]
+                    fn into_linear(encoded: u8) -> f32 {
+                        #table_ident[encoded as usize] as f32
+                    }
+                }
```
Ogeon, line 165: I think these can be "hand written", to keep them closer to the type's definition. They are trivial enough.

codegen/src/lut.rs
```diff
@@ -0,0 +1,470 @@
...
+                    #[inline]
+                    fn into_linear(encoded: u8) -> f64 {
+                        #table_ident[encoded as usize]
+                    }
+                }
+
+                impl IntoLinear<f32, u8> for #fn_type {
+                    #[inline]
+                    fn into_linear(encoded: u8) -> f32 {
+                        #table_ident[encoded as usize] as f32
```
Ogeon, line 163: Let's try adding an `f32` table to be used here, and see if that has any effect on the benchmark that got a slow-down.

codegen/src/lut.rs
```diff
@@ -0,0 +1,470 @@
...
+                        #table_ident[encoded as usize]
+                    }
+                }
+
+                #[cfg(feature = "gamma_lut_u16")]
+                impl IntoLinear<f32, u16> for #fn_type {
+                    #[inline]
+                    fn into_linear(encoded: u16) -> f32 {
+                        #table_ident[encoded as usize] as f32
+                    }
```
Ogeon, line 204: The same thing here as for `u8`. I wonder if it's also worth having a `u16` to `f32` table here. Let's see if it helps the `u8` case.

codegen/src/lut.rs
```diff
@@ -0,0 +1,470 @@
...
+            + 0.5 * (1.0 - alpha) * t * t
+    };
+
+    (
+        antiderive_y(end_x, end_t) - antiderive_y(start_x, start_t),
+        antiderive_ty(end_x, end_t) - antiderive_ty(start_x, start_t),
+    )
+}
+
+fn build_f32_to_u8_lut(entries: &[LutEntryU8]) -> TokenStream {
```
Ogeon, line 258: I would appreciate a comment that links to https://gist.github.com/2203834 for credit and explains any differences from it. Particularly the integration. It should be possible for someone who hasn't been part of the discussion to understand and bug fix it, with aid from the linked reference.

codegen/src/lut.rs
```diff
@@ -0,0 +1,470 @@
...
+                        <#fn_type>::from_linear(linear as f32)
+                    }
+                }
+
+                impl FromLinear<f32, u8> for #fn_type {
+                    #[inline]
+                    fn from_linear(linear: f32) -> u8 {
+                        lut::linear_f32_to_encoded_u8(linear, #min_float_bits, &#table_ident)
+                    }
+                }
```
Ogeon, line 355: Can we hand write these as well? The same for `u16` below.

palette/src/encoding/lut.rs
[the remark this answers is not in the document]
Ogeon: Alternatively, we remove the unsafe blocks from inside it. That may be better, so we are forced to mark the call site. :thinking:

Ogeon (commented):
I hope all is well and sorry for the absence. I think I lost more momentum than expected when I got sick, but it's better now after some time away from programming related things during my free time.

I have added a few comments where I would like to have some documentation. I think this will be good to go once them and the other comments have been resolved. How does that sound?

Kirk-Fox:
Sorry that I've been less active. The school semester started and is kicking my butt. Thank you for the extra comments and I'll try to address them soon

Ogeon:
Hi, how's it going? I just wanted to check in and also let you know that there are some fixes in `master` that you will need to (preferrably) rebase in if you get back to this. Clippy will complain about unrelated lifetime annotations otherwise.

How does it look regarding getting a moment for this PR? Would it help if we scope it down a bit? The most important change from my perspective is to have the rationale for the algorithm changes in your words. The rest are things I can take care of.

Kirk-Fox:
Yeah, scoping it down would help. I also am trying to type up a short document describing how I arrived at the formulae that I did, since it's a bit too in-depth to be written in comments in the code. I'll still include a basic explanation of what's going on in the documentation, though. Does that sound alright?

Ogeon:
That sounds great, thank you! I don't want you to feel like you are stuck with this, so the basic explanation could be good enough on its own if you want to keep it short. The point is to answer why this change was made (performance?) and what makes it equivalent to the original (referring to or showing an equation, algorithm, etc.). That usually works as reference for refactoring and bug fixing. Details for extra clarity are still appreciated, but don't sweat it.

When you feel like this is done, I would be grateful if you could also squash the commits into a single commit, like before. That should be it unless you want to address any of the other comments.

Ogeon:
Hello again! I hope things are going well. I just want to check in and see if there's a chance that you will have time to at least squash and rebase this change anytime soon. Skip the documentation if it's too much. That can always be improved later if you happen to have time. 

I would love to have it in the next release, sometime in the beginning of the next year. I may look into options to merge this anyway, if you didn't have time for it before, while of course preserving its connection to you for proper credits. Only to move it forwards. No hard feelings involved. That's not happening during the coming few weeks, though.

Kirk-Fox:
Okay, I did a lot to hopefully better organize the code and make it more readable/understandable. I still hope to make a full write-up at some point in the future, but it's pretty involved. I didn't quite understand what you meant regarding hand-writing certain functions, so I didn't change anything on that front.

I also added the `u8` to `f32` tables as you suggested. I think something might be wrong with the benchmarking tool, so I hope to see how that's affected things after that's back up and running. If you decide we should also include `u16` to `f32` tables, let me know and I will add them.

Let me know if you have any other comments. Once it looks good I can squash the commits into one again.

Ogeon:
Thank you, this is great! I should upgrade the benchmark action first, like you noticed... I'll let you know when it should work again.

>  I still hope to make a full write-up at some point in the future, but it's pretty involved.

I'm absolutely satisfied with the current comments, but feel free to add to them later if you want to and have the time. This is voluntary work, after all, and things like your studies are more important.

> I didn't quite understand what you meant regarding hand-writing certain functions, so I didn't change anything on that front.

It's fine, I just think they would be better to have in the "regular" source files, since they don't contain any generated values (other than the constant names), and only have the table constants in the generated code. Just to keep the types and their implementations together when possible. It's the kind of nit picks I can take care of later, so feel free to skip it.

> If you decide we should also include u16 to f32 tables, let me know and I will add them.

We can hold off on them for now, since it's not a regression if they are a bit slower. I would like to check how they affect the package size before deciding. I realized that large tables may eventually become an issue with the max package size.

Ogeon:
The benchmark should work now, after rebasing to bring in the latest test config. :slightly_smiling_face:

Ogeon:
Thank you and sorry for not noticing that you had already squashed it before. Github isn't always super clear in that regard.

I had a look at the benchmark and it seems like the regression for `f64` to `u8` is pretty consistent. I'm not sure what's causing it, but I'll accept it as a future improvement opportunity. There's also a regression for `f32` that's just below the threshold for appearing in the table: https://codspeed.io/Ogeon/palette/branches/Kirk-Fox%3Alookup. It could be that some slight difference prevents an optimization from kicking in. It's sometimes hard to know with tiny benchmarks like these.

But let's wrap this up. Thank you so much for your time and help! :star:

PR #418, opened 2024-10-13
Update `image` dev-dependency

waywardmonkeys:
Update the `image` dev dependency to the current version. This no visible end-user impact.


Ogeon:
Hi, thanks for tanking the time to bump it, but we'll have to hold off until it's time to increase the MSRV

Ogeon:
Hey, it has been a while, but if you are still around and interested in moving this forward, there's now a branch called `0.8` where we could merge your PR. You just need to rebase it onto that branch and set it as the base of this PR ("Edit" in the top right of the PR, then change the base under the title). It has a newer MSRV so the tests should pass once it's in the history of your branch.

PR #422, opened 2024-11-16
Add named color iterators and remove phf macro dependency

### LINKED ISSUE #421: Get a named color from a color value
## Description

Hello! Love this library, makes it very convenient to work with colors.

The ability to get a RGB value from a named color is amazing, but I've found myself wanting the reverse: I have a color (in my case, `Srgba<f32>`, but I assume the same could apply for other formats and types) and would like to get the named equivalent if it exists. If `COLORS` was exported, it would be trivial to go through the values to find that, but unfortunately it is not exposed.

## Motivation

My precise use case is transforming colors into the corresponding CSS, including a named color if it exists.

Ogeon:
This brushes up the `named` module a bit:

 * Add iterators for entries, names and color values.
 * Switch to `phf_codegen` instead of the proc macro.
 * Remove the need for the `"name_from_str"` feature.

## Closed Issues

* Closes #421, by providing a name-value pair iterator.


Ogeon:
Oops

PR #431, opened 2025-03-19
feat: implemented palette angle/bool_mask/num traits for simba's Wide* wrapper types

### LINKED ISSUE #430: Implement num/bool_mask/angle traits for `Wide*` types of the `simba` crate
First, thanks for this amazing crate :)

## TLDR
```console
error[E0277]: the trait bound `simba::WideF32x8: palette::num::Real` is not satisfied
```

The issue proposes adding implementations of palette’s num, bool_mask, and angle traits for [`simba`](https://docs.rs/simba/0.9.0/simba/)'s Wide* wrapper types (e.g., WideF32x4). This change — enabled behind a feature flag — would let palette work seamlessly with both SIMD and scalar data, allowing developers to write generic color manipulation code without extra wrappers, thus overcoming current orphan rule limitations.

## Motivation
Palette allows color manipulation, management and conversion pretty much independent on how the color is stored. That also means we can use SIMD (SoA) or non-SIMD types for almost everything in palette.
 
So it's the perfect crate to use if one needs to write code to work with color that should be generic over SIMD/non-SIMD color data (e.g. you want to compile a binary for targets that do support SIMD and for targets that don't support SIMD, or you want to use `palette` in code paths that handle SIMD SoA data, and in other paths that work with plain scalars.

The `wide` crate, which `palette` already implements most of it's num/bool_mask/angle traits for, is a great way to write portable SIMD code. However, it does not provide any way to be generic over how many SIMD-lanes you'd like to use, and you will need to write different code on whether the code works with `wide` types or with plain scalars.

That's where the crate [`simba`](https://docs.rs/simba/0.9.0/simba/) comes in. It defines traits that are implemented for scalar types as well as SIMD types, and it also supports the types from the `wide` crate.

However, in order to comply with the orphan rules for traits, the `simba` crate had to  [create wrapper types around the `wide` types, so e.g. `WideF32x4` for `f32x4`](https://github.com/dimforge/simba/blob/45241f49a0f2b0a28ddb7d86c3ccf567d5408f59/src/simd/wide_simd_impl.rs#L57) and implement their trait on the Wrapper types. 

This unfortunately means that these wrapper types are no longer usable with `palette`, because they don't implement `palette`'s num/bool_mask/angle traits:

```console
error[E0277]: the trait bound `simba::WideF32x8: palette::num::Real` is not satisfied
```

## Desired change
If `palette` could implement their num/bool_mask/angle traits (behind a feature flag) for the `simba` `WideF32*` wrapper types, it would enable writing such SIMD/non-SIMD generic code much easier and `palette` usable in such scenarios.

Due again to the orphan rules, the only crates that could implement `palette`'s traits for  the `simba` `WideF32*` wrapper types are `palette` and `simba`. Users of those crates are not able to implement these traits for the wrapper types without wrapping the `WideF32*` types again in another wrapper type. But the user's have to implement the `palette` traits as well as the `simba` traits for their wrapper type.

So I propose to implement the num/bool_mask/angle traits for `simba`'s `WideF32*` wrapper types behind a feature flag.

kije:
Implements the angle/bool_mask/num traits for simba's `Wide*` wrapper types behind a `simba` feature flag.

*Note*: to activate the implementation, both the `wide` and the `simba` feature must be enabled.

## Closed Issues

* Closes #430, by implementing palette's "primitive" traits for simba's wide wrappers

### DIFF
--- a/palette/Cargo.toml
+++ b/palette/Cargo.toml
@@ -35,6 +35,7 @@ find-crate = ["palette_derive/find-crate"]
 std = ["alloc", "approx?/std"]
 alloc = []
 gamma_lut_u16 = []
+wide = ["dep:wide", "simba?/wide"]
 
 # Deprecated. Alias for `"named"`.
 named_from_str = ["named"]
@@ -70,6 +71,10 @@ optional = true
 version = "0.7.3"
 optional = true
 default-features = false
+
+[dependencies.simba]
+version = "0.9"
+optional = true
 
 [dev-dependencies]
 serde_json = "1"
--- a/palette/src/angle.rs
+++ b/palette/src/angle.rs
@@ -8,6 +8,9 @@ use crate::{
 #[cfg(feature = "wide")]
 mod wide;
 
+#[cfg(all(feature = "simba", feature = "wide"))]
+mod simba_wide;
+
 /// Represents types that can express half of a rotation (i.e. 180 degrees).
 pub trait HalfRotation {
     /// Return a value that represents half of a rotation (i.e. 180 degrees).
--- a/palette/src/angle/simba_wide.rs
+++ b/palette/src/angle/simba_wide.rs
@@ -1,0 +1,62 @@
+use ::simba::scalar::SupersetOf;
+use ::simba::simd::{SimdPartialOrd, SimdRealField, WideF32x4, WideF32x8, WideF64x4};
+
+use super::*;
+
+macro_rules! impl_angle_simba_float {
+    ($($ty: ident ($scalar_ty: ident)),+) => {
+        $(
+
+            impl HalfRotation for $ty {
+                #[inline(always)]
+                fn half_rotation() -> Self {
+                    Self::from_subset(&(180.0 as $scalar_ty))
+                }
+            }
+
+            impl FullRotation for $ty {
+                #[inline(always)]
+                fn full_rotation() -> Self {
+                    Self::from_subset(&(360.0 as $scalar_ty))
+                }
+            }
+
+            impl RealAngle for $ty {
+                #[inline(always)]
+                fn degrees_to_radians(self) -> Self {
+                    let rads_per_degree = Self::simd_pi() / Self::from_subset(&(180.0 as $scalar_ty));
+                    self * rads_per_degree
+                }
+
+                #[inline(always)]
+                fn radians_to_degrees(self) -> Self {
+                    let pis_in_180 = Self::from_subset(&(57.2957795130823208767981548141051703 as $scalar_ty));
+                    self * pis_in_180
+                }
+            }
+
+            impl AngleEq for $ty {
+                #[inline(always)]
+                fn angle_eq(&self, other: &Self) -> Self::Mask {
+                    self.normalize_unsigned_angle().simd_eq(other.normalize_unsigned_angle())
+                }
+            }
+
+            impl SignedAngle for $ty {
+                #[inline(always)]
+                fn normalize_signed_angle(self) -> Self {
+                    self - Round::ceil(((self + Self::from_subset(&(180.0 as $scalar_ty))) / Self::from_subset(&(360.0 as $scalar_ty))) - Self::from_subset(&(1.0 as $scalar_ty))) * Self::from_subset(&(360.0 as $scalar_ty))
+                }
+            }
+
+            impl UnsignedAngle for $ty {
+                #[inline(always)]
+                fn normalize_unsigned_angle(self) -> Self {
+                    self - (Round::floor(self / Self::from_subset(&(360.0 as $scalar_ty))) * Self::from_subset(&(360.0 as $scalar_ty)))
+                }
+            }
+        )+
+    };
+}
+
+impl_angle_simba_float!(WideF32x4(f32), WideF32x8(f32), WideF64x4(f64));
--- a/palette/src/bool_mask.rs
+++ b/palette/src/bool_mask.rs
@@ -8,6 +8,9 @@ use core::ops::{BitAnd, BitOr, BitXor, Not};
 #[cfg(feature = "wide")]
 mod wide;
 
+#[cfg(all(feature = "simba", feature = "wide"))]
+mod simba_wide;
+
 /// Associates a Boolean type to the implementing type.
 ///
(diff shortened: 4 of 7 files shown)


palette/Cargo.toml
```diff
@@ -67,10 +68,14 @@ version = "1"
 optional = true
 
 [dependencies.wide]
-version = "0.7.3"
+version = "0.7"
```
Ogeon, line 71: On the contrary, it's better to be explicit about the lowest supported version. Did it cause any issues?

palette/Cargo.toml
```diff
@@ -67,10 +68,14 @@ version = "1"
...
 optional = true
 default-features = false
 
+[dependencies.simba]
+version = "0.9"
```
Ogeon, line 76: It would be good to specify a minor version here too.

Ogeon (commented):
Thank you! I'm going to be a bit busy today, so I may wait with a proper review until later, but I skimmed through it and it looks good so far.

palette/Cargo.toml
kije: Ahh yes, sorry I tested my implementation in another project by pointing cargo to my fork, and encountered some issues because ultimately two different versions of the `simba` crate got installed in my project, and the palette trait weren't implemented on the versions of the types I used in the application.

Was totally unrelated to this version here as it turned out, but while figuring out what was the issue I thought I quickly change it here to see if it makes any difference.

Have reverted this change

palette/Cargo.toml
kije: Done 👍

kije:
I don't think the last failing no_std test/check in CI is related to my changes

```console
error[E0557]: feature has been removed
 --> no_std_test/src/main.rs:1:42
  |
1 | #![cfg_attr(feature = "nightly", feature(start))]
  |                                          ^^^^^ feature has been removed
  |
  = note: not portable enough and never RFC'd

error: cannot find attribute `start` in this scope
  --> no_std_test/src/main.rs:10:3
   |
10 | #[start]
   |   ^^^^^
   |
   = note: `start` is in scope, but it is a function, not an attribute

error[E0[60](https://github.com/Ogeon/palette/actions/runs/13945384633/job/39033052288?pr=431#step:4:61)1]: `main` function not found in crate `no_std_test`
  --> no_std_test/src/main.rs:30:13
   |
30 | fn main() {}
   |             ^ consider adding a `main` function to `no_std_test/src/main.rs`

```

Ogeon:
Yeah, it looks like they changed some things on nightly. 🤔 I can look into it when I'm free and let you know when it should work again, unless someone else gets to it first.

Ogeon (commented):
The `no_std` test has been updated. You can rebase your branch to bring the change in. I have also eyed through the changes again and I think we just need to relax the features a bit to not pull in `std`. Speaking of, it may be a good opportunity to extend the `all_features` feature in the `no_std_test` crate with `palette/wide` and `palette/simba` to verify the setup.

Also, Palette doesn't use the `feat:`, etc. prefixes for anything, so you can drop them from the commits and PR. I'll ask you to squash everything at the end, so feel free to wait until then. :slightly_smiling_face:

But thanks again for submitting this and also for adding the unit tests! :tada:

palette/Cargo.toml
```diff
@@ -71,6 +72,10 @@ version = "0.7.3"
 optional = true
 default-features = false
 
+[dependencies.simba]
+version = "0.9.0"
+optional = true
```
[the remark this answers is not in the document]
kije: Thanks for the review and suggestions. I rebased my branch, and implemented your suggestion regarding `no_std_test` and prevented pulling in `std` when not enabled via palette's `std` feature.

Hope that is what you meant :)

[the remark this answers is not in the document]
Ogeon: Yes, thank you! I'm just wondering now if some of them can be skipped, if they aren't strictly necessary for the implementation here. :thinking: I looked around a bit and we can perhaps skip enabling most or all of them on `wide` and `simba`. Palette should only enable those that are necessary for compiling itself, after all.

Sorry for the back and forth, let me know if you rather prefer to move on.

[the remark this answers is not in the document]
Ogeon: Oh, it looks like `simba` + `wide` has some issues with `no_std` :eyes: https://github.com/Ogeon/palette/actions/runs/14018720715/job/39248266654?pr=431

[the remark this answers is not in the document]
Ogeon: We can just skip testing with `simba` enabled there for now.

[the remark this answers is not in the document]
Ogeon: No worries! Thanks for getting back to this. 🙂 I'll have another look at the code, but all the tests pass now so that's great!

palette/Cargo.toml
```diff
@@ -31,9 +31,11 @@ named = ["phf"]
 random = ["rand"]
 serializing = ["serde", "std"]
 find-crate = ["palette_derive/find-crate"]
-std = ["alloc", "approx?/std"]
+std = ["alloc", "approx?/std", "wide?/std", "simba?/std"]
```
Ogeon, line 34: Just double checking, are all of these required? We didn't need wide's "std" feature before, for example. The same for "libm". If we can get away without some of them, that's better.

palette/README.md
```diff
@@ -53,6 +53,7 @@ These features are disabled by default:
 * `"libm"` - Uses the [`libm`] floating point math library (for when the `std` feature is disabled).
 * `"bytemuck"` - Enables casting between plain data types using [`bytemuck`].
 * `"wide"` - Enables support for using SIMD types from [`wide`].
+* `"simba"` - Enables support for using (the wrapped [`wide`]) SIMD types from [`simba`].
```
Ogeon, line 56: I suppose it's a bit odd to not support the stable none-wide types they have. 🤔 Would that be possible to include, considering it's already implemented with macros?

Ogeon (commented):
This looks pretty much done. The only major request is to also support simba's stable none-wide types (I suppose that's only `AutoSimd`), so wide isn't necessarily required. I don't know why I didn't think of it earlier, so sorry for the scope creep. 😬

After that and minimizing the dependency features further, if possible, I would say it's ready to go. 🎉

PR #443, opened 2025-07-13
Update `rand` to 0.9

Jengamon:
Bumps palette to latest rand `0.9`

## Breaking Change

Because this updates which `rand` is used, yes.


Ogeon:
Hi, thanks for taking the time to submit this. As you noted, changing the `rand` version is breaking, so this PR will either have to be on ice until it's time to start with the next major version (after the next release in my plan) or we offer both versions under different names (like `rand09` for 0.9.X). Let me know which option you would prefer.

As for the test failures, I'll clean up the warnings. I have had a couple of busy months and haven't kept things up to date with the new diagnostics.

Jengamon:
I'll just wait for next version. I just needed the update (b/c that's the version of `rand` I'm using) and I can just patch it locally.

Ogeon:
Alright, everything looks good. Thanks a lot!

PR #464, opened 2026-03-22
Add support for `wide` 1.X

Ogeon:
Add support for `wide` 1.X and use it in tests and benchmarks. I's added as a separate feature for backwards compatibility.


Ogeon:
Never mind, this will need a rust version that supports the 2024 edition.

PR #471, opened 2026-06-07
Change derived `palette` traits to be pre-generated with `codegen` for `palette`'s own types

Ogeon:
This chonker of a commit changes the types in `palette` to use the `codegen` tool instead of proc macros for implementing `palette` traits. The long-term goal is to make `palette_derive` opt-in.

* Introduces the `palette_codegen` helper crate.
* Replaces `#[derive(WithAlpha, ArrayCast, FromColorUnclamped)]` derives with pre-generated code.
* Moves test code that uses the above derives to `integration_tests`.
* Updates `CONTRIBUTING.md` to match these changes.


Ogeon:
I will put this on ice for a bit and include it in the next major release instead. It will not have as much of a pay-off unless the derive macros are made opt-in. That will remove their dependencies from the dependency tree and allow `palette` to be built in parallel with them. It will also allow the removal of the `find_crate` dependency and feature.

